PR #5, opened 2020-09-07
libkrun: Refactor the API to make it future-proof

slp:
Refactor the public C API to add versioning and make it more granular,
so we can extend it without breaking existing users.

Signed-off-by: Sergio Lopez <slp@redhat.com>


giuseppe (approved):
LGTM

PR #25, opened 2021-02-11
Fix compiler warnings of init.c

wainersm:
This enables the `-Wall` GCC flag and then fix the only warning emitted by the compiler.


slp:
LGTM, thanks!

PR #33, opened 2021-07-03
Remove AtomicUsize deprecations 

flouthoc:
* Remove trailing colon , make build happy.
* Remove deprecated `compare_and_swap`
* Bump package to current release.

### DIFF
--- a/src/devices/src/virtio/fs/linux/passthrough.rs
+++ b/src/devices/src/virtio/fs/linux/passthrough.rs
@@ -663,10 +663,12 @@ fn forget_one(
             let new_count = refcount.saturating_sub(count);
 
             // Synchronizes with the acquire load in `do_lookup`.
-            if data
-                .refcount
-                .compare_and_swap(refcount, new_count, Ordering::Release)
-                == refcount
+            if data.refcount.compare_exchange(
+                refcount,
+                new_count,
+                Ordering::Release,
+                Ordering::Relaxed,
+            ) == Ok(refcount)
             {
                 if new_count == 0 {
                     // We just removed the last refcount for this inode. There's no need for an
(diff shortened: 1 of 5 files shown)


src/devices/src/virtio/fs/linux/passthrough.rs
```diff
@@ -665,8 +665,8 @@ fn forget_one(
             // Synchronizes with the acquire load in `do_lookup`.
             if data
                 .refcount
-                .compare_and_swap(refcount, new_count, Ordering::Release)
-                == refcount
+                .compare_exchange(refcount, new_count, Ordering::Release, Ordering::Release)
```
slp, line 668: According to [this](https://doc.rust-lang.org/std/sync/atomic/struct.AtomicUsize.html#method.compare_and_swap), the fourth argument should be `Ordering::Relaxed`.

slp (changes requested):
Thanks for the PR! A couple of comments:

- This should probably be two different commits, one removing the unneeded semicolons, and the other replacing `compare_and_swap` with `compare_exchange`.
- The commit addressing the replacement for `compare_and_swap` should also replace the occurrence at `src/logger/src/logger.rs:318`.
- Please add a [S-o-B](https://www.kernel.org/doc/html/v4.17/process/submitting-patches.html#developer-s-certificate-of-origin-1-1) to each commit, and a brief commit description in addition to the title.

flouthoc:
@slp added all the changes and created separate/signed commit for every change.

src/devices/src/virtio/fs/linux/passthrough.rs
flouthoc: Resolved in latest commit.

flouthoc:
ah rustfmt is missing let me do that.

flouthoc:
@slp PTAL

flouthoc:
@slp did fmt and added all the changes with separate/signed commit for every change.

slp (approved):
LGTM, thanks a lot!

PR #41, opened 2022-01-19
Fix unitialized access of err and erroneous error message.

haesbaert:
malloc(3) already sets errno and we should return on NULL.
--
Hi, sorry if this is too silly I was just going over the code and noticed it, also you could use err(3) or errx(3) and cut some lines.
I have more silly fixes like this I can push if you want.

### DIFF
--- a/examples/chroot_vm.c
+++ b/examples/chroot_vm.c
@@ -45,8 +45,7 @@ int main(int argc, char *const argv[])
     int i;
 
     if (argc < 3) {
-        printf("Invalid arguments\n");
-        printf("Usage: %s NEWROOT COMMAND [ARG...]\n", argv[0]);
+        fprintf(stderr, "usage: %s NEWROOT COMMAND [ARG...]\n", argv[0]);
         return -1;
     }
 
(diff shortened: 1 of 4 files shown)


haesbaert:
I don't have a proper setup to test this yet for regressions, so I didn't build it.

Again, I apologize if the changes are too pedantic.

slp:
Hi! Thanks a lot for taking the interest in working on this. Your changes look good to me, just a couple of comments:

- Could you please add a [Sign-off-by](https://www.kernel.org/doc/html/v4.17/process/submitting-patches.html#sign-your-work-the-developer-s-certificate-of-origin) line to the commits?

- Importing one of the ideas from Rust formatting, I'd prefer to avoid if-else blocks without braces, as I find them prone to errors. Could you please re-add them?

haesbaert:
Thanks for reviewing it.
I've re-added the braces as requested as well as the Sign-off-by.

tylerfanelli:
@haesbaert There seem to be some conflicts with this PR since a lot has been done since its submission. Do you mind rebasing on `main`?

PR #72, opened 2022-08-25
Replace all SEV ioctls with their sev library equivalents

tylerfanelli:
These patches rely on changes to the sev library that are not upstreamed yet, therefore I will be marking as draft at the moment. Will mark as ready for review when changes are upstreamed.


slp:
Thanks for PR @tylerfanelli ! 

I like it, it allows us to drop some code at everything looks cleaner. We'll need to wait not only for the changes to be upstream, but to the rust-sev crate to be updated in Fedora too.

PR #77, opened 2022-08-31
vsock/reaper: TIMEOUT shouldn't be const

tylerfanelli:
Without this fix, the compiler encounters the following error:

```
error: `Duration::new` is not yet stable as a const fn
  --> src/devices/src/virtio/vsock/reaper.rs:10:27
   │
10 | const TIMEOUT: Duration = Duration::new(5, 0);
   |                           ^^^^^^^^^^^^^^^^^^^
```

### DIFF
--- a/src/devices/src/virtio/vsock/reaper.rs
+++ b/src/devices/src/virtio/vsock/reaper.rs
@@ -7,7 +7,6 @@ use super::proxy::Proxy;
 use crossbeam_channel::Receiver;
 
 pub type ProxyMap = Arc<RwLock<HashMap<u64, Mutex<Box<dyn Proxy>>>>>;
-const TIMEOUT: Duration = Duration::new(5, 0);
 
 pub struct ReaperThread {
     receiver: Receiver<u64>,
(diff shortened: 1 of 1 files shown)


tylerfanelli:
cc @slp

slp:
@tylerfanelli The code looks good to me, but I wonder which version of Rust starts complaining about this. Is it some nightly?

tylerfanelli:
> @tylerfanelli The code looks good to me, but I wonder which version of Rust starts complaining about this. Is it some nightly?

`rustc 1.56.0 (09c42c458 2021-10-18)` as well as `rustc 1.64.0-nightly (4493a0f47 2022-08-02)` both failed for me.

slp:
There's something weird happening here, I've just tried with both `rustc 1.65.0-nightly (289279de1 2022-09-04)` and `rustc 1.63.0 (4b91a6ea7 2022-08-08)`, and neither of them complains about it. Perhaps your Rust versions are a bit old?

tylerfanelli:
@slp Yep, after updating my compiler, the problem goes away. Closing.

PR #78, opened 2022-08-31
examples/Makefile: Remove launch-tee binary on make clean

tylerfanelli:
cc @slp


tylerfanelli:
cc @slp

slp:
LGTM, thanks!

PR #81, opened 2022-09-08
amdsev: Replace implementation ioctls with sev library equivalents

tylerfanelli:
This relies on changes from the sev library and an update to 1.0. There will also be an upcoming SNP implementation.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -63,9 +63,9 @@ checksum = "cdb031dd78e28731d87d56cc8ffef4a8f36ca26c38fe2de700543e627f8a464a"
 
 [[package]]
 name = "bitfield"
-version = "0.13.2"
+version = "0.14.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "46afbd2983a5d5a7bd740ccb198caf5b82f45c40c09c0eed36052d91cb92e719"
+checksum = "2d7e60934ceec538daadb9d8432424ed043a904d8e0243f3c6446bce549a46ac"
 
 [[package]]
 name = "bitflags"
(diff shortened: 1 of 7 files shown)


tylerfanelli:
Waiting on: https://github.com/virtee/sev/issues/27

init/init.c
```diff
@@ -152,8 +152,6 @@ static int chroot_luks()
 		waitpid(pid, &wstatus, 0);
 	}
 
-	memset(pass, 0, pass_len);
-
```
slp, line 156: This `memset()` is here to clean up `pass` from memory, which sits in a region allocated by `malloc` above. Could you please explain further why is not needed?

tylerfanelli: Ah, I edited my `init.c` to have `pass` come from read-only memory (i.e. I just used a string literal instead of taking `KRUN_PASS` from the environment). I must've forgotten to remove this commit with some of the other I used for testing. Will drop this commit on next push.

tylerfanelli:
@slp This PR is now ready for review. Only the LAUNCH_UPDATE ioctl has been kept in (as we previously discussed). I only need to update the sev library to 1.0.0 (planning to do within the next couple of days). I will address UPDATE's usage (plus attestation needs) in the SNP patches.

tylerfanelli:
@slp 

I'm confused about the failing checks. All three of them mention two errors:
```
error[E0433]: failed to resolve: use of undeclared crate or module `sev`
  --> src/vmm/src/linux/vstate.rs:50:5
   |
50 | use sev::launch::sev::*;
   |     ^^^ use of undeclared crate or module `sev`

warning: unused import: `std::os::unix::io::RawFd`
  --> src/vmm/src/linux/vstate.rs:13:5
   |
13 | use std::os::unix::io::RawFd;
   |     ^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default
```

One line of code shows both of these errors to be false in `vstate.rs`:
`558 |         launcher: Launcher<Started, RawFd, RawFd>,`

`Launcher` is from `sev::launch::sev::*`, and `RawFd` is from `std::os::unix::io::RawFd`. If you take away either of the above two imports, this file won't be able to compile.

tylerfanelli:
Closing and grouping these commits with the AMD SEV-SNP patches PR.

PR #83, opened 2022-10-21
vsock: Fix signal logic in Muxer

slp:
The signal logic in Muxer was missing the ability to signal through an intc, if present. Fix it here.

Signed-off-by: Sergio Lopez <slp@redhat.com>

### DIFF
--- a/src/devices/src/virtio/vsock/muxer.rs
+++ b/src/devices/src/virtio/vsock/muxer.rs
@@ -242,7 +242,9 @@ impl VsockMuxer {
         if update.signal_queue {
             self.interrupt_status
                 .fetch_or(VIRTIO_MMIO_INT_VRING as usize, Ordering::SeqCst);
-            if let Err(e) = self.interrupt_evt.write(1) {
+            if let Some(intc) = &self.intc {
+                intc.lock().unwrap().set_irq(self.irq_line.unwrap());
+            } else if let Err(e) = self.interrupt_evt.write(1) {
                 warn!("failed to signal used queue: {:?}", e);
             }
         }


src/devices/src/virtio/vsock/muxer.rs
```diff
@@ -242,7 +242,9 @@ impl VsockMuxer {
         if update.signal_queue {
             self.interrupt_status
                 .fetch_or(VIRTIO_MMIO_INT_VRING as usize, Ordering::SeqCst);
-            if let Err(e) = self.interrupt_evt.write(1) {
+            if let Some(intc) = &self.intc {
+                intc.lock().unwrap().set_irq(self.irq_line.unwrap());
```
flouthoc, line 246: Just a drive-by comment, I have no context of prior logic ( please feel free to ignore if this doesn't makes sense )

`LockResult<` is ignored isn't this going to panic in case of err ? or is fatal panic expected here. Maybe something like.

```rust
if let Som(intc) = &self.intc {
if Ok(lock) = intc.lock() {
   ..
   } else {
     // self.iterrupt_evt.write(1)
     // or
     // warn
   }
}

slp: If `Mutex::lock()` fails, it means something has gone very, very wrong (like another thread panicking while holding the lock), and there's no way for us to fully recover, so it's reasonable to panic there. It's one of the few cases where the use of `unwrap()` can be rightfully justified.

flouthoc: @slp ah i see, sure makes sense if fatal panic is expected here. Feel free to mark this as resolved.

PR #108, opened 2023-01-17
SNP Attestation

tylerfanelli:


slp:
So far is looking really good, minor details aside. Do you want a full review now or should we wait until this PR is no longer a draft?

tylerfanelli:
@slp I don't suppose a full review is necessary yet. I'm fine with your comment, and just wanted to gauge your initial thoughts. Thanks!

tylerfanelli:
@slp This is ready for full review.

examples/launch-tee.c
```diff
@@ -74,15 +74,6 @@ int main(int argc, char *const argv[])
         return -1;
     }
 
-    // Use the third (optional) command line argument as the disk image containing a data disk.
-    if (argc > 3) {
-        if (err = krun_set_data_disk(ctx_id, argv[3])) {
```
slp, line 79: Given that this is just an example, and probably we'll switch to double partitions quite soon, I think it'd be better to not change `launch_tee` and just specify `snp-example-data-disk.img` as `DATA_DISK_IMAGE`. That should work, shouldn't it?

init/init.c
```diff
@@ -61,55 +65,170 @@ static void set_rlimits(const char *rlimits)
...
+        }
+
+        if (mount("/dev/vdb", "/teeconfig", "ext4",
+                MS_NOEXEC | MS_NOSUID | MS_RELATIME, NULL) < 0) {
+                perror("mount(/dev/vdb)");
+
+                goto rmdir_teeconfig;
+        }
+
+        fd = open("/teeconfig/snp-config-attest.json", O_RDONLY);
```
slp, line 118: We're using `krun-sev.json` for the filename in `crun` and `oci2cw`, so it's probably a good idea using the same name here.

init/init.c
```diff
@@ -61,55 +65,170 @@ static void set_rlimits(const char *rlimits)
...
+                        url_found = 1;
+                }
+        }
 
-	if (mkdir("/sfs", 0755) < 0 && errno != EEXIST) {
-		perror("mkdir(/sfs)");
-		return NULL;
-	}
-
-	if (mount("securityfs", "/sfs", "securityfs",
```
slp, line 81: We still need this code for SEV/SEV-ES. After opening `krun-sev.json`, we need to check the "tee" field and either use this code or, for the SNP case, use the new one.

init/tee/kbs/kbs_curl.c
```diff
@@ -0,0 +1,164 @@
...
+}
+
+/*
+ * Based on the given cURL error, print the KBS error and return an error
+ * indicator.
+ */
+static int
+KBS_CURL_ERR(char *errmsg)
+{
+        printf("ERROR (kbs_curl_post): %s\n", errmsg);
```
slp, line 209: I'm not entirely sure this function is useful. We're losing the ability to pass format strings (varargs) and it's used from more places than `kbs_curl_post`, so the function name is confusing.

Probably a macro that uses `__func__` to print the function name and supports passing varargs to `printf` would work better.

init/tee/kbs/kbs_types.c
```diff
@@ -50,3 +67,354 @@ kbs_challenge(CURL *curl, char *url, char *json_request, char *nonce)
...
+        size_t certs_size, char *json, BIGNUM *mod, BIGNUM *exp)
+{
+        char buf[4096];
+
+        sprintf(buf, "{");
+        strcpy(json, buf);
+
+        kbs_attestation_marshal_tee_pubkey(json, mod, exp);
+
+        sprintf(buf, "\"tee-evidence\":\"{\\\"report\\\":\\\"{");
```
slp, line 207: Is it worth marshaling  the entire report into JSON? Seems quite expensive to me when compared to, for instance, treating the entire report as a binary blob, encoding it in an hexadecimal string and sending it to the Attestation Server. Then the AS can decode the hexadecimal string into a `[u8]` array, and transmute it into a `sev::firmware::guest::types::AttestationReport`.

init/tee/snp_attest.h
```diff
@@ -3,6 +3,103 @@
...
+        struct cert_table_entry {
+                uuid_t guid;
+                uint32_t offset;
+                uint32_t len;
+        } *entry;
+};
+
+/*
+ * SNP attestation report structure. Based off of the attestation report
+ * structure described in firmware version 1.51.
```
slp, line 50: Is there any chance they've changed this struct again in `1.52`? I had to put a 32 bit padding before `report_data` to align properly its contents.

slp (changes requested):
I see some really nice work here. I've tested and works fine, except for the problem with the `snp_report` misalignment and the measurement comparison (I suspect we might need to update the vmsa blob too).

In addition to the comments in the code, I'd also suggest that, for consistency with `init.c`, please use tabs instead of spaces and always use curly braces even when the conditional has a single statement (we should also put a coding style somewhere).

examples/launch-tee.c
tylerfanelli: Sure. I rebased the commit to just use `snp-example-data-disk.img` as `DATA_DISK_IMAGE`. However, since we require that `DATA_DISK_IMAGE` to attest, I forced the usage of a `DATA_DISK_IMAGE` by failing if `argc != 4`.

init/init.c
tylerfanelli: Changed.

tylerfanelli:
@slp Another note. I'll be removing the certificate code and modifying `SNP_GET_EXT_REPORT` to `SNP_GET_REPORT`, since we're basically leaving it up to the attestation server to retrieve the certificates.

init/tee/kbs/kbs_types.c
```diff
@@ -197,228 +193,35 @@ static void
...
-        sprintf(buf, "\\\\\\\"sig_algo\\\\\\\":%u,", report->signature_algo);
-        strcat(json, buf);
-
-        kbs_attestation_marshal_tcb(json, "current_tcb", &report->current_tcb);
-        strcat(json, ",");
-
-        sprintf(buf, "\\\\\\\"plat_info\\\\\\\":%lu,", report->platform_info);
+        OPENSSL_buf2hexstr_ex(report_hexstr, 0x1000, &report_hexstr_len,
+                (unsigned char *) report, sizeof(*report), '\0');
+        report_hexstr[report_hexstr_len] = '\0';
```
tylerfanelli, line 216: `TODO`: Is this required? Or does OpenSSL null-terminate the string?

init/tee/snp_attest.h
tylerfanelli: Fixed in [1272e95](https://github.com/containers/libkrun/pull/108/commits/1272e956cf212f3877b3eb01b1f4d66719a7f8c7)

init/tee/kbs/kbs_types.c
tylerfanelli: Hex encoded in [1272e95](https://github.com/containers/libkrun/pull/108/commits/1272e956cf212f3877b3eb01b1f4d66719a7f8c7)

init/tee/kbs/kbs_curl.c
tylerfanelli: Fixed in [28e1566](https://github.com/containers/libkrun/pull/108/commits/28e1566dca9ce33a7b1ec05fe3a2864fb7d64fce)

init/init.c
tylerfanelli: Fixed in [dcd5828](https://github.com/containers/libkrun/pull/108/commits/dcd5828f89215578f69b47541bd146490de6d037)

tylerfanelli:
@slp Can you re-review? I've added changes fixing all of your original concerns.

slp:
I've been testing this and it's working great after [fixing the SNP measurement generation in libkrunfw](https://github.com/containers/libkrunfw/pull/42). I've also confirmed still works fine with SEV/SEV-ES.

Could you please rebase so we can merge it?

tylerfanelli:
@slp Rebased :) thanks!

slp (approved):

PR #118, opened 2023-02-16
examples: Fix error handling of krun_create_ctx

rwmjones:
Looking at the source of this function, it appears to return a negative errno on error, or a >= index into a map for success.  Fix the calling code to set errno correctly.


tylerfanelli (approved):
LGTM.

Issue #80, opened 2022-09-08
Clock skew


c3d:
I have very significant clock skew with libkrun:

```
make[1]: Warning: File 'recorder' has modification time 69 s in the future
make[1]: warning:  Clock skew detected.  Your build may be incomplete.
```

The skew seems to increase and be somewhat random. I noticed skew values with 5 digits.

How is time managed in libkrun?

PR #124, opened 2023-03-01
macos: implement host->guest time sync

slp:
macOS can be quite aggressive when it comes to force apps to take long naps for power saving reasons. In addition to this, CNTVCT_EL0 doesn't increase during these naps, meaning the guest is not aware that's missing ticks.

Both things combined mean that guests that run for long enough will suffer from clock skews.

To address this, we introduce here a simple and lightweight host to guest time sync. From the host side, the VMM will send DGRAM packets via vsock with the current time at regular intervals (60 seconds) and when an oversleep (host thread sleeping 3 times more than expected) is detected. On the guest side, "init" spawns a process that listens for such packets and sets the system clock if a delta bigger than 100ms is detected.

Fixes: #80

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,8 +10,12 @@ ifeq ($(SEV),1)
     INIT_DEFS := -DSEV=1
 endif
 
+INIT_DEFS =
 ifeq ($(ROSETTA),1)
-    INIT_DEFS := -D__ROSETTA__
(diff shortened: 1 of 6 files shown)


crobinso (dismissed):
A brief scan looks sane to me, but that's as much as I can do :)

slp:
@crobinso thanks you! :-D

PR #129, opened 2023-03-28
Fix build error on missing uuid headers

pfw:
build_on_krunvm.sh was failing with an error of a missing header in the linux vm.

```
gcc -O2 -static -Wall -D__ROSETTA__ -o init/init init/init.c -D__ROSETTA__
In file included from init/init.c:26:
init/tee/snp_attest.h:8:10: fatal error: uuid/uuid.h: No such file or directory
    8 | #include <uuid/uuid.h>
      |          ^~~~~~~~~~~~~
compilation terminated.
make: *** [Makefile:64: init/init] Error 1
There was a problem building init/init in the VM
```

### DIFF
--- a/build_on_krunvm.sh
+++ b/build_on_krunvm.sh
@@ -17,7 +17,7 @@ if [ $? != 0 ]; then
 	exit -1
 fi
 
-krunvm start libkrun-builder /usr/bin/dnf -- install -y glibc-static gcc make
+krunvm start libkrun-builder /usr/bin/dnf -- install -y glibc-static gcc make libuuid-devel
 if [ $? != 0 ]; then
     krunvm delete libkrun-builder
 	echo "Error running command on VM"


slp:
Thanks for the PR! We shouldn't be bringing `snp_attest.h` into the non-SEV builds. I think we right fix would be putting https://github.com/containers/libkrun/blob/d31747aa92cf83df2abaeb87e2a83311c135d003/init/init.c#L26, https://github.com/containers/libkrun/blob/d31747aa92cf83df2abaeb87e2a83311c135d003/init/init.c#L37 and https://github.com/containers/libkrun/blob/d31747aa92cf83df2abaeb87e2a83311c135d003/init/init.c#L38 between `#ifdef SEV` build-time conditionals.

@pfw Do you want to update this PR with this approach?

pfw:
No problems, I'll look at doing that.

tylerfanelli:
@pfw Is there any update on this? Would you like me to make these changes?

pfw:
Hi @tylerfanelli I’ve not had a chance to look at this but will in the next couple of days but I’m more than happy for you to take a look if you have time.

PR #133, opened 2023-05-18
Remove libfdt dependency

### LINKED ISSUE #30: Remove libfdt dependency
crosvm has reimplemented the functionality required from libfdt into a Rust. Consider importing that code to remove our dependency on libfdt.

https://chromium-review.googlesource.com/c/chromiumos/platform/crosvm/+/2713569/8/arch/src/fdt.rs

arkkors:
Fixes #30, #134 by:

- replacing libfdt with `FdtWriter` from https://crates.io/crates/vm-fdt and removing all bindings,
- fix compiling issues with stable rust in `hvf` by deleting the `_OSUnalignedU16/32` structs which do not seem to be used anywhere, 
- fix compiling issues with stable rust in `virtio_gen` crate by removing the `virtio_gen` crate and replacing it with `virtio-bindings` from `rust-vmm/vm-virtio`
- fix compiler issues with stable rust in `arch_gen` crate by regenerating the `bootparam` bindings with the latest `rust-bindgen`, care has been taken to use the same commit as in the previously generated file 
- remove some memory-related dead code in `arch` to fix tests on aarch64

Tested on aarch64 on current Ubuntu, x86_64 on Linux.


slp (approved):
Hi @arkkors. Thanks a lot for working on this!

It would have been better to split-up unrelated changes (changes in bindings) into a separate PR, but it probably isn't work re-spinning the PR just for that. LGTM otherwise!

Issue #136, opened 2023-06-18
Out of range slice panic for `init` virtio reads.


Reisen:
I'm testing out libkrun with podman and ran into the following error:

```
$ podman run --runtime=/usr/local/bin/krun --rm docker.io/hello-world
panicked at 'range end index 733184 out of range for slice of length 732776', src/devices/src/virtio/fs/linux/passthrough.rs:987:29
```

Which I noticed `732776` is the size of my `init` binary. Indeed this seems to happen specifically when the read is detected against the `init` file's inode.

I believe this may be related to [this check in `setupmapping`](https://github.com/containers/libkrun/blob/main/src/devices/src/virtio/fs/linux/passthrough.rs#L1779)? Which I'm guessing would normally lead to a SIGBUS error when trying to read beyond the end of a partially filled page. I worked around this by padding my `init` with 0's:

```
truncate -s 733184 init/init
```

This does work, but it is a bit of a nasty hack. I'm guessing the main cause here is an off by one page mapping bug of some kind but I am not familiar at all with virtio/fuse so I'm making wild guesses, but it happens with the standard suggested build process.

**EDIT:**

Just spotted  #132 -- I'm assuming this is related, just for context this is happening on master for me, my page size for reference:

```
$ getconf PAGESIZE
4096
```

slp:
Thanks for the report! There was a bug in that piece of code. Userspace may request to read beyond the end of the file, and we must return as many bytes as possible. Could you please check if #138 fixes the problem for you?

PR #138, opened 2023-06-30
devices/fs: fix reading the end of init.krun

slp:
Userspace may request to read beyond the end of the file, in which case we should return as many bytes as possible (which can be zero).

Fixes: #136


tylerfanelli (approved):
LGTM

PR #142, opened 2023-08-16
virtio-net implementation using passt

mtjhrc:


mtjhrc:
Slightly more complex, forward looking API could look something like this: [https://gist.github.com/mtjhrc/76148ab3de9ca0c37e109f4f668d523e](https://gist.github.com/mtjhrc/76148ab3de9ca0c37e109f4f668d523e)

slp:
> More complex, forward looking API could look something like this: https://gist.github.com/mtjhrc/76148ab3de9ca0c37e109f4f668d523e

It's an interesting approach, but I'd prefer to have transparent structs that are both understood by C and Rust sides. This allows us to easily identify API breaking changes that should be dealt by increasing the major version of the package.

Furthermore, I think we should avoid implementing starting `passt` from libkrun, and always require to be passed a file descriptor for the opened socket, leaving that work to the library consumers (crun, krun, chroot_vm...).

This means we can get away by extending the API with a single new function call like this:

```
krun_set_net_passt(uint32_t ctx_id, int fd, uint32_t flags);
```

I'm not really sure if we need the `flags` argument if the binary is executed by our consumers. If it isn't, it can be dropped. If that function is not called, libkrun will default to the original behavior, using TSI.

examples/chroot_vm.c
```diff
@@ -17,6 +17,8 @@
 #define MAX_PATH 4096
 #endif
 
+#define NET_MODE KRUN_NET_MODE_TSI
+
```
slp, line 21: Instead of a build time conditional, please turn this into a command line flag. This might require finally adopting something like `getopt()` to process the command line, but I think it's worth it.

mtjhrc:
Note that I cherry-picked the commits from virtio-gpu PR, to resolve clippy warnings, so they should not be part of this PR

slp (approved):
LGTM, thanks!

PR #143, opened 2023-08-16
init: Report an error when execution of the user binary fails

mtjhrc:
Previously running a non existing binary would just silently fail, now it reports an error.

Not sure about the exit code, but there are two previous exit calls, with values -1 and -2, so I just put -3 here.


slp (approved):
LGTM, thanks.

tylerfanelli (approved):

PR #144, opened 2023-09-26
Import rutabaga_gfx+virtio_gpu from crosvm

slp:
The code is gated behind the "gpu" feature.


alyssarosenzweig:
Is vendoring these libs necessary? Any way to pull them in as external dependencies for easier tracking of upstream? or is that a no-go for upstream?

slp:
> Is vendoring these libs necessary? Any way to pull them in as external dependencies for easier tracking of upstream? or is that a no-go for upstream?

@alyssarosenzweig `virtio-gpu` is not intended to be externally consumed and we also need a number of small but critical changes to make it compatible with libkrun, but I don't think it's really an option.

The case of `rutabaga_gfx` is more interesting, as it seems like in `v0.1.2` they've deliberately dropped some internal dependencies (`base` and `data_model`) and published it in crates.io so it can be internally consumed. On the other hand, I kind of prefer to keep this new crates vendorized, eventually dropping them once things settle upstream, [just as we did with kbs-types](https://github.com/containers/libkrun/commit/d31747aa92cf83df2abaeb87e2a83311c135d003).

tylerfanelli:
@slp Is there any docs showing how I could I try this for myself, to test?

alyssarosenzweig:
> > Is vendoring these libs necessary? Any way to pull them in as external dependencies for easier tracking of upstream? or is that a no-go for upstream?
> 
> @alyssarosenzweig `virtio-gpu` is not intended to be externally consumed and we also need a number of small but critical changes to make it compatible with libkrun, so I don't think it's really an option.
> 
> The case of `rutabaga_gfx` is more interesting, as it seems like in `v0.1.2` they've deliberately dropped some internal dependencies (`base` and `data_model`) and published it in crates.io so it can be internally consumed. On the other hand, I kind of prefer to keep this new crates vendorized, eventually dropping them once things settle upstream, [just as we did with kbs-types](https://github.com/containers/libkrun/commit/d31747aa92cf83df2abaeb87e2a83311c135d003).

ah, entirely reasonable. :+1:

slp:
@tylerfanelli To test it you need a image with sommelier. I've just added a sommelier package to [this COPR repo](https://copr.fedorainfracloud.org/coprs/slp/asahi-krun/) so you can take it from there if you don't want to build it yourself. Something like this should do the trick:

1. Prepare an OCI image with sommelier, gedit (or some other Wayland app to test) and firefox (to test Xwayland):
```
$ podman run -ti --name fedora-gpu fedora:38
# dnf install -y dnf-plugins-core
# dnf copr enable slp/asahi-krun 
# dnf install -y sommelier gedit firefox xorg-x11-server-Xwayland
# exit
```

2. Increase chroot_vm microVM settings (to run firefox) and expand the OCI in a directory:
```
$ cd examples
$ sed -i 's/1, 512/4, 4096/g' chroot_vm.c
$ make
$ mkdir fedora-gpu
$ podman export fedora-gpu | tar xf - -C fedora-gpu
```

3. Start the microVM and run some apps with sommelier:
```
$ ulimit -n 524288
$ buildah unshare
# ./chroot_vm fedora-gpu/ /bin/bash
bash-5.2# echo "nameserver 1.1.1.1" > /etc/resolv.conf
bash-5.2# export XDG_RUNTIME_DIR=/tmp
bash-5.2# sommelier --virtgpu-channel gedit
bash-5.2# sommelier --virtgpu-channel -X  firefox
```

mtjhrc:
For anyone else wants to try this, note that you have to enable virtio-gpu for libkrufw:
```
diff --git a/config-libkrunfw_x86_64 b/config-libkrunfw_x86_64
index 8fb4eb1..abceeb3 100644
--- a/config-libkrunfw_x86_64
+++ b/config-libkrunfw_x86_64
@@ -1369,7 +1369,8 @@ CONFIG_BCMA_POSSIBLE=y
 #
 # Graphics support
 #
-# CONFIG_DRM is not set
+CONFIG_DRM=y
+CONFIG_DRM_VIRTIO_GPU=y
 # CONFIG_DRM_DEBUG_MODESET_LOCK is not set
 
 #
```

PR #146, opened 2023-10-04
Make krun_set_vm_config use the same argument type for num_vcpus as C…

mtjhrc:
Make krun_set_vm_config use the same argument type for num_vcpus as C header

This was previously changed in commit 815d830c5, without any explanation (presumably committed by accident).

### DIFF
--- a/src/libkrun/src/lib.rs
+++ b/src/libkrun/src/lib.rs
@@ -311,7 +311,7 @@ pub extern "C" fn krun_free_ctx(ctx_id: u32) -> i32 {
 }
 
 #[no_mangle]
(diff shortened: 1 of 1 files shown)


tylerfanelli:
@slp Was there a reason for the change to `u32` in https://github.com/containers/libkrun/commit/815d830c5daaf13f2ac4b0fa90ab9637890b3c13?

slp:
I *think* the change was motivated to stop imposing an arbitrary CPU limit, but as @mtjhrc said the change got probably mixed up with the other commit by accident. We haven't run enough tests to raise that limit anyway, so I'm okay with reverting it.

slp (approved):

PR #149, opened 2023-10-18
virtio net bugfixes and performance improvement

mtjhrc:
Since bug https://bugs.passt.top/show_bug.cgi?id=74 is now fixed in passt, it now became possible to test with small socket buffers, doing so revealed some bugs and a performance issue on the libkrun side.

If you want to test this, please use passt  0^20231004.gf851084, because of the aforementioned bug.

Note that there is now a performance issue where increasing the socket buffer sizes to the previously recommended 16Mib decreases performance. The performance of libkrun with small socket buffers is better than QEMU+passt, with bigger buffer sizes they are both kind of slow. Curiously limiting iperf3 speed (using -b) increases the performance. This seems to be another passt issue.

### DIFF
--- a/src/devices/src/virtio/net/device.rs
+++ b/src/devices/src/virtio/net/device.rs
@@ -172,7 +172,12 @@ impl Net {
             .passt
             .try_finish_write(vnet_hdr_len(), &self.tx_frame_buf[..self.tx_frame_len])
         {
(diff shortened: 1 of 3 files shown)


mtjhrc:
The clippy issues seem to be fixed in #148 so not fixing them in this PR.

slp:
@mtjhrc LGTM, could you please force-push this PR to retrigger the checks?

slp (approved):
Thanks, let's merge it now.

PR #150, opened 2023-10-25
Make krun_start_enter error when root directory does not exist

mtjhrc:
Previously krun_start_enter would succeed and the guest kernel would just panic. The root filesystem directory was opened lazily when the guest kernel used fuse init opcode.
This commit changes it so the root directory is opened when creating the fs device.

This only fixes it on Linux, but not in the macOS implementation.

### DIFF
--- a/src/devices/src/virtio/fs/device.rs
+++ b/src/devices/src/virtio/fs/device.rs
@@ -95,7 +95,7 @@ impl Fs {
             device_state: DeviceState::Inactive,
             config,
             shm_region: None,
-            server: Server::new(PassthroughFs::new(fs_cfg).unwrap()),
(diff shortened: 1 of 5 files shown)


mtjhrc:
@slp I changed the macOS implementation too, mostly blindly but it compiles with `cargo build --target aarch64-apple-darwin`.

mtjhrc:
Pipeline fails because of clippy issue fixed (separately) both in #148  and #149.

tylerfanelli:
Is this still in progress?

mtjhrc:
Not sure how this will map to libkrun 2.0 API (what function at what stage will return an error). Let's close this PR and open a new one once we get a new API.

PR #152, opened 2023-11-19
Makefile: De-couple KBS init sources from SEV-SNP

tylerfanelli:
The KBS attestation sources can be separated from the SEV-SNP attestation sources. This will make it a bit easier to introduce new TEEs in the future.


slp:
LGTM, but could you please drop `Fix clippy errors`? After merging #148 we shouldn't need it anymore.

tylerfanelli:
@slp dropped.

slp (approved):

PR #153, opened 2023-12-21
Update rust-vmm deps and bump version for upcoming release

slp:


src/vmm/Cargo.toml
```diff
@@ -14,7 +14,7 @@ crossbeam-channel = "0.5"
 env_logger = "0.9.0"
 libc = ">=0.2.39"
 log = "0.4.0"
-vm-memory = { version = "0.8.0", features = ["backend-mmap"] }
+vm-memory = { version = "0.13", features = ["backend-mmap"] }
```
dm0-, line 17: This dependency will have an upper bound that the others don't.

slp: Right, thanks for pointing this one. Fixed now.

tylerfanelli (approved):

PR #159, opened 2024-02-05
init: accept arguments from the "args" Field

slp:
Podman uses the "args" field to set the executable and the arguments to be run in the container, so parse it if found.


tylerfanelli:
If `args` and `Cmd` are both present, then `config_argv` may be overwritten. Is this expected?

PR #164, opened 2024-02-09
Implement `krun_add_vsock_port()` and `UnixProxy` for guest communication with host UNIX sockets.

tylerfanelli:
This PR introduces two main components:

1) `krun_add_vsock_port()`, which allows a user to map guest vsock port numbers to UNIX IPC sockets running on the host. With this, users can enable VM guest IPC with UNIX sockets running on a host.

2) `vsock::unix` module, implementing the `UnixProxy`, a vsock proxy that facilitates the communication between a guest vsock and UNIX IPC socket running on the host.

### DIFF
--- a/src/devices/src/virtio/vsock/muxer.rs
+++ b/src/devices/src/virtio/vsock/muxer.rs
@@ -11,18 +11,21 @@ use super::defs;
 use super::defs::uapi;
 use super::muxer_rxq::{rx_to_pkt, MuxerRxQ};
 use super::muxer_thread::MuxerThread;
-use super::packet::{TsiGetnameRsp, VsockPacket};
+use super::packet::{TsiConnectReq, TsiGetnameRsp, VsockPacket};
 use super::proxy::{Proxy, ProxyRemoval, ProxyUpdate};
 use super::reaper::ReaperThread;
 use super::tcp::TcpProxy;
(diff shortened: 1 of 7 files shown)


src/devices/src/virtio/vsock/muxer.rs
```diff
@@ -518,10 +518,11 @@ impl VsockMuxer {
                     addr: Ipv4Addr::new(0, 0, 0, 0),
                     port: 0,
                 };
-                unix.connect(pkt, tsi);
+                let update = unix.connect(pkt, tsi);
                 unix.confirm_connect(pkt);
                 proxy_map.insert(id, Mutex::new(Box::new(unix)));
                 ipc_map.remove(&pkt.dst_port());
```
slp, line 524: What's the reason to remove the port from the map? That means the connection can only the established once, doesn't it?

tylerfanelli: Right. Do we want to keep it around for multiple connections?

slp: I think I would expect that to be the case. Sounds safer for now.

tylerfanelli: Fixed.

slp (approved):
LGTM, thanks!

PR #169, opened 2024-02-14
virtio/net: implement gvproxy backend

slp:
Implement a backend in virtio-net compatible with the "-listen-vfkit" mode in gvproxy. In this mode, the transport operates at L2 without adding any headers, and the UNIX socket between libkrun and gvproxy operates in DGRAM mode.


slp:
@mtjhrc could you please take a look at the refactor?

PR #181, opened 2024-04-27
Fix constness when taking an array of string pointers

teohhanhui:
i.e. change `char *const *` to `const char *const *`

The problem was apparent from incorrect constness in the `bindgen` generated bindings: https://gist.github.com/teohhanhui/647227ea483dabdcd8317c6ea228665e

e.g. `*const *mut c_char` instead of the expected `*const *const c_char`, which matches https://github.com/containers/libkrun/blob/5e376689d58c1c56a22d3935722e66fe09d28685/src/libkrun/src/lib.rs

### DIFF
--- a/include/libkrun.h
+++ b/include/libkrun.h
@@ -103,7 +103,7 @@ int32_t krun_set_data_disk(uint32_t ctx_id, const char *disk_path);
  * Returns:
  *  Zero on success or a negative error number on failure.
  */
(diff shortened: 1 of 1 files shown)


tylerfanelli (approved):
LGTM, thanks.

teohhanhui:
clippy failures not related...

slp:
Thanks for the PR! Could you please rebase on top of #182 once it gets merged?

PR #187, opened 2024-05-22
Add SMBIOS OEM Strings support (aarch64 only)

germag:
This PR adds support for SMBIOS, specifically to be able to set
SMBIOS OEM Strings. The main use case is to pass credentials to
systemd[0], this is a requirement to support krunkit on podman-bootc[1].

It adds a new API `krun_set_smbios_oem_strings()` to set SMBIOS OEM
Strings (type 11), and all the internal machinery to provide them to 
the firmware. The included EDK2 firmware expects the SMBIOS tables on
address 0x4000_F000.  This is only supported on aarch64 architectures.

[0] https://systemd.io/CREDENTIALS/
[1] https://github.com/containers/podman-bootc/


germag:
The EDK2 firmware was created using https://github.com/slp/edk2/pull/1

Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [workspace]
-members = ["src/libkrun"]
+members = ["src/libkrun", "src/smbios"]
```
slp, line 2: We don't need to add the crate as a workspace member.

include/libkrun.h
```diff
@@ -243,6 +243,19 @@ int32_t krun_set_snd_device(uint32_t ctx_id, bool enable);
...
+ *
+ * Arguments:
+ *  "ctx_id"      - the configuration context ID.
+ *  "oem_strings" - an array of string pointers.
+ *  "count"       - number of elements in the array.
+ *
+ * Returns:
+ *  Zero on success or a negative error number on failure.
+ */
+int32_t krun_set_smbios_oem_strings(uint32_t ctx_id, const char *const oem_strings[], char count);
```
slp, line 257: It's a bit weird passing `count` as `char`. And, while I agree this strategy is better than using `from_raw_parts` with `MAX_ARGS`, please use it and drop `count` for consistency with other functions in the API. We'll change all of them at once in v2.0.

src/vmm/src/lib.rs
```diff
@@ -202,6 +202,8 @@ pub struct Vmm {
     mmio_device_manager: MMIODeviceManager,
     #[cfg(target_arch = "x86_64")]
     pio_device_manager: PortIODeviceManager,
+
+    _smbios_oem_strings: Option<Vec<String>>,
```
slp, line 206: Let's avoid adding a new field to `Vmm`. Instead, please extend `configure_system` to accept a third argument with `smbios_oem_strings`.

slp (changes requested):

include/libkrun.h
germag: Ok, should I explicitly ask for a NULL ptr terminator of the array?, like:

```
/**
 * Sets the SMBIOS OEM Strings.
 *
 * Arguments:
 *  "ctx_id"      - the configuration context ID.
 *  "oem_strings" - an array of string pointers. Is terminated with an additional NULL pointer.
 *
 * Returns:
 *  Zero on success or a negative error number on failure.
 */
int32_t krun_set_smbios_oem_strings(uint32_t ctx_id, const char *const oem_strings[]);

```

I used `char` (btw it should be `unsigned char`) because the SMBIOS use a `u8` to define the number of strings, and I think it's better to have these restriction in the type system than as a runtime check.

germag: I just tested it more, and I got a segfault, we need to explicitly ask for a NULL terminator (but the from_raw_part is still UB)

germag: I just realized that `inttypes.h` is included... instead of `unsigned char` I could have use `uint8_t`

slp (approved):

PR #188, opened 2024-05-24
builder: only enable TSI if there are no netdevs

slp:
If the "net" feature is enabled, only pass "tsi_hijack" to the kernel if there are no network devices configured.


slp:
@tylerfanelli @mtjhrc PTAL

tylerfanelli (approved):
LGTM.

PR #191, opened 2024-06-02
virtio/fs: Fix corrupted dirents due to trailing garbage

asahilina:
The dirent name length is supposed to be the actual `strlen()` of the name, without any padding. `SYS_getdents64`, however, will pad the records it returns (for alignment, etc.) as well as NUL-terminate the name.

This was causing dirents in the guest kernel to have spurious NUL padding treated as part of the name, which in turn was breaking `getcwd()` under some conditions (and probably other more subtle things).

### DIFF
--- a/src/devices/src/virtio/fs/linux/passthrough.rs
+++ b/src/devices/src/virtio/fs/linux/passthrough.rs
@@ -527,6 +527,11 @@ impl PassthroughFs {
             debug_assert!(namelen <= back.len(), "back is smaller than `namelen`");
 
             let name = &back[..namelen];
+            let term = name
+                .iter()
+                .position(|&a| a == 0)
+                .expect("LinuxDirent64 name not NUL-terminated");
+            let name = &name[..term];
             let res = if name.starts_with(CURRENT_DIR_CSTR) || name.starts_with(PARENT_DIR_CSTR) {
                 // We don't want to report the "." and ".." entries. However, returning `Ok(0)` will
                 // break the loop so return `Ok` with a non-zero value instead.


src/devices/src/virtio/fs/linux/passthrough.rs
```diff
@@ -527,6 +527,8 @@ impl PassthroughFs {
             debug_assert!(namelen <= back.len(), "back is smaller than `namelen`");
 
             let name = &back[..namelen];
+            let term = name.iter().position(|&a| a == 0).expect("LinuxDirent64 name not NUL-terminated");
+            let name = &name[..term];
```
andy9a9, line 531: Shouldn't be just `name` without `let`, while the variable exists already?

alyssarosenzweig: https://users.rust-lang.org/t/mut-vs-redefinition/51098/3

asahilina: That would not work since `name` is not mutable, but there's also no reason to make it mutable (that's more for loops and things like that). As far as I know rebinding names like this is considered good practice.

andy9a9: Fine with me 👍

andy9a9 (approved):
Ok, fine with me 👍

asahilina:
Fixed the code formatting ^^

slp (approved):
Nice catch. Let's get it merged and cut a new minor release.

PR #194, opened 2024-06-04
rutabaga: cross_domain: Fix EINTR handling

asahilina:
If the krun process gets a signal (including things like attaching strace), the resulting EINTR was causing the cross_domain thread to die. Fix this by explicitly handling EINTR and retrying the call.


slp (approved):
LGTM, thanks.

PR #196, opened 2024-06-07
hvf: support ARM PSCI SYSTEM_OFF

djs55:
This is part of the ARM Power State Coordination Interface, see https://developer.arm.com/documentation/den0022/f/?lang=en

There are 2 calls: 1 for "system off" and one for "system reset". Previously we supported only "system reset", add support for "system off" as well.

I tested with a small Go `init` process which ran:
```
func main() {
        _ = syscall.Reboot(syscall.LINUX_REBOOT_CMD_POWER_OFF)
}
```
Before the patch it would keep running. After this patch it powers off as expected.

### DIFF
--- a/src/hvf/src/lib.rs
+++ b/src/hvf/src/lib.rs
@@ -389,7 +389,8 @@ impl<'a> HvfVcpu<'a> {
                         let ret = match val {
                             0x8400_0000 => Some(2),
                             0x8400_0006 => Some(2),
-                            0x8400_0009 => return Ok(VcpuExit::Shutdown),
+                            0x8400_0008 => return Ok(VcpuExit::Shutdown), // ARM Power State Coordination Interface SYSTEM_OFF
+                            0x8400_0009 => return Ok(VcpuExit::Shutdown), // ARM Power State Coordination Interface SYSTEM_RESET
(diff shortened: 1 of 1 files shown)


src/hvf/src/lib.rs
```diff
@@ -389,7 +389,8 @@ impl<'a> HvfVcpu<'a> {
                         let ret = match val {
                             0x8400_0000 => Some(2),
                             0x8400_0006 => Some(2),
-                            0x8400_0009 => return Ok(VcpuExit::Shutdown),
+                            0x8400_0008 => return Ok(VcpuExit::Shutdown), // ARM Power State Coordination Interface SYSTEM_OFF
+                            0x8400_0009 => return Ok(VcpuExit::Shutdown), // ARM Power State Coorgination Interface SYSTEM_RESET
```
dgageot, line 393: Coordination

djs55: Thanks! I've pushed a fix

slp:
Tested here with `systemctl poweroff` and works fine. All checks are green, so let's get this one merged. Thanks!

slp (approved):

PR #200, opened 2024-06-21
vmm: Fix the nomodule kernel parameter

dm0-:
I noticed this kernel parameter was incorrect in the command-line (torvalds/linux@02608bef8f774c058779546926889a2f2717a499).  It looks like it will have no effect since the default libkrunfw configs disable `CONFIG_MODULES`, so the code that handles this parameter isn't built.  Maybe the parameter could be dropped, but this corrects it to take effect if anyone tries to use a modular kernel.


slp (approved):
LGTM, thanks!

PR #202, opened 2024-06-24
Remove contention on the Gic

dgageot:
Instead of locking a single Gic, we move the contention down to the vCPU level.
This gives us better performance for CPU bound tasks.

On a 8 cores M3:
+ Building the Linux Kernel: 7m21s -> 6:58s
+ Building Wasmer: 3m13s -> 2m30s

### DIFF
--- a/src/devices/src/legacy/gic.rs
+++ b/src/devices/src/legacy/gic.rs
@@ -2,9 +2,9 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use crossbeam_channel::Sender;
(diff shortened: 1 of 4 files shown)


dgageot:
Hi!
I've pushed a fix for the linter. Can someone re-trigger the Code Quality jobs, so that I check if something  else is missing?
Thanks!

slp (approved):
LGTM, thanks!

@tylerfanelli could you please take another look?

tylerfanelli (approved):
LGTM, thanks!

PR #203, opened 2024-06-24
virtiofs: fallocate: should enlarge but not shrink the file

djs55:
Before this patch on virtiofs we had:
```
$ fallocate -o 0 -l 1024 f
$ ls -l f
-rw-r--r--    1 root     root          1024 Jun 24 13:00 f
$ fallocate -o 0 -l 512 f
$ ls -l f
-rw-r--r--    1 root     root           512 Jun 24 13:00 f
```

Unfortunately the second `fallocate` call caused the file to shrink, due to the call to `ftruncate`.

On a tmpfs for comparison the file doesn't shrink:
```
$ fallocate -o 0 -l 1024 f
$ ls -l f
-rw-r--r--    1 root     root          1024 Jun 24 12:55 f
$ fallocate -o 0 -l 512 f
$ ls -l f
-rw-r--r--    1 root     root          1024 Jun 24 12:56 f
```

This patch only calls ftuncate() if the proposed_length is larger than the current length.

### DIFF
--- a/src/devices/src/virtio/fs/macos/passthrough.rs
+++ b/src/devices/src/virtio/fs/macos/passthrough.rs
@@ -1731,11 +1731,12 @@ impl FileSystem for PassthroughFs {
 
         let fd = data.file.write().unwrap().as_raw_fd();
 
+        let proposed_length = (offset + length) as i64;
         let mut fs = libc::fstore_t {
             fst_flags: libc::F_ALLOCATECONTIG,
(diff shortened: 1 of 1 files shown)


p1-0tr (approved):

slp (approved):
LGTM, thanks!

I've also noticed `SetattrValid::SIZE` is currently not supported. I'll send a PR to fix that.

PR #205, opened 2024-07-06
Bring back DAX

asahilina:
Re-add the shmem region required for DAX again (after it got removed when virtio-gpu was implemented), and enable DAX.

### DIFF
--- a/src/vmm/src/builder.rs
+++ b/src/vmm/src/builder.rs
@@ -554,13 +554,27 @@ pub fn build_microvm(
         )?;
     }
 
+    // Use half of the shmem region for virtio-gpu and half for DAX
+    let _shm_region_size = arch_memory_info.shm_size >> 1;
+
     #[cfg(not(feature = "tee"))]
-    let _shm_region = Some(VirtioShmRegion {
+    let dax_shm_region = Some(VirtioShmRegion {
         host_addr: guest_memory
             .get_host_address(GuestAddress(arch_memory_info.shm_start_addr))
             .unwrap() as u64,
         guest_addr: arch_memory_info.shm_start_addr,
-        size: arch_memory_info.shm_size as usize,
+        size: _shm_region_size as usize,
+    });
+
+    #[cfg(feature = "gpu")]
+    let gpu_shm_region = Some(VirtioShmRegion {
+        host_addr: guest_memory
+            .get_host_address(GuestAddress(
+                arch_memory_info.shm_start_addr + _shm_region_size,
+            ))
+            .unwrap() as u64,
+        guest_addr: arch_memory_info.shm_start_addr + _shm_region_size,
+        size: _shm_region_size as usize,
     });
 
     let mut vmm = Vmm {
@@ -591,7 +605,7 @@ pub fn build_microvm(
         attach_gpu_device(
             &mut vmm,
             event_manager,
-            _shm_region,
+            gpu_shm_region,
             intc.clone(),
             virgl_flags,
             #[cfg(target_os = "macos")]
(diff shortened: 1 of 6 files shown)


dgageot:
Hey @asahilina, what's the best way to test that DAX is working again?

asahilina:
The kernel will try to use it unconditionally if the mount flag is present (that's how I figured out what was missing, enabling DAX just broke everything until I got it right).

I haven't truly tested it end to end though, but I'll find out soon enough if it's working as advertised since a use case I'm interested in relies on coherent /dev/shm memory mappings between the host and the guest. ^^

dgageot:
> The kernel will try to use it unconditionally if the mount flag is present (that's how I figured out what was missing, enabling DAX just broke everything until I got it right). I haven't truly tested it end to end though, but I'll find out soon enough if it's working as advertised since a use case I'm interested in relies on coherent /dev/shm memory mappings between the host and the guest. ^^
> […](#)

Thanks a lot! If I understand correctly, your use case is using DAX on the boot "disk" that's over virtiofs. I'm interested in using DAX on virtiofs mounts for file sharing between the host and the guest and couldn't see a difference with this PR.
But it's more than possible that I understood it wrong or that I'm doing it wrong.

asahilina:
You need to set the "dax" mount option on your mount, as the last commit (revert) does for the root filesystem. If that works and the flag shows up in /proc/mounts then it should be working as far as I know!

dgageot:
Thank! Yes, that's what I did and in does support the `dax` mount option now, whereas it didn't before. I didn't see any difference on the performance, though.

src/vmm/src/builder.rs
```diff
@@ -591,15 +605,15 @@ pub fn build_microvm(
...
+            gpu_shm_region,
             intc.clone(),
             virgl_flags,
             #[cfg(target_os = "macos")]
             _map_sender,
         )?;
     }
     #[cfg(not(feature = "tee"))]
-    attach_fs_devices(&mut vmm, &vm_resources.fs, None, intc.clone())?;
+    attach_fs_devices(&mut vmm, &vm_resources.fs, dax_shm_region, intc.clone())?;
```
dgageot, line 616: I have the feeling, from what I have tested, that this share memory region can't be used by more than one virtio device.

src/vmm/src/builder.rs
```diff
@@ -554,13 +554,27 @@ pub fn build_microvm(
         )?;
     }
 
+    // Use half of the shmem region for virtio-gpu and half for DAX
+    let _shm_region_size = arch_memory_info.shm_size >> 1;
```
dgageot, line 558: My rust is still very recent: should the underscore be removed, here?

src/vmm/src/builder.rs
asahilina: Seems to work fine for me? I don't see why it wouldn't work as long as the range used by each device does not overlap (which is why I split it in half here).

src/vmm/src/builder.rs
asahilina: That silences the unused variable warning, in the case of a build with the `tee` feature enabled and the `gpu` feature disabled (which would cause it to become unused).

dgageot: Sorry for the noise!

src/vmm/src/builder.rs
dgageot: Yes, the half/half splitting between gpu and fs clearly works.
If the user has two virtiofs devices, they end up sharing the exact same portion, though.
That makes all my virtiofs devices fail to setup, except for the first one.

asahilina: Ohh, sorry, I misunderstood. Maybe @slp knows what the intent was here? I only brought back the previous behavior from back when DAX was enabled, so I'm guessing this same problem already existed before?

slp: We need to implement some kind of shm region manager that manages the window and can be used to request regions of a certain size. I'm adding this to my TODO list and I think I'll have bandwidth to implement it next week, in case no one beats me implementing it first.

slp:
This one got superseded by #212

PR #211, opened 2024-07-29
Add support for arm cca

MatiasVara:
This PR aims at adding support to build realm guests. First commit adds support for `create_guest_memfd()` and `set_user_memory_region2()`. To do this, the `memory_init()` is modified by adding a boolean parameter. This is required when building a confidential guest for arm cca and probably also required by other cases.
The second commit imports the `virtee/cca` crate and adds the steps to build a cca guest. The following items should be completed before merge the PR:
- [x] use populate() only for kernel and initialize() for non-kernel area
- [x] handle the case in which the guest switch from shared  -> private
- [x] correctly calculation of max_ipa

This has been testing using the v7 series for Linux as a guest and v8 series for KVM on FVP model.

Feedback is welcome.

### DIFF
--- a/src/arch/Cargo.toml
+++ b/src/arch/Cargo.toml
@@ -5,6 +5,7 @@ authors = ["The Chromium OS Authors"]
 edition = "2021"
 
 [features]
+cca = [ "tee" ]
 tee = []
 amd-sev = [ "tee" ]
 efi = []
--- a/src/arch/Cargo.toml
+++ b/src/arch/Cargo.toml
@@ -19,8 +19,12 @@ smbios = { path = "../smbios" }
 utils = { path = "../utils" }
 
 [target.'cfg(target_os = "linux")'.dependencies]
-kvm-bindings = { version = ">=0.11", features = ["fam-wrappers"] }
-kvm-ioctls = ">=0.21"
+kvm-bindings = { version = ">=0.8", features = ["fam-wrappers"] , git = "https://github.com/virtee/kvm-bindings", branch = "add_bindings_for_realms" }
+kvm-ioctls = { version = ">=0.17", git = "https://github.com/virtee/kvm-ioctls", branch = "cca" }
+
+
+[target.'cfg(target_arch = "aarch64")'.dependencies]
+vm-fdt = ">= 0.2.0"
 
 [dev-dependencies]
 utils = { path = "../utils" }
--- a/src/devices/Cargo.toml
+++ b/src/devices/Cargo.toml
@@ -6,6 +6,7 @@ edition = "2021"
 
 [features]
 tee = []
+cca = [ "tee" ]
 amd-sev = ["blk", "tee"]
 net = []
 blk = []
--- a/src/devices/Cargo.toml
+++ b/src/devices/Cargo.toml
@@ -43,8 +43,8 @@ lru = ">=0.9"
 [target.'cfg(target_os = "linux")'.dependencies]
 rutabaga_gfx = { path = "../rutabaga_gfx", features = ["x"], optional = true }
 caps = "0.5.5"
-kvm-bindings = { version = ">=0.11", features = ["fam-wrappers"] }
-kvm-ioctls = ">=0.21"
+kvm-bindings = { version = ">=0.8", features = ["fam-wrappers"] , git = "https://github.com/virtee/kvm-bindings", branch = "add_bindings_for_realms" }
+kvm-ioctls = { version = ">=0.17", git = "https://github.com/virtee/kvm-ioctls", branch = "cca" }
 
 [target.'cfg(any(target_arch = "aarch64", target_arch = "riscv64"))'.dependencies]
 vm-fdt = ">= 0.2.0"
--- a/src/libkrun/src/lib.rs
+++ b/src/libkrun/src/lib.rs
@@ -85,8 +85,10 @@ pub struct KrunfwBindings {
         unsafe extern "C" fn(*mut u64, *mut u64, *mut size_t) -> *mut c_char,
     >,
     #[cfg(feature = "tee")]
+    #[cfg(target_arch = "x86_64")]
     get_initrd: libloading::Symbol<'static, unsafe extern "C" fn(*mut size_t) -> *mut c_char>,
     #[cfg(feature = "tee")]
+    #[cfg(target_arch = "x86_64")]
     get_qboot: libloading::Symbol<'static, unsafe extern "C" fn(*mut size_t) -> *mut c_char>,
 }
 
@@ -101,8 +103,10 @@ impl KrunfwBindings {
             KrunfwBindings {
                 get_kernel: krunfw.get(b"krunfw_get_kernel")?,
                 #[cfg(feature = "tee")]
+                #[cfg(target_arch = "x86_64")]
                 get_initrd: krunfw.get(b"krunfw_get_initrd")?,
                 #[cfg(feature = "tee")]
+                #[cfg(target_arch = "x86_64")]
                 get_qboot: krunfw.get(b"krunfw_get_qboot")?,
             }
         })
@@ -255,6 +259,7 @@ impl ContextConfig {
     }
 
     #[cfg(feature = "tee")]
+    #[cfg(target_arch = "x86_64")]
     fn get_tee_config_file(&self) -> Option<PathBuf> {
         self.tee_config_file.clone()
     }
@@ -1591,6 +1596,7 @@ unsafe fn load_krunfw_payload(
     vmr.set_kernel_bundle(kernel_bundle).unwrap();
 
     #[cfg(feature = "tee")]
+    #[cfg(target_arch = "x86_64")]
     {
         let mut qboot_size: usize = 0;
         let qboot_host_addr = unsafe { (krunfw.get_qboot)(&mut qboot_size as *mut usize) };
@@ -1738,6 +1744,7 @@ pub extern "C" fn krun_start_enter(ctx_id: u32) -> i32 {
      * fail.
      */
     #[cfg(feature = "tee")]
+    #[cfg(target_arch = "x86_64")]
     if let Some(tee_config) = ctx_cfg.get_tee_config_file() {
         if let Err(e) = ctx_cfg.vmr.set_tee_config(tee_config) {
             error!("Error setting up TEE config: {e:?}");
@@ -1848,6 +1855,9 @@ pub extern "C" fn krun_start_enter(ctx_id: u32) -> i32 {
     }
 
     #[cfg(feature = "amd-sev")]
+    vmm::worker::start_worker_thread(_vmm.clone(), _receiver.clone()).unwrap();
+
+    #[cfg(feature = "cca")]
     vmm::worker::start_worker_thread(_vmm.clone(), _receiver.clone()).unwrap();
 
     loop {
--- a/src/libkrun/src/lib.rs
+++ b/src/libkrun/src/lib.rs
@@ -41,7 +41,7 @@ use vmm::vmm_config::block::BlockDeviceConfig;
 use vmm::vmm_config::boot_source::{BootSourceConfig, DEFAULT_KERNEL_CMDLINE};
 #[cfg(not(feature = "tee"))]
 use vmm::vmm_config::external_kernel::{ExternalKernel, KernelFormat};
-#[cfg(not(feature = "tee"))]
+#[cfg(any(not(feature = "tee"), feature = "cca"))]
 use vmm::vmm_config::fs::FsDeviceConfig;
 #[cfg(not(feature = "efi"))]
 use vmm::vmm_config::kernel_bundle::KernelBundle;
@@ -474,7 +474,7 @@ pub extern "C" fn krun_set_vm_config(ctx_id: u32, num_vcpus: u8, ram_mib: u32) -
 
 #[allow(clippy::missing_safety_doc)]
 #[no_mangle]
-#[cfg(not(feature = "tee"))]
+#[cfg(any(not(feature = "tee"), feature = "cca"))]
 pub unsafe extern "C" fn krun_set_root(ctx_id: u32, c_root_path: *const c_char) -> i32 {
     let root_path = match CStr::from_ptr(c_root_path).to_str() {
         Ok(root) => root,
--- a/src/vmm/src/builder.rs
+++ b/src/vmm/src/builder.rs
@@ -59,7 +59,7 @@ use crate::terminal::term_set_raw_mode;
 #[cfg(feature = "blk")]
 use crate::vmm_config::block::BlockBuilder;
 use crate::vmm_config::boot_source::DEFAULT_KERNEL_CMDLINE;
-#[cfg(not(any(feature = "tee", feature = "nitro")))]
+#[cfg(not(any(feature = "tee", feature = "nitro", feature = "cca")))]
 use crate::vmm_config::fs::FsDeviceConfig;
 #[cfg(target_os = "linux")]
 use crate::vstate::KvmContext;
@@ -68,7 +68,7 @@ use crate::vstate::MeasuredRegion;
 use crate::vstate::{Error as VstateError, Vcpu, VcpuConfig, Vm};
 use arch::{ArchMemoryInfo, InitrdConfig};
 use device_manager::shm::ShmManager;
-#[cfg(not(any(feature = "tee", feature = "nitro")))]
+#[cfg(not(any(feature = "tee", feature = "nitro", feature = "cca")))]
 use devices::virtio::{fs::ExportTable, VirtioShmRegion};
 use flate2::read::GzDecoder;
 #[cfg(feature = "tee")]
@@ -878,7 +878,7 @@ pub fn build_microvm(
         vm_resources.console_output.clone(),
     )?;
 
-    #[cfg(not(any(feature = "tee", feature = "nitro")))]
+    #[cfg(not(any(feature = "tee", feature = "nitro", feature = "cca")))]
     let export_table: Option<ExportTable> = if cfg!(feature = "gpu") {
         Some(Default::default())
     } else {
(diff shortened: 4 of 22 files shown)


jakecorrenti:
@MatiasVara is this ready for review, or should I wait?

tylerfanelli:
@jakecorrenti I'm addressing the KVM guest_memfd changes (which @MatiasVara also added in this PR) to my latest SEV-SNP patches. This will probably require a rebase after that. After that, it will likely be ready for a review.

MatiasVara:
> @jakecorrenti I'm addressing the KVM guest_memfd changes (which @MatiasVara also added in this PR) to my latest SEV-SNP patches. This will probably require a rebase after that. Once that happens, it will likely be ready for a review.

I think the PR is ready to review. The current issue is that it does not work for the latest series for KVM (v6). I am investigating the issue.

MatiasVara:
I just updated the PR with the fix for v6 series. The fix is only to set the `KVM_ARM_VCPU_REC` feature during the initialization of VCPUs otherwise `vcpu_finalize` would fail.

MatiasVara:
@jakecorrenti @tylerfanelli @slp Feel free to give a quick look. I am pretty sure there many things to fix.

src/vmm/src/builder.rs
```diff
@@ -72,6 +78,11 @@ use vm_memory::Bytes;
 use vm_memory::GuestRegionMmap;
 use vm_memory::{GuestAddress, GuestMemory, GuestMemoryMmap};
 
+use vm_memory::GuestMemoryRegion;
+
+use crossbeam_channel::Sender;
```
jakecorrenti, line 83: We already import this for macos

src/vmm/src/builder.rs
```diff
@@ -72,6 +78,11 @@ use vm_memory::Bytes;
...
+use kvm_bindings::KVM_ARM_VCPU_REC;
```
jakecorrenti, line 84:
```suggestion
#[cfg(feature = "cca")]
use kvm_bindings::KVM_ARM_VCPU_REC;
```

src/vmm/src/builder.rs
```diff
@@ -674,7 +731,7 @@ pub fn build_microvm(
     if let Some(vsock) = vm_resources.vsock.get() {
         attach_unixsock_vsock_device(&mut vmm, vsock, event_manager, intc.clone())?;
         #[cfg(not(feature = "net"))]
-        vmm.kernel_cmdline.insert_str("tsi_hijack")?;
+        //vmm.kernel_cmdline.insert_str("tsi_hijack")?;
         #[cfg(feature = "net")]
```
jakecorrenti, line 735: Should this be CCA only? There's another one a few lines below as well

src/vmm/src/linux/vstate.rs
```diff
@@ -45,7 +50,12 @@ use kvm_bindings::{
     Msrs, KVM_CLOCK_TSC_STABLE, KVM_IRQCHIP_IOAPIC, KVM_IRQCHIP_PIC_MASTER, KVM_IRQCHIP_PIC_SLAVE,
     KVM_MAX_CPUID_ENTRIES, KVM_PIT_SPEAKER_DUMMY,
 };
-use kvm_bindings::{kvm_userspace_memory_region, KVM_API_VERSION};
+use kvm_bindings::{
+    kvm_create_guest_memfd, kvm_memory_attributes, kvm_userspace_memory_region,
+    kvm_userspace_memory_region2, KVM_API_VERSION, KVM_MEMORY_ATTRIBUTE_PRIVATE,
+    KVM_MEMORY_EXIT_FLAG_PRIVATE, KVM_MEM_GUEST_MEMFD, KVM_VM_TYPE_ARM_IPA_SIZE_MASK,
+    KVM_VM_TYPE_ARM_REALM,
+};
```
jakecorrenti, line 58: Arm-specific bindings shouldn't be imported for all flavors

src/vmm/src/linux/vstate.rs
```diff
@@ -1228,12 +1347,36 @@ impl Vcpu {
...
+                    }
+                }
                 // Documentation specifies that below kvm exits are considered
                 // errors.
                 VcpuExit::FailEntry(reason, vcpu) => {
                     error!("Received KVM_EXIT_FAIL_ENTRY signal: reason={reason}, vcpu={vcpu}");
                     Err(Error::VcpuUnhandledKvmExit)
                 }
+                // TODO: to remove this
+                Unsupported(39) => Ok(VcpuEmulation::Handled),
```
jakecorrenti, line 1379: leftover TODO

src/vmm/src/linux/vstate.rs
```diff
@@ -927,7 +1035,12 @@ impl Vcpu {
     /// * `exit_evt` - An `EventFd` that will be written into when this vcpu exits.
     /// * `create_ts` - A timestamp used by the vcpu to calculate its lifetime.
     #[cfg(target_arch = "aarch64")]
-    pub fn new_aarch64(id: u8, vm_fd: &VmFd, exit_evt: EventFd) -> Result<Self> {
+    pub fn new_aarch64(
+        id: u8,
+        vm_fd: &VmFd,
+        exit_evt: EventFd,
+        sender_io: Sender<MemProperties>,
```
jakecorrenti, line 1042: should be CCA only

src/vmm/src/builder.rs
```diff
@@ -1103,17 +1231,19 @@ fn create_vcpus_aarch64(
     guest_mem: &GuestMemoryMmap,
     entry_addr: GuestAddress,
     exit_evt: &EventFd,
+    sender_io: Sender<MemProperties>,
```
jakecorrenti, line 1234: should be CCA only

jakecorrenti (changes requested):
I didn't actually run it, just looked at the code. Overall looks good. Few comments, though.

Thanks for the work!

src/vmm/src/builder.rs
MatiasVara: I can't remember why I commented this line.

src/vmm/src/linux/vstate.rs
MatiasVara: Should I import them only when `#[cfg(feature = "tee")]`?

src/arch/Cargo.toml
```diff
@@ -18,8 +20,8 @@ smbios = { path = "../smbios" }
 utils = { path = "../utils" }
 
 [target.'cfg(target_os = "linux")'.dependencies]
-kvm-bindings = { version = ">=0.8", features = ["fam-wrappers"] }
-kvm-ioctls = ">=0.17"
+kvm-bindings = { version = ">=0.8", features = ["fam-wrappers"] , git = "https://github.com/virtee/kvm-bindings", branch = "add_bindings_for_realms" }
+kvm-ioctls = { version = ">=0.17", git = "https://github.com/virtee/kvm-ioctls", branch = "cca" }
```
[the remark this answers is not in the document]
jakecorrenti: Exactly. I'm going to have to make some changes for my TDX work, so I think we'll need to merge your CCA work and my TDX work into the main branch and then use that here.

src/vmm/src/linux/vstate.rs
tylerfanelli: The memory fault exception isn't TEE-specific (although it could be used that way), so I don't think I agree that this should be CCA only.

jakecorrenti: You're right. I did this part of the review before you and I had talked about the order of patches.

src/libkrun/src/lib.rs
```diff
@@ -1225,9 +1235,12 @@ pub extern "C" fn krun_start_enter(ctx_id: u32) -> i32 {
     #[cfg(target_os = "macos")]
     let (sender, receiver) = unbounded();
 
+    let (io_sender, receiver) = unbounded();
+
     let _vmm = match vmm::builder::build_microvm(
         &ctx_cfg.vmr,
         &mut event_manager,
+        io_sender,
         ctx_cfg.shutdown_efd,
```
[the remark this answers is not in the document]
MatiasVara: I am not sure about the design either.  I do not like that all flavors spawn an extra thread so I think I will change that. Why couldn't you have a function signature as you proposed?

src/libkrun/src/lib.rs
```diff
@@ -1242,6 +1255,61 @@ pub extern "C" fn krun_start_enter(ctx_id: u32) -> i32 {
...
+                                );
+                            }
+                        }
+                    }
+                // from shared to private
+                } else {
+                    for (_index, region) in guest_mem.iter().enumerate() {
+                        if (region.start_addr().raw_value() + region.size() as u64) > m.addr {
+                            let offset = m.addr - region.start_addr().raw_value();
+                            let host_startaddr = m.addr + offset;
```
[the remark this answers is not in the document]
MatiasVara: Yes, it is based on `ram_block_discard_range()`. You are right, it should use actual host address. Thanks.

src/libkrun/src/lib.rs
```diff
@@ -1242,6 +1255,61 @@ pub extern "C" fn krun_start_enter(ctx_id: u32) -> i32 {
...
+                            unsafe {
+                                let _ret = madvise(
```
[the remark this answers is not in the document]
MatiasVara: Right. How do I handle the error in this case?

src/devices/Cargo.toml
```diff
@@ -5,7 +5,9 @@ authors = ["The Chromium OS Authors"]
 edition = "2021"
 
 [features]
+default = ["cca"]
 tee = []
+cca = []
```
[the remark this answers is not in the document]
MatiasVara: Should I not leave `tee =[]` here?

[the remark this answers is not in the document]
jakecorrenti: You should. I think that was a suggestion error. I was trying to suggest just getting rid of `default = ...`

src/libkrun/src/lib.rs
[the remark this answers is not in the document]
jakecorrenti: I don't know how to handle this off the top of my head, but at the very least I think we should probably check if the ret value is < 0 and then log the error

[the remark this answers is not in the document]
MatiasVara: Yes, I will check for that condition.

src/arch/Cargo.toml
```diff
@@ -5,6 +5,8 @@ authors = ["The Chromium OS Authors"]
 edition = "2021"
 
 [features]
+default = ["cca"]
+cca = []
```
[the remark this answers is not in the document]
MatiasVara: If I understand correctly, when I compile with cca flags:
```bash
cargo build --release --features cca
```
The `tee` feature would be enabled. However, most of the code for `tee` is not required and I ended up doing this everywhere:
```diff
-    #[cfg(feature = "tee")]
+    #[cfg(all(feature = "tee", not(feature = "cca")))]
     let qboot_bundle = vm_resources
         .qboot_bundle()
         .ok_or(StartMicrovmError::MissingKernelConfig)?;
```
Can't be `cca` just a different feature?

src/vmm/src/linux/vstate.rs
MatiasVara: After checking this, I figured out this should remain. I'll explain in the code.

src/libkrun/src/lib.rs
```diff
@@ -1242,6 +1255,61 @@ pub extern "C" fn krun_start_enter(ctx_id: u32) -> i32 {
...
+                let _ret = vm
+                    .lock()
+                    .unwrap()
+                    .set_memory_attributes(kvm_memory_attributes {
+                        address: m.addr,
+                        size: m.size,
+                        attributes: m.attributes as u64,
+                        flags: 0,
+                    });
+
```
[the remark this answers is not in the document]
MatiasVara: The problem here is that I still need the index to get the fd.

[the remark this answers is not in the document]
MatiasVara: I open https://github.com/rust-vmm/vm-memory/issues/315 because I think we need an structure to keep track of the mapping `guestdf_mem` and `regions`.

src/vmm/src/linux/vstate.rs
MatiasVara: I think we can enable the extra this extra thread only when the `tee` feature is enabled.

MatiasVara:
After reviewing the Jake comments I came to the conclusion that this PR should be split otherwise is hard to review:

0. Add `Use create_guest_memfd() and set_user_memory_region2()` only when `tee` is enabled.
1. Add the handling of `KVM_MEMORY_FAULT` only for `tee`, which includes the extra thread for handling shared to private and viceverse.
2. `Tee` should be replaced with `tee and x86-64` since all `tee` is for `x86-64`.
3. Add CCA support. This shall be added with `cca and tee` enabled.

src/libkrun/src/lib.rs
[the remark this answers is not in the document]
jakecorrenti: Great idea. I'll keep track of that

jakecorrenti:
> After reviewing the Jake comments I came to the conclusion that this PR should be split otherwise is hard to review:
> 
>     0. Add `Use create_guest_memfd() and set_user_memory_region2()` only when `tee` is enabled.
> 
>     1. Add the handling of `KVM_MEMORY_FAULT` only for `tee`, which includes the extra thread for handling shared to private and viceverse.
> 
>     2. `Tee` should be replaced with `tee and x86-64` since all `tee` is for `x86-64`.

I don't know if this is idiomatic by any means, but could we have something like `tee-x86` and `tee`? `tee-x86` for the x86 specific functionality, and then `tee` for the shared bits? Would that be cleaner than just writing `#[cfg(all(feature = "tee", target_arch = "x86"))]`? Again, I don't know if that's "proper" Rust. 
> 
>     3. Add CCA support. This shall be added with `cca and tee` enabled.
> 
>     4. Eventually some code will be shared for `tee` in `x86-64` and `aarch64` but I can't see it yet.
> 
> 
> @jakecorrenti @tylerfanelli @slp WDYT?

Other than my question above, that all sounds good to me.

tylerfanelli:
> After reviewing the Jake comments I came to the conclusion that this PR should be split otherwise is hard to review:
> 
> 0. Add `Use create_guest_memfd() and set_user_memory_region2()` only when `tee` is enabled.

I don't think `set_user_memory_region2` should be `tee` only, as I believe that on newer kernels, `set_user_memory_region` is deprecated.

> 1. Add the handling of `KVM_MEMORY_FAULT` only for `tee`, which includes the extra thread for handling shared to private and viceverse.

I'm addressing that in my SEV-SNP PR currently.

> 2. `Tee` should be replaced with `tee and x86-64` since all `tee` is for `x86-64`.

Makes sense to me.

> 3. Add CCA support. This shall be added with `cca and tee` enabled.

Do you mean `aarch64 and tee`?

MatiasVara:
> > After reviewing the Jake comments I came to the conclusion that this PR should be split otherwise is hard to review:
> > ```
> > 0. Add `Use create_guest_memfd() and set_user_memory_region2()` only when `tee` is enabled.
> > 
> > 1. Add the handling of `KVM_MEMORY_FAULT` only for `tee`, which includes the extra thread for handling shared to private and viceverse.
> > 
> > 2. `Tee` should be replaced with `tee and x86-64` since all `tee` is for `x86-64`.
> > ```
> 
> I don't know if this is idiomatic by any means, and it depends on whether you can have target arch-specific features, but could we have something like `tee-x86` and `tee`? `tee-x86` for the x86 specific functionality, and then `tee` for the shared bits? Would that be cleaner than just writing `#[cfg(all(feature = "tee", target_arch = "x86_64"))]`? Again, I don't know if that's "proper" Rust or even possible.
> 
I do not know either. I though that `tee and x86` would have code that `tdx` and `snv` share, if that exists.

> > ```
> > 3. Add CCA support. This shall be added with `cca and tee` enabled.
> > 
> > 4. Eventually some code will be shared for `tee` in `x86-64` and `aarch64` but I can't see it yet.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > @jakecorrenti @tylerfanelli @slp WDYT?
> 
> Other than my question above, that all sounds good to me.

MatiasVara:
> > After reviewing the Jake comments I came to the conclusion that this PR should be split otherwise is hard to review:
> > 
> > 0. Add `Use create_guest_memfd() and set_user_memory_region2()` only when `tee` is enabled.
> 
> I don't think `set_user_memory_region2` should be `tee` only, as I believe that on newer kernels, `set_user_memory_region` is deprecated.
> 

That's OK. I could just leave `set_user_memory_region2` for the future.

> > 1. Add the handling of `KVM_MEMORY_FAULT` only for `tee`, which includes the extra thread for handling shared to private and viceverse.
> 
> I'm addressing that in my SEV-SNP PR currently.
> 

OK but I need to split those chunks from my PR anyway.

> > 2. `Tee` should be replaced with `tee and x86-64` since all `tee` is for `x86-64`.
> 
> Makes sense to me.
> 
> > 3. Add CCA support. This shall be added with `cca and tee` enabled.
> 
> Do you mean `aarch64 and tee`?

Yes for the moment.

MatiasVara:
> After reviewing the Jake comments I came to the conclusion that this PR should be split otherwise is hard to review:
> 
> 0. Add `Use create_guest_memfd() and set_user_memory_region2()` only when `tee` is enabled.
> 1. Add the handling of `KVM_MEMORY_FAULT` only for `tee`, which includes the extra thread for handling shared to private and viceverse.
> 2. `Tee` should be replaced with `tee and x86-64` since all `tee` is for `x86-64`.
> 3. Add CCA support. This shall be added with `cca and tee` enabled.
> 4. Eventually some code will be shared for `tee` in `x86-64` and `aarch64` but I can't see it yet.
> 
> @jakecorrenti @tylerfanelli @slp WDYT?

I just split the PR into a few commits. It is the same code though. The first three commits are independent of cca and should be shared by all tees.

MatiasVara:
I've Just rebased the PR and tried it on v8 series.

MatiasVara:
I just rebased and tested on v8. Also, I am now relying on the rust-vmm/kvm repo.

MatiasVara:
I think the following issues shall be addressed before merging:
- [x] support for blk
- [x] Check the EXIT_FAULT
- [ ] Check FDT hardcoded
- [x] Try v9 series

jakecorrenti:
I'm going to close this for the time being since there hasn't been any activity in 11 months. Feel free to re-open it whenever you want to start working on it again.

PR #212, opened 2024-08-09
Implement SHM region management and re-enable DAX in virtiofs

slp:
We have multiple devices (fs and gpu) that can use SHM regions. So far,
    we were creating a single SHM region so only one device could make use
    of it. In this commit, we implement SHM region management so multiple
    devices can request their own regions.

    For the moment, we're hardcoding the SHM region sizes for gpu and fs. A
    future commit will extend the API so users can configure those sizes as
    desired.

### DIFF
--- a/include/libkrun.h
+++ b/include/libkrun.h
@@ -122,6 +122,24 @@ int32_t krun_add_virtiofs(uint32_t ctx_id,
                           const char *c_path);
 
 /**
+ * Adds an independent virtio-fs device pointing to a host's directory with a tag. This
+ * variant allows specifying the size of the DAX window.
+ *
+ * Arguments:
+ *  "ctx_id"         - the configuration context ID.
+ *  "c_tag"          - tag to identify the filesystem in the guest.
+ *  "c_path"         - full path to the directory in the host to be exposed to the guest.
+ *  "shm_size"       - size of the DAX SHM window in bytes.
+ *
+ * Returns:
+ *  Zero on success or a negative error number on failure.
+ */
+int32_t krun_add_virtiofs2(uint32_t ctx_id,
+                           const char *c_tag,
+                           const char *c_path,
+                           uint64_t shm_size);
+
+/**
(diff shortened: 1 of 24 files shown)


slp:
cc/ @asahilina

tylerfanelli:
@slp Any way I can test?

tylerfanelli:
@slp Any way I can test?

slp:
@tylerfanelli yes, with this commit in and building with gpu, in the guest you should see:

- virtio-gpu gets a memory window:
```
[    1.191469] [drm] Host memory window: 0x280000000 +0x200000000
```
- each virtio-fs device gets its own DAX window:
```
[    1.982987] virtiofs virtio4: Cache len: 0x40000000 @ 0x700000000
[    1.989999] virtiofs virtio5: Cache len: 0x40000000 @ 0x740000000
```
- mounting the virtio-fs volume with DAX should work (even on macOS!), and reading files should be possible:
```
slp@localhost:~$ sudo mount -t virtiofs -o dax test /mnt
slp@localhost:~$ dd if=/mnt/mistral-7b-instruct-v0.2.Q4_0.gguf of=/dev/null bs=128k
31348+1 records in
31348+1 records out
4108917024 bytes (4.1 GB, 3.8 GiB) copied, 1.23938 s, 3.3 GB/s
```

slp:
Let's put this on hold until we figure out what's happing in https://github.com/containers/libkrun/pull/213#issuecomment-2305357235

slp:
@tylerfanelli PTAL, should be ready now.

tylerfanelli (approved):
LGTM on x86_64 Linux

@slp Side note, with this built and latest version of libkrunfw, the chroot_vm example is failing for me (x86_64 Linux). I'll submit an issue so we can track this. Can you confirm?

slp:
@tylerfanelli yes, please, open an issue detailing that.

tylerfanelli:
Tracked with #216

PR #214, opened 2024-09-04
devices/legacy: use pl011 for earlycon in aarch64

MatiasVara:
Follow up of 5329be7 commit. It is more common to have a PL011 than UART.


slp:
Could you please rebase on top of HEAD to fix the tests?

slp (approved):
LGTM, thanks!

PR #217, opened 2024-09-09
Add krun_add_disk() method

alyssarosenzweig:
This generalizes over krun_set_{root,data}_disk(). Notably:
    
    * It allows arbitrarily many disk images.
    * It allows mounting disks read-only.
    
    To implement, we refactor a bit the existing handling while allowing
    both APIs to be used together.

This is required to mount systemwide squashfs images which Asahi will use for FEX.

### DIFF
--- a/src/libkrun/src/lib.rs
+++ b/src/libkrun/src/lib.rs
@@ -86,6 +86,8 @@ struct ContextConfig {
     net_cfg: NetworkConfig,
     mac: Option<[u8; 6]>,
     #[cfg(feature = "blk")]
+    block_cfgs: Vec<BlockDeviceConfig>,
+    #[cfg(feature = "blk")]
     root_block_cfg: Option<BlockDeviceConfig>,
     #[cfg(feature = "blk")]
     data_block_cfg: Option<BlockDeviceConfig>,
@@ -156,13 +158,13 @@ impl ContextConfig {
     }
 
     #[cfg(feature = "blk")]
-    fn set_root_block_cfg(&mut self, block_cfg: BlockDeviceConfig) {
-        self.root_block_cfg = Some(block_cfg);
+    fn add_block_cfg(&mut self, block_cfg: BlockDeviceConfig) {
+        self.block_cfgs.push(block_cfg);
     }
 
     #[cfg(feature = "blk")]
-    fn get_root_block_cfg(&self) -> Option<BlockDeviceConfig> {
-        self.root_block_cfg.clone()
+    fn set_root_block_cfg(&mut self, block_cfg: BlockDeviceConfig) {
+        self.root_block_cfg = Some(block_cfg);
     }
 
     #[cfg(feature = "blk")]
@@ -171,8 +173,20 @@ impl ContextConfig {
     }
 
     #[cfg(feature = "blk")]
-    fn get_data_block_cfg(&self) -> Option<BlockDeviceConfig> {
-        self.data_block_cfg.clone()
+    fn get_block_cfg(&self) -> Vec<BlockDeviceConfig> {
+        // For backwards compat, when cfgs is empty (the new API is not used), this needs to be
+        // root and then data, in that order. Also for backwards compat, root/data are setters and
+        // need to discard redundant calls. So we have simple setters above and fix up here.
+        //
+        // When the new API is used, this is simpler.
+        if self.block_cfgs.is_empty() {
+            [&self.root_block_cfg, &self.data_block_cfg]
+                .into_iter()
+                .filter_map(|cfg| cfg.clone())
+                .collect()
+        } else {
+            self.block_cfgs.clone()
+        }
     }
 
     fn set_net_cfg(&mut self, net_cfg: NetworkConfig) {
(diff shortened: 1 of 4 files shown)


src/libkrun/src/lib.rs
```diff
@@ -484,14 +499,53 @@ pub unsafe extern "C" fn krun_set_mapped_volumes(
...
-    //let fs_id = "/dev/root".to_string();
-    //let shared_dir = root_path.to_string();
+    let block_id = match CStr::from_ptr(c_block_id).to_str() {
+        Ok(block_id) => block_id,
+        Err(_) => return -libc::EINVAL,
+    };
+
+    // These block IDs are special cased in the block_cfgs logic for legacy compatibility. Just ban
+    // them for the new API, force the user to avoid unclear/ambiguous API usage.
+    if block_id == "root" || block_id == "data" {
```
tylerfanelli, line 520: Instead of banning them outright, I *think* you could do something like:

`krun_set_root_disk` would forward to `krun_add_disk` with `c_block_id = "root"`
`krun_set_data_disk` would forward to `krun_add_disk` with `c_block_id = "data"`

`krun_add_disk` would have a special case with `c_block_id = "root"` to set the root disk.

AFAIU, there would be no functional difference between any added disk *except* the root disk, correct? So `krun_set_data_disk` may be able to just forward to `krun_add_disk`.

Just trying to make sure we don't end up with useless APIs when replacements would be preferred.

alyssarosenzweig: > AFAIU, there would be no functional difference between any added disk except the root disk, correct? So krun_set_data_disk may be able to just forward to krun_add_disk.

Even the root disk is just a data disk here, there's no special cases whatsoever.

The problem is backwards compat. It's tempting to redefine set_*_disk as adds. However, consider the follow application code:

```
krun_set_data(foo)
krun_set_data(bar)
krun_set_root(baz)
```

Currently, that results in exactly two partitions: root=baz and data=bar, in that guaranteed order.

If we aliased the sets to adds, however, we get three partitions `data=foo, data=bar, root=baz`, in the wrong order, and with colliding block IDs.

If we didn't have a stable API I'd just delete the root/data methods altogether. The compromise I added here is carefully written to ensure no change to existing users. Restricting just the IDs "root/""data" makes it very easy to do that without introducing piles of logic with unclear semantics. (if you add "root" and then set_root, what happens? what if you set_root but then add "root"? etc. Lot of hairy interactions without clear answers. Better to just ban it and do the simplest correct thing.)

tylerfanelli: But for the example you just gave:
```
krun_set_data(foo)
krun_set_data(bar)
krun_set_root(baz)
```
I don't see why that application would exist in the first place. Specifically because a user would not set the data disk twice, since the second call would essentially replace the data disk set first, as you pointed out with `root=baz, data=bar`.

slp: I think there's no "right" solution here, either approach would be fine by me. We'll fix this mess (and many others in the API) in 2.0.

src/libkrun/src/lib.rs
```diff
@@ -156,23 +158,36 @@ impl ContextConfig {
...
-        self.root_block_cfg = Some(block_cfg);
+    fn add_block_cfg(&mut self, block_cfg: BlockDeviceConfig) {
+        self.block_cfgs.push(block_cfg);
     }
 
     #[cfg(feature = "blk")]
-    fn get_root_block_cfg(&self) -> Option<BlockDeviceConfig> {
-        self.root_block_cfg.clone()
+    fn set_root_block_cfg(&mut self, block_cfg: BlockDeviceConfig) {
+        self.root_block_cfg = block_cfg;
```
slp, line 167: This needs to be `Some(block_cfg)`.

src/libkrun/src/lib.rs
```diff
@@ -156,23 +158,36 @@ impl ContextConfig {
...
     }
 
     #[cfg(feature = "blk")]
     fn set_data_block_cfg(&mut self, block_cfg: BlockDeviceConfig) {
-        self.data_block_cfg = Some(block_cfg);
+        self.data_block_cfg = block_cfg;
```
slp, line 172: Ditto.

src/libkrun/src/lib.rs
```diff
@@ -156,23 +158,36 @@ impl ContextConfig {
...
+        // not used), this needs to be root and then data, in that order. Also for backwards
+        // compat, root/data are setters and need to discard redundant calls. So we have simple
+        // setters above and fix up here.
+        for cfg in [self.root_block_cfg, self.data_block_cfg] {
+            if let Some(cfg) = cfg {
+                cfgs.push(cfg);
+            }
+        }
+
+        cfgs
```
slp, line 190: To conform with the comment, I'd suggest replacing this with something like:

```
        if self.block_cfgs.is_empty() {
            [&self.root_block_cfg, &self.data_block_cfg]
                .into_iter()
                .filter_map(|cfg| cfg.clone())
                .collect()
        } else {
            self.block_cfgs.clone()
        }
```

slp (changes requested):

src/libkrun/src/lib.rs
alyssarosenzweig: > I don't see why that application code would (currently) exist in the first place

I mean, sure. But I'm not about to break b/w compat without a semver major bump.

My recommendation remains thus: have the gunk here to retain old behaviour, consider the set* methods deprecated, and then in 2.0 remove the `set*` methods in favour of only `add` and then you can garbage collect this legacy cruft.

src/libkrun/src/lib.rs
alyssarosenzweig: Sorry, why?

alyssarosenzweig: Oh, I see. Fixing, thanks.

src/libkrun/src/lib.rs
alyssarosenzweig: done.

slp (approved):

PR #219, opened 2024-09-11
virtio/fs/linux: announce submounts

slp:
To avoid st.st_ino collisions, rely in the previously introduced support for FUSE submounts and announce them when parent st_dev != inode st_dev, or parent mnt_id != inode mnt_id.


slp:
@asahilina could you please confirm this fixes the issue you were seeing in #204

PR #220, opened 2024-09-26
init: support other locations for krun_config

slp:
So far, the configuration was always expected to be in "/.krun_config.json". To support it to be located elsewhere, use the path stored in KRUN_CONFIG, falling back to the original location if not found there.

### DIFF
--- a/init/init.c
+++ b/init/init.c
@@ -574,6 +574,7 @@ static int config_parse_file(char ***argv, char **workdir)
 	jsmntok_t *tokens;
 	struct stat stat;
 	char *data;
+	char *config_file;
 	char **config_argv;
 	char **entrypoint;
 	int parsed_env, parsed_workdir, parsed_args, parsed_entrypoint;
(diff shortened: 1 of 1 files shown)


init/init.c
```diff
@@ -582,7 +583,13 @@ static int config_parse_file(char ***argv, char **workdir)
...
 	int i;
 
-	fd = open(CONFIG_FILE_PATH, O_RDONLY);
+	config_file = getenv("KRUN_CONFIG");
+	if (config_file) {
+		fd = open(config_file, O_RDONLY);
+	} else {
+		fd = open(CONFIG_FILE_PATH, O_RDONLY);
+	}
+
```
alyssarosenzweig, line 592: A bit cleaner, maybe:

```
	config_file = getenv("KRUN_CONFIG");
	if (!config_file) {
                config_file = CONFIG_FILE_PATH;
       }
	fd = open(config_file, O_RDONLY);
```

alyssarosenzweig: Or if you don't care about MSVC, `open(getenv("KRUN_CONFIG") ?: CONFIG_FILE_PATH, O_RDONLY)` is a one-liner with the elvis operator :man_dancing:

PR #227, opened 2024-10-21
virtio/fs/linux: Fix redundant close()

asahilina:
The fd is owned, so it will be closed on drop. Fixes race condition that could sometimes cause another file to be closed.


slp:
LGTM.

slp (approved):

PR #230, opened 2024-10-22
rutabaga: cross_domain: Fix EPOLLRDHUP logic

asahilina:
This was spuriously returning hung_up: true when no hangup occurred.


asahilina:
Not sure if the merge push was intended? Either way, I just rebased on main.

PR #213, opened 2024-08-20
[RFC] Add X11 cross domain channel type.

WhatAmISupposedToPutHere:
This adds the X11 cross domain channel and an ability to share futexes between the vm and the host. Since it requires dax to work, the commits are now on top of @asahilina's branch that enables dax, but it should be possible to rebase it on top of whatever solution will be used for it.

PR #231, opened 2024-10-22
X11 cross-domain prep work

asahilina:
Prep work for the X11 channel type that does not strictly implement the code. This will be a dependency of #213.

Most of this is implementing support for sharing file handles between the FS code and the GPU code, to implement fence passing. Please let me know if this approach looks sensible.

### DIFF
--- a/src/devices/src/virtio/fs/device.rs
+++ b/src/devices/src/virtio/fs/device.rs
@@ -2,7 +2,7 @@
 use crossbeam_channel::Sender;
 use std::cmp;
 use std::io::Write;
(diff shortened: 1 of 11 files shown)


asahilina:
Note that the original code is also arguably buggy in that it's not handling the fence properly in the EOF case (it should requeue a `HandleFence` job to allow processing to continue without sending out a ring message), though in practice this doesn't make much difference since EOF wasn't signaled anyway so it's going to hang.

slp (approved):

PR #232, opened 2024-10-22
X11 passthrough and futex sharing

asahilina:
Add the X11 cross domain channel type and futex sharing. This is the new approach using explicit FD passing, which depends on #231 (will rebase once that is merged). Also includes a bunch of bug fixes over the previous PR, and CI now passes. I think this should be good to start reviewing.

This approach no longer needs the `LD_PRELOAD`. You can just use nothing (the ptrace code in x112virtgpu should be more robust now), or `xshmwrap` (packaged with x112virgpu) to avoid the ptrace hack. I also sent a [PR](https://gitlab.freedesktop.org/xorg/lib/libxshmfence/-/merge_requests/9) upstream to libxshmfence so we can do this with a simple env var in the future, without hacks.

Compared to #213, this removes the rootfs DAX enablement, but I think @slp wants to do the DAX toggle another way anyway. DAX is still required for /dev/shm for any of this to work, but best keep that out of the scope of this PR.

Supersedes #213.

### DIFF
--- a/src/devices/Cargo.toml
+++ b/src/devices/Cargo.toml
@@ -37,3 +37,6 @@ rutabaga_gfx = { path = "../rutabaga_gfx", features = ["virgl_renderer", "virgl_
 [target.'cfg(target_os = "macos")'.dependencies]
 hvf = { path = "../hvf" }
 lru = ">=0.9"
+
+[target.'cfg(target_os = "linux")'.dependencies]
+rutabaga_gfx = { path = "../rutabaga_gfx", features = ["x"], optional = true }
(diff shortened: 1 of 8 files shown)


asahilina:
Question: Should we renumber CROSS_DOMAIN_CMD_FUTEX_NEW and friends and CROSS_DOMAIN_ID_TYPE_SHM to something higher, in case upstream decides to add commands or types in the future?

asahilina:
There's still more work to be done on the X11 cross-domain channel so I'm marking this as draft. The prep work MR should be OK to review though (and this one too, it's just incomplete and not ready to merge yet).

asahilina:
This should now be ready, Steam finally works stably and Portal 2 runs (including windowed) 🎉

slp:
Please rebase on top of #231

asahilina:
Done!

slp (approved):
Tested both with sommelier and x112virtgpu

PR #237, opened 2024-11-12
Add support for Qcow2 disk images

jakecorrenti:
Use the [imago](https://www.gitlab.com/hreitz/imago) crate to add support for multiple disk image types. This specifically adds support for the use of both Raw and Qcow2 disk images.

### DIFF
--- a/include/libkrun.h
+++ b/include/libkrun.h
@@ -95,16 +95,20 @@ int32_t krun_set_root_disk(uint32_t ctx_id, const char *disk_path);
 int32_t krun_set_data_disk(uint32_t ctx_id, const char *disk_path);
 
 /**
- * Adds a disk image to be used as a general partition for the microVM.
+ * Adds a disk image to be used as a general partition for the microVM. The only supported image
+ * format is "raw".
  *
  * This API is mutually exclusive with the deprecated krun_set_root_disk and
  * krun_set_data_disk methods and must not be used together.
  *
+ * This function deliberately only handles images in the Raw format, because it doesn't allow
+ * specifying an image format, and probing an image's format is dangerous. For more information,
+ * see the security note on `krun_add_disk2`, which allows opening non-Raw images.
+ *
  * Arguments:
  *  "ctx_id"    - the configuration context ID.
  *  "block_id"  - a null-terminated string representing the partition.
- *  "disk_path" - a null-terminated string representing the path leading to the disk image that
- *                contains the root file-system.
+ *  "disk_path" - a null-terminated string representing the path leading to the disk image.
  *  "read_only" - whether the mount should be read-only. Required if the caller does not have
  *                write permissions (for disk images in /usr/share).
  *
@@ -112,6 +116,59 @@ int32_t krun_set_data_disk(uint32_t ctx_id, const char *disk_path);
  *  Zero on success or a negative error number on failure.
  */
 int32_t krun_add_disk(uint32_t ctx_id, const char *block_id, const char *disk_path, bool read_only);
+
+/* Supported disk image formats */
+#define KRUN_DISK_FORMAT_RAW 0
+#define KRUN_DISK_FORMAT_QCOW2 1
+/**
+ * Adds a disk image to be used as a general partition for the microVM. The supported
+ * image formats are: "raw" and "qcow2".
+ *
+ * This API is mutually exclusive with the deprecated krun_set_root_disk and
+ * krun_set_data_disk methods and must not be used together.
+ *
+ * SECURITY NOTE:
+ * Non-Raw images can reference other files, which libkrun will automatically open, and to which the
+ * guest will have access. Libkrun should therefore never be asked to open an image in a non-Raw
+ * format when it doesn't come from a fully trustworthy source.
+ *
+ * Consequently, probing an image's format is quite dangerous and to be avoided if at all possible,
+ * which is why libkrun provides no facilities for doing so. If it's not clear what format an image
+ * has, it may also not be clear whether it can be trusted to not reference files to which the guest
+ * shouldn't have access.
+ *
+ * If probing absolutely can't be avoided, it must only be done on images that are fully trusted, i.e.
+ * before a potentially untrusted guest had write access to it. Specifically, consider that a guest has
+ * full access to all of a Raw image, and can therefore turn it into a file in an arbitrary format, for
+ * example, into a Qcow2 image, referencing and granting a malicious guest access to arbitrary files.
+ * To hand a Raw image to an untrusted and potentially malicious guest, and then to re-probe it after
+ * the guest was able to write to it (when it can no longer be trusted), would therefore be a severe
+ * security vulnerability.
+ *
+ * Therefore, after having probed a yet fully trusted image once, the result must be remembered so the
+ * image will from then on always be opened in the format that was detected originally. When adhering
+ * to this, a guest can write anything they want to a Raw image, it's always going to be opened as a
+ * Raw image, preventing the security vulnerability outlined above.
+ *
+ * However, if at all possible, the image format should be explicitly selected based on knowledge
+ * obtained separately from the pure image data, for example by the user.
+ *
+ * Arguments:
+ *  "ctx_id"      - the configuration context ID.
+ *  "block_id"    - a null-terminated string representing the partition.
+ *  "disk_path"   - a null-terminated string representing the path leading to the disk image.
+ *  "disk_format" - the disk image format (i.e. KRUN_DISK_FORMAT_{RAW, QCOW2})
+ *  "read_only"   - whether the mount should be read-only. Required if the caller does not have
+ *                  write permissions (for disk images in /usr/share).
+ *
+ * Returns:
+ *  Zero on success or a negative error number on failure.
+ */
+int32_t krun_add_disk2(uint32_t ctx_id,
+                       const char *block_id,
+                       const char *disk_path,
+                       uint32_t disk_format,
+                       bool read_only);
 
 /**
  * NO LONGER SUPPORTED. DO NOT USE.
--- a/src/devices/Cargo.toml
+++ b/src/devices/Cargo.toml
@@ -34,6 +34,8 @@ utils = { path = "../utils" }
 polly = { path = "../polly" }
 rutabaga_gfx = { path = "../rutabaga_gfx", features = ["virgl_renderer", "virgl_renderer_next"], optional = true }
 
+imago = { version = "0.1.2", features = ["sync-wrappers", "vm-memory"] }
+
 [target.'cfg(target_os = "macos")'.dependencies]
 hvf = { path = "../hvf" }
 lru = ">=0.9"
--- a/src/devices/src/virtio/block/device.rs
+++ b/src/devices/src/virtio/block/device.rs
@@ -8,7 +8,7 @@
 use std::cmp;
 use std::convert::From;
 use std::fs::{File, OpenOptions};
-use std::io::{self, Seek, SeekFrom, Write};
+use std::io::{self, Write};
 #[cfg(target_os = "linux")]
 use std::os::linux::fs::MetadataExt;
 #[cfg(target_os = "macos")]
@@ -19,6 +19,9 @@ use std::sync::atomic::AtomicUsize;
 use std::sync::{Arc, Mutex};
 use std::thread::JoinHandle;
 
+use imago::file::File as ImagoFile;
+use imago::qcow2::Qcow2;
+use imago::SyncFormatAccess;
 use log::{error, warn};
 use utils::eventfd::{EventFd, EFD_NONBLOCK};
 use virtio_bindings::{
@@ -33,7 +36,7 @@ use super::{
 };
 
 use crate::legacy::Gic;
-use crate::virtio::ActivateError;
+use crate::virtio::{block::ImageType, ActivateError};
 
 /// Configuration options for disk caching.
 #[derive(Clone, Copy, Debug, Default, Eq, PartialEq)]
@@ -51,22 +54,18 @@ pub enum CacheType {
 /// Helper object for setting up all `Block` fields derived from its backing file.
 pub(crate) struct DiskProperties {
     cache_type: CacheType,
-    pub(crate) file: File,
+    pub(crate) file: Arc<SyncFormatAccess<ImagoFile>>,
     nsectors: u64,
     image_id: Vec<u8>,
 }
 
 impl DiskProperties {
     pub fn new(
-        disk_image_path: String,
-        is_disk_read_only: bool,
+        disk_image: Arc<SyncFormatAccess<ImagoFile>>,
+        disk_image_id: Vec<u8>,
         cache_type: CacheType,
     ) -> io::Result<Self> {
-        let mut disk_image = OpenOptions::new()
-            .read(true)
-            .write(!is_disk_read_only)
-            .open(PathBuf::from(&disk_image_path))?;
-        let disk_size = disk_image.seek(SeekFrom::End(0))?;
+        let disk_size = disk_image.size();
 
         // We only support disk size, which uses the first two words of the configuration space.
         // If the image is not a multiple of the sector size, the tail bits are not exposed.
@@ -81,13 +80,13 @@ impl DiskProperties {
         Ok(Self {
             cache_type,
             nsectors: disk_size >> SECTOR_SHIFT,
-            image_id: Self::build_disk_image_id(&disk_image),
+            image_id: disk_image_id,
             file: disk_image,
         })
     }
 
-    pub fn file_mut(&mut self) -> &mut File {
-        &mut self.file
+    pub fn file(&self) -> &SyncFormatAccess<ImagoFile> {
+        self.file.as_ref()
     }
 
     pub fn nsectors(&self) -> u64 {
@@ -141,7 +140,7 @@ impl Drop for DiskProperties {
                     error!("Failed to flush block data on drop.");
                 }
                 // Sync data out to physical media on host.
-                if self.file.sync_all().is_err() {
+                if self.file.sync().is_err() {
                     error!("Failed to sync block data on drop.")
                 }
             }
@@ -168,8 +167,8 @@ pub struct Block {
     // Host file and properties.
     disk: Option<DiskProperties>,
     cache_type: CacheType,
-    disk_image_path: String,
-    is_disk_read_only: bool,
+    disk_image: Arc<SyncFormatAccess<ImagoFile>>,
+    disk_image_id: Vec<u8>,
     worker_thread: Option<JoinHandle<()>>,
     worker_stopfd: EventFd,
 
@@ -203,10 +202,32 @@ impl Block {
         partuuid: Option<String>,
         cache_type: CacheType,
         disk_image_path: String,
+        disk_image_format: ImageType,
         is_disk_read_only: bool,
     ) -> io::Result<Block> {
+        let disk_image = OpenOptions::new()
+            .read(true)
+            .write(!is_disk_read_only)
+            .open(PathBuf::from(&disk_image_path))?;
+
+        let disk_image_id = DiskProperties::build_disk_image_id(&disk_image);
+
+        let disk_image = match disk_image_format {
+            ImageType::Qcow2 => {
+                let mut qcow_disk_image =
+                    Qcow2::<ImagoFile>::open_path_sync(disk_image_path, !is_disk_read_only)?;
+                qcow_disk_image.open_implicit_dependencies_sync()?;
+                SyncFormatAccess::new(qcow_disk_image)?
+            }
+            ImageType::Raw => {
+                let raw = imago::raw::Raw::open_path_sync(disk_image_path, !is_disk_read_only)?;
+                SyncFormatAccess::new(raw)?
+            }
+        };
+        let disk_image = Arc::new(disk_image);
+
         let disk_properties =
-            DiskProperties::new(disk_image_path.clone(), is_disk_read_only, cache_type)?;
+            DiskProperties::new(Arc::clone(&disk_image), disk_image_id.clone(), cache_type)?;
 
         let mut avail_features = (1u64 << VIRTIO_F_VERSION_1)
             | (1u64 << VIRTIO_BLK_F_FLUSH)
@@ -234,8 +255,8 @@ impl Block {
             config,
             disk: Some(disk_properties),
             cache_type,
-            disk_image_path,
-            is_disk_read_only,
+            disk_image,
+            disk_image_id,
             avail_features,
             acked_features: 0u64,
             interrupt_status: Arc::new(AtomicUsize::new(0)),
@@ -348,8 +369,8 @@ impl VirtioDevice for Block {
         let disk = match self.disk.take() {
             Some(d) => d,
             None => DiskProperties::new(
-                self.disk_image_path.clone(),
-                self.is_disk_read_only,
+                Arc::clone(&self.disk_image),
+                self.disk_image_id.clone(),
                 self.cache_type,
             )
             .map_err(|_| ActivateError::BadActivate)?,
--- a/src/devices/src/virtio/file_traits.rs
+++ b/src/devices/src/virtio/file_traits.rs
@@ -6,11 +6,15 @@ use std::fs::File;
 use std::io::{Error, ErrorKind, Result};
 use std::os::unix::io::AsRawFd;
 
+#[cfg(feature = "blk")]
+use imago::io_buffers::{IoVector, IoVectorMut};
 use vm_memory::VolatileSlice;
 
 use libc::{c_int, c_void, read, readv, size_t, write, writev};
 
 use super::bindings::{off64_t, pread64, preadv64, pwrite64, pwritev64};
+#[cfg(feature = "blk")]
+use super::block::device::DiskProperties;
 
 /// A trait for setting the size of a file.
 /// This is equivalent to File's `set_len` method, but
@@ -411,3 +415,42 @@ macro_rules! volatile_impl {
 }
 
 volatile_impl!(File);
+
+#[cfg(feature = "blk")]
+impl FileReadWriteAtVolatile for DiskProperties {
+    fn read_at_volatile(&self, slice: VolatileSlice, offset: u64) -> Result<usize> {
+        self.read_vectored_at_volatile(&[slice], offset)
+    }
+
+    fn read_vectored_at_volatile(&self, bufs: &[VolatileSlice], offset: u64) -> Result<usize> {
+        if bufs.is_empty() {
+            return Ok(0);
+        }
+
+        let (iovec, _guard) = IoVectorMut::from_volatile_slice(bufs);
+        let full_length = iovec
+            .len()
+            .try_into()
+            .map_err(|e| Error::new(ErrorKind::InvalidData, e))?;
+        self.file().readv(iovec, offset)?;
+        Ok(full_length)
+    }
+
+    fn write_at_volatile(&self, slice: VolatileSlice, offset: u64) -> Result<usize> {
+        self.write_vectored_at_volatile(&[slice], offset)
+    }
+
+    fn write_vectored_at_volatile(&self, bufs: &[VolatileSlice], offset: u64) -> Result<usize> {
+        if bufs.is_empty() {
+            return Ok(0);
+        }
+
+        let (iovec, _guard) = IoVector::from_volatile_slice(bufs);
+        let full_length = iovec
+            .len()
+            .try_into()
+            .map_err(|e| Error::new(ErrorKind::InvalidData, e))?;
+        self.file().writev(iovec, offset)?;
+        Ok(full_length)
+    }
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,15 @@
 version = 3
 
 [[package]]
+name = "addr2line"
+version = "0.24.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "dfbe277e56a376000877090da837660b4427aad530e3028d44e0bffe4f89a1c1"
+dependencies = [
+ "gimli",
+]
+
+[[package]]
 name = "adler2"
 version = "2.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -73,6 +82,17 @@ name = "arch_gen"
 version = "0.1.0"
 
 [[package]]
+name = "async-trait"
+version = "0.1.83"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "721cae7de5c34fbb2acd27e21e6d2cf7b886dce0c27388d46c4e6c47ea4318dd"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn",
+]
+
+[[package]]
 name = "atty"
 version = "0.2.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -88,6 +108,21 @@ name = "autocfg"
 version = "1.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ace50bade8e6234aa140d9a2f552bbee1db4d353f69b8217bc503490fc1a9f26"
+
+[[package]]
+name = "backtrace"
+version = "0.3.74"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8d82cb332cdfaed17ae235a638438ac4d4839913cc2af585c3c6746e8f8bee1a"
+dependencies = [
+ "addr2line",
+ "cfg-if",
+ "libc",
+ "miniz_oxide",
+ "object",
+ "rustc-demangle",
+ "windows-targets 0.52.6",
+]
 
 [[package]]
 name = "base64"
@@ -316,6 +351,7 @@ dependencies = [
  "crossbeam-channel",
  "env_logger",
  "hvf",
+ "imago",
  "libc",
  "log",
  "lru",
@@ -510,6 +546,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "gimli"
+version = "0.31.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "07e28edb80900c19c28f1072f2e8aeca7fa06b23cd4169cefe1af5aa3260783f"
+
+[[package]]
 name = "glob"
 version = "0.3.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -583,6 +625,25 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f31827a206f56af32e590ba56d5d2d085f558508192593743f16b2306495269f"
 dependencies = [
  "cc",
+]
+
+[[package]]
+name = "imago"
+version = "0.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "301cd8e9a7de4545ed7387d5d563f4cc5adbdc44d8e658ffbb548bccb5bfe194"
+dependencies = [
+ "async-trait",
+ "bincode",
+ "futures",
+ "libc",
+ "miniz_oxide",
+ "rustc_version",
+ "serde",
+ "tokio",
+ "tracing",
+ "vm-memory",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -869,6 +930,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "object"
+version = "0.36.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "aedf0a2d09c573ed1d8d85b30c119153926a2b36dce0ab28322c09a117a4683e"
+dependencies = [
+ "memchr",
+]
+
+[[package]]
 name = "once_cell"
 version = "1.20.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1111,10 +1181,25 @@ dependencies = [
 ]
 
 [[package]]
+name = "rustc-demangle"
+version = "0.1.24"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "719b953e2095829ee67db738b3bfa9fa368c94900df327b3f07fe6e794d2fe1f"
+
+[[package]]
 name = "rustc-hash"
 version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "08d43f7aa6b08d49f382cde6a7982047c3426db949b1424bc4b7ec9ae12c6ce2"
+
+[[package]]
+name = "rustc_version"
+version = "0.4.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cfcb3a22ef46e85b45de6ee7e79d063319ebb6594faafcf1c225ea92ab6e9b92"
+dependencies = [
+ "semver",
+]
 
 [[package]]
 name = "rutabaga_gfx"
@@ -1146,6 +1231,12 @@ checksum = "01227be5826fa0690321a2ba6c5cd57a19cf3f6a09e76973b58e61de6ab9d1c1"
 dependencies = [
  "windows-sys 0.59.0",
 ]
+
+[[package]]
+name = "semver"
+version = "1.0.23"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "61697e0a1c7e512e84a621326239844a24d8207b4669b41bc18b32ea5cbf988b"
 
 [[package]]
 name = "serde"
@@ -1336,6 +1427,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "tokio"
+version = "1.41.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "22cfb5bee7a6a52939ca9224d6ac897bb669134078daa8735560897f69de4d33"
+dependencies = [
+ "backtrace",
+ "pin-project-lite",
+]
+
+[[package]]
 name = "toml"
 version = "0.8.19"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1367,6 +1468,37 @@ dependencies = [
  "serde_spanned",
  "toml_datetime",
  "winnow",
+]
+
+[[package]]
+name = "tracing"
+version = "0.1.40"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c3523ab5a71916ccf420eebdf5521fcef02141234bbc0b8a49f2fdc4544364ef"
+dependencies = [
+ "pin-project-lite",
+ "tracing-attributes",
+ "tracing-core",
+]
+
+[[package]]
+name = "tracing-attributes"
+version = "0.1.27"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "34704c8d6ebcbc939824180af020566b01a7c01f80641264eba0999f6c2b6be7"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn",
+]
+
+[[package]]
+name = "tracing-core"
+version = "0.1.32"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c06d3da6113f116aaee68e4d601191614c9053067f9ab7f6edbcb161237daa54"
+dependencies = [
+ "once_cell",
 ]
 
 [[package]]
--- a/src/devices/src/virtio/block/mod.rs
+++ b/src/devices/src/virtio/block/mod.rs
@@ -32,3 +32,10 @@ pub enum Error {
     /// Guest gave us a write only descriptor that protocol says to read from.
     UnexpectedWriteOnlyDescriptor,
 }
+
+/// Supported disk image formats
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum ImageType {
+    Raw,
+    Qcow2,
+}
--- a/src/devices/src/virtio/block/worker.rs
+++ b/src/devices/src/virtio/block/worker.rs
@@ -228,7 +228,7 @@ impl BlockWorker {
                     Err(RequestError::InvalidDataLength)
                 } else {
                     writer
-                        .write_from_at(&self.disk.file, data_len, request_header.sector * 512)
+                        .write_from_at(&self.disk, data_len, request_header.sector * 512)
                         .map_err(RequestError::WritingToDescriptor)
                 }
             }
@@ -238,15 +238,15 @@ impl BlockWorker {
                     Err(RequestError::InvalidDataLength)
                 } else {
                     reader
-                        .read_to_at(&self.disk.file, data_len, request_header.sector * 512)
+                        .read_to_at(&self.disk, data_len, request_header.sector * 512)
                         .map_err(RequestError::ReadingFromDescriptor)
                 }
             }
             VIRTIO_BLK_T_FLUSH => match self.disk.cache_type() {
                 CacheType::Writeback => {
-                    let diskfile = self.disk.file_mut();
+                    let diskfile = self.disk.file();
                     diskfile.flush().map_err(RequestError::FlushingToDisk)?;
-                    diskfile.sync_all().map_err(RequestError::FlushingToDisk)?;
+                    diskfile.sync().map_err(RequestError::FlushingToDisk)?;
                     Ok(0)
                 }
                 CacheType::Unsafe => Ok(0),
--- a/src/libkrun/src/lib.rs
+++ b/src/libkrun/src/lib.rs
@@ -18,6 +18,8 @@ use std::sync::Mutex;
 
 #[cfg(target_os = "macos")]
 use crossbeam_channel::unbounded;
+#[cfg(feature = "blk")]
+use devices::virtio::block::ImageType;
 #[cfg(feature = "net")]
 use devices::virtio::net::device::VirtioNetBackend;
 #[cfg(feature = "blk")]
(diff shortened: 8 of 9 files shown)


jakecorrenti:
cc @germag

src/devices/src/virtio/block/disk/mod.rs
```diff
@@ -0,0 +1,89 @@
...
+    read_from_disk(f, 0, &mut magic.data[0..magic_read_len], overlapped_mode)?;
+
+    f.seek(SeekFrom::Start(orig_seek))
+        .map_err(Error::SeekingFile)?;
+
+    #[allow(unused_variables)] // magic4 is only used with the qcow or android-sparse features.
+    if let Some(magic4) = magic.data.get(0..4) {
+        if magic4 == QCOW_MAGIC.to_be_bytes() {
+            return Ok(ImageType::Qcow2);
+        }
```
tylerfanelli, line 74: Does `clippy` throw an error here with `unused_variables`? Surprising if so.

src/devices/Cargo.toml
```diff
@@ -34,6 +34,8 @@ utils = { path = "../utils" }
 polly = { path = "../polly" }
 rutabaga_gfx = { path = "../rutabaga_gfx", features = ["virgl_renderer", "virgl_renderer_next"], optional = true }
 
+imago = { git = "https://gitlab.com/hreitz/imago.git", branch = "main", features = ["sync-wrappers", "vm-memory"] }
```
tylerfanelli, line 37: @XanClic are you planning on releasing a 0.1.0 of `imago`?

tylerfanelli (commented):
Nice work! Some slight nits but overall this LGTM.

tylerfanelli:
Nit: in https://github.com/containers/libkrun/pull/237/commits/8a1795b023c3fac18a9e6c27323561b5fbc0b032 commit description there is a typo: `SyncFormatAccess<File` should be `SyncFormatAccess<File>`

src/devices/Cargo.toml
XanClic: Yes, right. I will do a release. Do we need anything except me declaring 0.1.0 done and pushing it to crates.io?

src/devices/src/virtio/file_traits.rs
```diff
@@ -411,3 +411,49 @@ macro_rules! volatile_impl {
...
+        }
+
+        let mut offset = offset as usize;
+        let mut read = 0;
+        for buf in bufs {
+            let bytes = self.read_at_volatile(*buf, offset as u64)?;
+            offset += bytes;
+            read += bytes;
+        }
+        Ok(read)
```
XanClic, line 435: I think this should just be
```rust
let (iovec, _guard) = imago::io_buffers::IoVectorMut::from_volatile_slice(bufs);
let full_length = iovec.len();
self.readv(iovec, offset)?;
Ok(full_length)
```

But I see that libkrun has `&[VolatileSlice]` here, whereas virtiofsd had `&[&VolatileSlice]`, so I implemented the latter (because that’s what I used for testing)…

I’ll add an interface specifically for `&[VS]`. I’m trying to allow the existing `from_volatile_slice()` to accept it (I hope I can get something done with `IntoIter<Item = &'a VS>`), but if it turns out not to be possible, I’ll add a `from_owned_volatile_slice()` variant.

src/devices/Cargo.toml
XanClic: I’ve just pushed 0.1.0 to crates.io.

src/devices/src/virtio/file_traits.rs
XanClic: I was able to make the interface accept either (basically `IntoIterator<Item: AsRef<VolatileSlice>, IntoIter: ExactSizeIterator>`, to be specific (with a custom `AsRef`)), so the code I put above should now work.

src/devices/Cargo.toml
jakecorrenti: Thank you! AFAIK that's all we need.

src/devices/src/virtio/block/device.rs
```diff
@@ -62,11 +67,29 @@ impl DiskProperties {
         is_disk_read_only: bool,
         cache_type: CacheType,
     ) -> io::Result<Self> {
-        let mut disk_image = OpenOptions::new()
+        let disk_image = OpenOptions::new()
```
XanClic, line 70: It is vital that all accesses to a single disk image go through a single object (write accesses at least), or you risk data corruption because of the internal metadata cache.

So as I suggested before, I think this should be an `Arc<>`; the image should be opened in `Block::new()`, the `Arc<SyncFormatAccess<ImagoFile>>` kept in `Block` (replacing the `disk_image_path`), and consequently also the image ID needs to be built there, then. Something like [0001-Keep-single-instance-of-SyncFormatAccess.txt](https://github.com/user-attachments/files/17734560/0001-Keep-single-instance-of-SyncFormatAccess.txt) should work, I believe. (Sorry for it being .txt instead of .patch, github seems to hate .patch files from Linux users for some reason[[1]](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/attaching-files).)

src/devices/src/virtio/block/disk/mod.rs
```diff
@@ -0,0 +1,89 @@
...
+    }
+}
+/// Detect the type of an image file by checking for a valid header of the supported formats.
+pub fn detect_image_type(file: &File, overlapped_mode: bool) -> Result<ImageType> {
+    let mut f = file;
+    let disk_size = f.get_len().map_err(Error::SeekingFile)?;
+    let orig_seek = f.stream_position().map_err(Error::SeekingFile)?;
+
+    // Try to read the disk in a nicely-aligned block size unless the whole file is smaller.
+    const MAGIC_BLOCK_SIZE: usize = 4096;
```
XanClic, line 46: Is this important? Imago will assume no alignment restrictions unless `direct` is used.

I know it really should probe the actual alignment restrictions (left for a future date…), but also I didn’t know there may be alignment restrictions for non-`direct` I/O. If that’s important, I should fix it. (Either by probing or at least by allowing the “automatically determined” restrictions to be overridden.)

src/devices/src/virtio/block/device.rs
jakecorrenti: Should be fixed now

src/devices/src/virtio/file_traits.rs
jakecorrenti: Also should be fixed.

src/devices/src/virtio/block/disk/mod.rs
jakecorrenti: I honestly don't know enough to give you a definitive answer. I'm going to look into it.

XanClic: That answer’s good enough for me. :smile:

src/devices/src/virtio/file_traits.rs
```diff
@@ -425,14 +425,13 @@ impl FileReadWriteAtVolatile for imago::SyncFormatAccess<imago::file::File> {
...
-            read += bytes;
-        }
-        Ok(read)
+        let (iovec, _guard) = imago::io_buffers::IoVectorMut::from_volatile_slice(bufs);
+        let full_length = iovec
+            .len()
+            .try_into()
+            .map_err(|e| Error::new(ErrorKind::InvalidData, e))?;
+        self.readv(iovec, offset)?;
+        Ok(full_length)
```
XanClic: I think this hunk (and the next one in `write_vectored_at_volatile()`) should be squashed into the previous commit.

XanClic:
I just realized one general thing (because we had that problem in qemu, for at least two CVEs): While it may seem obvious, note that `open_implicit_dependencies_sync()` can/will open other files, whose data will probably be accessible by the guest. Someone needs to verify this. It’s OK to say the user needs to, but still, something to note.

This becomes especially important when it comes to probing the format, as is done here, though: If you use a raw image for a guest, the guest can then write a qcow2 image header into it, and the next time you launch the VM, it will be opened as a qcow2 image.  Because of the external file references qcow2 images have, this basically allows the guest access to any file on the host system the disk emulation has access to.

QEMU got around this problem by (a) very strongly encouraging users to specify the image format for every image, especially for (supposed) raw images; and (b) if raw was probed as a format, blocking writes to the first 512 bytes of the image that it would recognize as an image format header[1].

So (a) is a bit cumbersome to users, and if it were enforced for raw images, it would break compatibility.  You could enforce it for non-raw images, to the same effect (i.e. format not specified = force raw), which keeps compatibility, but may be more cumbersome than you want.

That would leave only (b), i.e. to check if the offset is 0, then take the head of the I/O vector and probe it, potentially denying a write.  Something like
```rust
let mut _copied_head = None;
if offset == 0 {
    let (head, tail) = iovec.split_at(cmp::max(self.mem_align(), 4) as u64);
    _copied_head = Some(head.try_into_owned(self.mem_align())?);
    let head_slice = _copied_head.as_ref().unwrap().as_ref().into_slice();
    if head_slice.get(0..4) == Some(&QCOW_MAGIC.to_be_bytes()) {
        return Err(...);
    }
    iovec = tail.with_inserted(0, head_slice);
}
```

Problem is that the guest could still write those four bytes one by one. qemu solves this with magic[2], but it may be reasonable to just require the guest to write the first sector as a whole (in case of a raw probed image), or at least the first four bytes, and reject anything else.

[1] More detail: When the guest writes to the first 512 bytes of an image that was probed to be raw (i.e. users didn’t specify a format, qemu couldn’t recognize any, so defaulted to using the image as raw), qemu will pass the data that is to be written to every image format driver it has, and let them check whether they would recognize it as an image in their format.  If so, the write is denied.  Note that for this process, the data must be copied into a bounce buffer, or the guest could exploit a race where it pretends it’s writing a normal boot sector, qemu verifies this, passes the buffer on, and then the guest modifies the buffer to be a qcow2 header right before it’s actually written to the file.

[2] For probed raw images, it sets the enforced request alignment to 512, ensuring there’s always a full read-modify-write cycle on the whole first 512 bytes, so it can check it all. Doesn’t work so easily here, because (a) imago does alignment on storage nodes, not format nodes, and (b) this is outside of imago. Now, if you really wanted to, you could emulate the same thing, but it requires a bit of hackiness:
* (This all needs to be done aligned to both memory and request alignment, so: `let full_align = cmp::max(512, cmp::max(self.req_align(), self.mem_align());`)
* You need to get the storage node (of the raw image) first, which you would do by `let Mapping::Raw { storage, offset: _, writable: _ } = self.get_mapping_sync(0, full_align).unwrap().0;`.
* Then you install a strong write blocker: `let _rmw_guard = storage.strong_write_blocker(0..full_align).await;` (I could add a function to `SyncFormatWrapper` to allow running async code in its runtime)
* Then you read the first sector (into an aligned buffer, i.e. `IoBuffer::new(full_align, full_align)`), paste the data from the guest into it, and probe it whether you recognize it as a format. If so, return an error.
* If not, you write the first sector (via `storage.pure_writev()` to prevent deadlocks because of the write blocker)
* Then you write the rest, if any.

germag:
@XanClic Idk how difficult could be with imago, but could also be a solution to just skip/hide the first 512 bytes of a raw image?, I mean from the guest POV the image will be 512 bytes smaller, so the guest will be able to write in the first 512 bytes of its reduced view

germag:
IMO a strict a) in libkrun, with "format not specified = force raw" in the krunkit side, is better. 

You will need to add a new api like ` krun_add_disk_with_format()` that will take a format parameter, keeping  `krun_add_disk()` (and ` krun_set_data_disk()`)  only for RAW images. I'll also add a big doc warning about this.

XanClic:
> @XanClic Idk how difficult could be with imago, but could also be a solution to just skip/hide the first 512 bytes of a raw image?, I mean from the guest POV the image will be 512 bytes smaller, so the guest will be able to write in the first 512 bytes of its reduced view (like a blank header for raw images)

Well, easy, because e.g. qemu’s raw “driver” allows you to specify `offset` and `length` parameters, and we could do the same, but…

> Second thought that is incompatible with the idea of a "raw" image... so forget about it

Yes. Some image formats actually have their header in the first 512 bytes and then the rest is raw (I think there’s a vdi variant to that effect?). You could also create a thin qcow2 image that has the raw file as its external data file, but… It’s just not a raw image then.

src/devices/src/virtio/file_traits.rs
jakecorrenti: Yes you're right. I edited the wrong commit. Nice catch!

germag:
> What I'm gathering from the feedback is that there are opportunities for misusing the disk image. Mitigation techniques:
> 
>     * Enforce the user specify the disk image format if they want to use something other than a raw file for the disk image. This could be an API like `krun_add_disk_with_format(ctx_id, block_id, disk_path, disk_format, read_only)` which German suggested. It's important to thoroughly document this for the user and emphasize avoiding probing at all costs. Libkrun should be the only one probing

(1) Can probing inside libkrun be avoided, expecting the lib user (like krunkit) to do the probing?

> 
>     * If the user originally provided a raw file, prevent the guest from changing the image header to represent a formatted image. We can check on write if the guest is trying to write a different header. But we also need to prevent the guest trying to write a different header one byte at a time.

Idk if it's a good idea: the mitigation should be optional, like an extra parameter in  `krun_add_disk_with_format()`

> But we also need to prevent the guest trying to write a different header one byte at a time.

...and in any order

jakecorrenti:
> > What I'm gathering from the feedback is that there are opportunities for misusing the disk image. Mitigation techniques:
> > ```
> > * Enforce the user specify the disk image format if they want to use something other than a raw file for the disk image. This could be an API like `krun_add_disk_with_format(ctx_id, block_id, disk_path, disk_format, read_only)` which German suggested. It's important to thoroughly document this for the user and emphasize avoiding probing at all costs. Libkrun should be the only one probing
> > ```
> 
> Can probing inside libkrun be avoided, expecting the lib user (like krunkit) to do the probing?

What I gather from Hanna's comment (https://github.com/containers/libkrun/pull/237#issuecomment-2476389107) is yes we can. We would just need to ensure that the probe is happening **once** and before first boot. But it sounds like it would make one of the mitigation options irrelevant.

> 
> > ```
> > * If the user originally provided a raw file, prevent the guest from changing the image header to represent a formatted image. We can check on write if the guest is trying to write a different header. But we also need to prevent the guest trying to write a different header one byte at a time.
> > ```
> 
> Idk if it's a good idea: ~the mitigation should be optional, like an extra parameter in `krun_add_disk_with_format()`~
> 
> ...it's not a good idea
> 
> > But we also need to prevent the guest trying to write a different header one byte at a time.
> 
> ...and in any order

XanClic:
>> Can probing inside libkrun be avoided, expecting the lib user (like krunkit) to do the probing?
>
> What I gather from Hanna's comment (https://github.com/containers/libkrun/pull/237#issuecomment-2476389107) is yes we can. We would just need to ensure that the probe is happening once and before first boot. But it sounds like it would make one of the mitigation options irrelevant.

Indeed.

However, I personally highly doubt it is reasonable to expect every libkrun user to always ensure they probe in this secure manner. To me, it feels like just pushing the security issue can down the road, so that if something happens we can point at the documentation and say “we warned you”.

Then again, that’ll be done anyway for the question of whether a qcow2 image is safe to open, i.e. consider a downloaded qcow2 image that has /etc/shadow as its backing file[1]. libkrun can’t check whether those file references are safe, the user must do that, and who knows whether they will (experience says probably not).

> ...it's not a good idea
 >>   But we also need to prevent the guest trying to write a different header one byte at a time.
>
> ...and in any order

As I described, doable if you always write the header in an RMW manner, i.e. if anything is written to the first sector at all, the whole first sector is read, the new data is embedded, then you check the resulting sector data as a whole, and write it as a whole, thus ensuring you never write an image header into it.

[1] Yes, I know, shouldn’t be accessible anyway, it’s just an example.

XanClic:
> Idk how cumbersome will be adding (and using) a "safe" API, that for instance requires a list of allowed/deny locations/path prefixes for backing files.

If we come up with something that seems nice and generic enough, it could be done right in imago so libkrun could continue to use `open_implicit_dependencies()` and wouldn’t need to re-implement recursively opening the backing chain. Could be an `Fn(&str) -> bool` callback that’s invoked for every file imago would like to open.

germag:
> > Idk how cumbersome will be adding (and using) a "safe" API, that for instance requires a list of allowed/deny locations/path prefixes for backing files.
> 
> If we come up with something that seems nice and generic enough, it could be done right in imago so libkrun could continue to use `open_implicit_dependencies()` and wouldn’t need to re-implement recursively opening the backing chain. Could be an `Fn(&str) -> bool` callback that’s invoked for every file imago would like to open.

I like this idea

src/devices/src/virtio/file_traits.rs
```diff
@@ -411,3 +412,98 @@ macro_rules! volatile_impl {
...
+            #[cfg(feature = "blk")]
+            if head_slice.get(0..4) == Some(&disk::QCOW_MAGIC.to_be_bytes()) {
+                return Err(Error::new(
+                    ErrorKind::InvalidData,
+                    "writing QCOW2 header to file",
+                ));
+            }
+
+            iovec = tail.with_inserted(0, head_slice);
+        }
```
jakecorrenti, line 458: I don't like how this bit is duplicated. I tried pulling it out into its own function to be shared but I got some errors regarding the fact that `IoVector` didn't implement the `Copy` trait.

Are there any suggestions on how I could possibly write this better to be more concise?

jakecorrenti:
In the latest batch of changes I tried to implement the mitigations in order to prevent the guest from misusing the disk image

src/devices/src/virtio/file_traits.rs
germag: > I don't like how this bit is duplicated. I tried pulling it out into its own function to be shared but I got some errors regarding the fact that `IoVector` didn't implement the `Copy` trait.

Can you share here that code?

> 
> Are there any suggestions on how I could possibly write this better to be more concise?

just a nit, but I prefer to add an empty line before `#[cfg(feature = "blk")]`. 
Adding empty lines around conditional code blocks makes it easy to read/spot them

Also, why `_copied_head` is `_` prefixed?

src/devices/src/virtio/block/disk/mod.rs
```diff
@@ -0,0 +1,89 @@
+// Copyright 2019 The ChromiumOS Authors
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
```
germag, line 4: Is this file a new one?, if that is the case, why `Copyright 2019 The ChromiumOS Authors`?

src/devices/src/virtio/file_traits.rs
XanClic: > `IoVector` didn't implement the `Copy` trait.

That’s very much intentional, the guest has control over these buffers (hence vm-memory calling it `VolatileSlice`), so the data in there must be used exactly once, not more.

> Also, why _copied_head is _ prefixed?

Otherwise Rust complains that it’s not read.

include/libkrun.h
```diff
@@ -113,6 +119,35 @@ int32_t krun_set_data_disk(uint32_t ctx_id, const char *disk_path);
...
+ *  "disk_path"   - a null-terminated string representing the path leading to the disk image that
+ *                  contains the root file-system.
+ *  "disk_format" - the formatted disk image format
+ *  "read_only"   - whether the mount should be read-only. Required if the caller does not have
+ *                  write permissions (for disk images in /usr/share).
+ *
+ * Returns:
+ *  Zero on success or a negative error number on failure.
+ */
+int32_t krun_add_formatted_disk(uint32_t ctx_id,
```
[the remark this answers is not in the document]
slp: @jakecorrenti Please rename this to `krun_add_disk2` (all the `...2` functions will be removed in 2.0) and turn `raw` into an explicit format too, like `qcow2`.

src/devices/src/virtio/file_traits.rs
XanClic: This code doesn’t handle the case when the guest writes bytes one by one, so it’s not so useful on its own. It’s also unnecessary for non-raw images.

So I would say this code should only be run if the user hasn’t specified the disk format, i.e. libkrun probed the format, and no format was detected, i.e. the image is used raw. Then, it should reject all writes to the first four byes that are not writing them in full. (Or even the first 512 bytes, in anticipation of later-added formats that need more bytes for identification.)

src/devices/src/virtio/block/device.rs
```diff
@@ -215,6 +216,13 @@ impl Block {
 
         let image_type = disk::detect_image_type(&disk_image, false).unwrap();
 
+        if image_type != disk_image_format {
+            return Err(io::Error::new(
+                ErrorKind::InvalidInput,
+                "image type reported does not match image header",
+            ));
+        }
+
```
XanClic, line 225: I would have expected the user’s choice overrides all possible probing. I.e., if the user specifies an image as raw, it should always be opened as raw regardless of whether there’s a qcow2 magic in the first four bytes.

src/devices/src/virtio/block/disk/mod.rs
slp: You're right. If it comes from crosvm, we need to keep the copyright notice.

src/devices/src/virtio/file_traits.rs
jakecorrenti: This doesn't cover the Raw-only aspect of your comment, though.

XanClic: Ah, right! I missed that.

The check could be incorporated into `guard_header_bytes()` e.g. like so:
```rust
fn guard_header_bytes<'a>(                                                                             
    image: &imago::SyncFormatAccess<imago::file::File>,                                                
    iovec: imago::io_buffers::IoVector<'a>,
    offset: u64,
    head_buffer: &'a mut Option<imago::io_buffers::IoBuffer>,                                                                                                                                                   
) -> Result<imago::io_buffers::IoVector<'a>> {                                                         
    if (1..4).contains(&offset) || (offset + iovec.len() < 4) {                                        
        // Make sure the guest writes to the first four bytes as a whole. This is done to prevent a
        // Raw disk image byte-by-byte, and in any order, writing the header of a formatted disk       
        // image into the first sector of the disk image.                                              
        return Err(Error::new(
            ErrorKind::InvalidInput,
            "partial write to first four bytes of disk image",                                         
        ));
    }

    if offset > 0 {
        return Ok(iovec);
    }

    let (head, tail) = iovec.split_at(std::cmp::max(image.mem_align(), 4) as u64);                     
    *head_buffer = Some(head.try_into_owned(image.mem_align())?);                                      
    let head_slice = head_buffer.as_ref().unwrap().as_ref().into_slice();                              
    if head_slice.get(0..4) == Some(&disk::QCOW_MAGIC.to_be_bytes()) {                                 
        return Err(Error::new(                                                                         
            ErrorKind::InvalidData,                                                                    
            "writing QCOW2 header to disk image",                                                      
        ));                                                                                            
    }                                                                                                  

    Ok(tail.with_inserted(0, head_slice))
}
```

Which would be used as:
```rust
let (iovec, _guard) = imago::io_buffers::IoVector::from_volatile_slice(bufs);
let mut _copied_head = None;
let iovec = guard_header_bytes(self, iovec, offset, &mut _copied_head)?;
```

By the way, I’d also make the whole `impl FileReadWriteAtVolatile for SyncFormatAccess<File>` and the whole `guard_header_bytes()` function conditional on `cfg(feature = "blk")`, not just the `QCOW_MAGIC` portion.

src/devices/src/virtio/block/device.rs
jakecorrenti: Should be fixed.

src/devices/src/virtio/block/device.rs
```diff
@@ -203,10 +204,41 @@ impl Block {
...
+        let disk_image_id = DiskProperties::build_disk_image_id(&disk_image);
+
+        let image_type = disk::detect_image_type(&disk_image, false).unwrap();
+
+        if image_type != disk_image_format {
+            return Err(io::Error::new(
+                ErrorKind::InvalidInput,
+                "image type reported does not match image header",
+            ));
+        }
```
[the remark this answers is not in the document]
jakecorrenti: No, its not. Fixed!

include/libkrun.h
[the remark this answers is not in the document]
jakecorrenti: Should be fixed now. Let me know if that's what you were expecting.

src/devices/src/virtio/file_traits.rs
jakecorrenti: I think I've addressed most of the comments in the latest push. I still need to make it so the header check only happens on Raw files, but I need to figure out the best way to do that.

XanClic: How about implementing `FileReadWriteAtVolatile` for `DiskProperties` instead of `SyncFormatAccess`? Then you could store probing and format information in `DiskProperties` and access it in the trait’s implementation.

jakecorrenti: Ah, yes! That should work. I like that; thanks!

jakecorrenti:
I'm going to work on splitting up the commits. They should be broken up some more and into smaller chunks.

mtjhrc:
I am not sure if `libkrun` should provide the ability to probe the image. I understand that users may implement this incorrectly but @XanClic  mentioned here that the best way to probe the image was before first running it and store the detected format, but `libkrun` is not in a good place to do that, because libkrun is a library that leaves task of storing (or hardcoding) the vm settings up the user of the library (things like number of cores, the attached devices...), which is the logical place to store the probed disk image format before the first run.

include/libkrun.h
```diff
@@ -113,6 +120,41 @@ int32_t krun_set_data_disk(uint32_t ctx_id, const char *disk_path);
...
+#define KRUN_DISK_FORMAT_QCOW2 2
+/**
+ * Adds a disk image to be used as a general partition for the microVM. The supported
+ * image formats are: "raw" and "qcow2". 
+ *
+ * This API is mutually exclusive with the deprecated krun_set_root_disk and
+ * krun_set_data_disk methods and must not be used together.
+ *
+ * By all possible means, one should avoid probing the disk image that will be given to libkrun.
+ * If a raw file is probed, and a QCOW2 header is written into the raw file's header, the next time
```
mtjhrc, line 135: For clarity, I would explicitly mention it is the guest that could override the header.
```suggestion
 * If a raw file is probed, and a QCOW2 header is written into the raw file's header by the guest, the next time
```

include/libkrun.h
```diff
@@ -113,6 +120,41 @@ int32_t krun_set_data_disk(uint32_t ctx_id, const char *disk_path);
...
+#define KRUN_DISK_FORMAT_RAW 1
+#define KRUN_DISK_FORMAT_QCOW2 2
+/**
+ * Adds a disk image to be used as a general partition for the microVM. The supported
+ * image formats are: "raw" and "qcow2". 
+ *
+ * This API is mutually exclusive with the deprecated krun_set_root_disk and
+ * krun_set_data_disk methods and must not be used together.
+ *
+ * By all possible means, one should avoid probing the disk image that will be given to libkrun.
```
mtjhrc, line 134: Since we are worried about users trying to probe the disc, maybe make the doc comment stand out more?
 
```suggestion
 *
 * SECURITY NOTE:
 * By all possible means, one should avoid probing the disk image that will be given to libkrun.
```

mtjhrc, line 134: Since we are worried about users trying to probe the disc, maybe make the doc comment stand out more?
 
```suggestion
 *
 * SECURITY NOTE:
 * By all possible means, one should avoid probing the disk image that will be given to libkrun.
```

include/libkrun.h
```diff
@@ -113,6 +120,41 @@ int32_t krun_set_data_disk(uint32_t ctx_id, const char *disk_path);
...
+ * If a raw file is probed, and a QCOW2 header is written into the raw file's header, the next time
+ * libkrun launches the VM it will boot as a QCOW2 image. QCOW2 files can, and will, open other files.
+ * The guest will most likely have access to the data in those files, essentially giving the guest
+ * access to any file on the host system the disk emulation has access to.
+ *
+ * Arguments:
+ *  "ctx_id"      - the configuration context ID.
+ *  "block_id"    - a null-terminated string representing the partition.
+ *  "disk_path"   - a null-terminated string representing the path leading to the disk image that
+ *                  contains the root file-system.
```
mtjhrc, line 144: This doesn't have to be the "root file-system" though, does it? Since this function replaces `krun_set_data_disk`...

jakecorrenti: Fixed.

include/libkrun.h
jakecorrenti: Good idea. Fixed!

include/libkrun.h
jakecorrenti: Fixed as well

include/libkrun.h
```diff
@@ -113,6 +120,41 @@ int32_t krun_set_data_disk(uint32_t ctx_id, const char *disk_path);
  */
 int32_t krun_add_disk(uint32_t ctx_id, const char *block_id, const char *disk_path, bool read_only);
 
+/* Supported disk image formats */
+#define KRUN_DISK_FORMAT_UNSPECIFIED 0
+#define KRUN_DISK_FORMAT_RAW 1
+#define KRUN_DISK_FORMAT_QCOW2 2
```
tylerfanelli, line 126: It'd probably be safer to make this an `enum`.

include/libkrun.h
```diff
@@ -113,6 +120,41 @@ int32_t krun_set_data_disk(uint32_t ctx_id, const char *disk_path);
...
+ *  "block_id"    - a null-terminated string representing the partition.
+ *  "disk_path"   - a null-terminated string representing the path leading to the disk image.
+ *  "disk_format" - the disk image format
+ *  "read_only"   - whether the mount should be read-only. Required if the caller does not have
+ *                  write permissions (for disk images in /usr/share).
+ *
+ * Returns:
+ *  Zero on success or a negative error number on failure.
+ */
+int32_t krun_add_disk2(uint32_t ctx_id,
```
tylerfanelli, line 152: Although not very nice and desperately needing a change when revamp the API for libkrun 2.0, we *could* just check the file extension for `.qcow2` or `.raw` instead of adding another API.

germag: I don't think that is a good idea, it's like probing but worse. Also, extensions means nothing in a unix world, for instance, I have many qcow2 and raw images sharing the same extension `.img` 


>> @mtjhrc wrote: I am not sure if libkrun should provide the ability to probe the image.

> @jakecorrenti wrote: To me this feels like we are putting a lot of trust on the user to do the right thing.

What I suggested, is to spilt the API into probing and adding so you will have something like:
```
int32_t krun_add_disk2(..., format format_type)  // Add a disk with a specified format
format_type krun_detect_format(...)  // Returns the format
```
(maybe detect_format should be part of imago?)

So, the user has the flexibility to do probing or strictly only accepting a disk + its format

Sadly this API is not perfect, It could be possible some TOCTOU (I cannot think of any scenario right now), but it can be solved and while having a split functionality

germag:
> > I am not sure if `libkrun` should provide the ability to probe the image. I understand that users may implement this incorrectly but @XanClic mentioned here, that the best way to probe the image is before first running it and store the detected format, but `libkrun` is not in a good place to do that, because libkrun is a library that leaves task of storing (or hardcoding) the vm settings up the user of the library (things like number of cores, the attached devices...), which is the logical place to store the probed disk image format before the first run.
> 
> To me this feels like we are putting _a lot_ of trust on the user to do the right thing.

I have mixed "feelings" about this, I'm in favor that an application should prevent the user to do bad things (in general not as a hard rule), but I'm not sure a _library_ should do that, I think we should provide the means to do "the right thing" but forcing the user. Also as a general rule, and it depends on the case, for example I think preventing writing a qcow2 header to raw disk is ok, the only drawback is it will prevent that specific use case: "the user wants to create a qcow2 image in a block device from inside a VM". But, probably is such a rare case that we can just wait until someone ask for it.


> 
> If the user probes, and something happens, they're probably going to come to us with "What happened?". Is that something we're ok with?

include/libkrun.h
XanClic: (Sorry, this is one of my comments that needs headlines to separate my rambling…)

Agreed with the comment on `.img`. (I think `.img` specifically was the reason why we never pursued using filename extensions in qemu.)

### Split probing functionality (format detection vs. opening)

Certainly imago could provide a `detect_format`, though asking that actually leads to another topic in itself.[1]

Providing such a split functionality in libkrun has the downside that libkrun won’t know whether the image format was probed or not, so the safeguards preventing malicious guests from formatting raw images need to either be on or off for all raw images all the time, regardless of whether the format was probed or not. (Of course, the user could enable/disable these safeguards with a flag, but then you can also just allow users to pass `NULL`/`None` as the format to make it probe.)

As you said, this would fully prevent guests e.g. from creating qcow2 images on raw block devices, if that’s what someone wants. (FWIW, may be a nested virtualization case. There are use cases of qcow2 on LVM.)

### To probe or not to probe

That said, I don’t want to sound like I’m in favor of probing. It’s proven itself to be a bad practice. But I do assume it’s a reality that must be lived with.

I’m afraid that if libkrun doesn’t provide a usable probing interface, people will find ways to do their own probing, which will be worse. Consider that imago doesn’t provide a probing interface, so this pull request implements it in libkrun.

I don’t think it’s unreasonable to decide to have libkrun not provide any probing, so it doesn’t get its hands dirty, while putting big fat warnings into its documentation, which may prevent its users from doing the wrong things, too. I just don’t have enough experience to know how effective such warnings really are in practice. (I should have done this for imago’s documentation, to see whether it works, as a social experiment. 😄)

### Probing in imago

[1] Putting any probing functionality into imago would beg some further questions:
- Returning an `enum` doesn’t work, because formats may be added, breaking compatibility. An opaque object with `.is_qcow2()` etc. would probably be best. (Could also be further extended to general image information.)
  - Honestly we’ll probably need some format representation for the “Is it OK to open this file?” callback anwyay.
- Why not allow imago to auto-open images in any format then? (Note: Already does this with backing files.)
- “If anyone does probing, it should be libkrun, because it controls the guest’s I/O path” applies to imago, too – why not move those safeguards into imago then? That would also save libkrun from duplicating a bit of probing code, specifically detecting the qcow2 magic sequence.

So really, same questions as we now have here for libkrun (“should it probe?”, “how to provide probing functionality?”, “should there be safeguards for probed raw images?”). I think any layer that were to implement probing would need to answer these questions.

Imago already has probing for qcow2 backing files, in case their format isn’t given in the qcow2 image. That’s not such a hot topic, though, because `qemu-img` these days requires users to put the backing format there; backing files are generally read-only; and so the only way to mess this up is basically by having the guest write e.g. a qcow2 image header to a raw format, and then taking a snapshot (so it becomes a backing file), without writing “raw” into the overlay’s backing format field (which I would consider quite wrong not to do).

Now, while the questions for how to probe are probably always the same regardless of on which layer you do it, maybe it’s easier to answer those questions if we consider probing were done in imago. Then, libkrun would be the user of this probing code, not the provider, so maybe it’s easier to know what you’d expect from it than to guess what users might expect from libkrun.

include/libkrun.h
mtjhrc: Code style wise sure, this could be nicer. But how it is safer?

If we also change the `krun_add_disk2` argument to be the enum type this could get more unsafe on the ABI level. Because theoretically, you cannot control what backing type the enum is in C, which complicates the ffi on the rust side (or other language for that matter). I am not sure how likely this is to be a weird type and using an `i32` on the Rust side is very likely safe, but not guaranteed.

Relevant quotes in C spec: https://stackoverflow.com/a/1113869

XanClic:
> I think we should provide the means to do "the right thing" instead of forcing the user.

I think if you don’t provide functionality users want (probing), they’ll work around it. Yes, it’s on them if they don’t follow the documentation and it goes wrong (re-probing raw images guests had write access to).

But if you can provide something that users would happily take instead of working around it, and that is reasonably safe, it should be considered. To me, the only question is, can we?

germag:
Ok, just to move this forward, and since the current users of libkrun will not need probing and the APIl will be reworked for v2.0 soonish. What about just provide `krun_add_disk2()` keeping `krun_add_disk()` just for raw images, and not probing at all.

While working on providing a sane and safe API for v2.0, so we can see what should be done in libkrun and/or imago, wdyt?

XanClic:
> […] What about just provide `krun_add_disk2()` keeping `krun_add_disk()` just for raw images, and not probing at all.

Seems good to me.

include/libkrun.h
mtjhrc: > * Returning an `enum` doesn’t work, because formats may be added, breaking compatibility. An opaque object with `.is_qcow2()` etc. would probably be best. (Could also be further extended to general image information.)
Can't you use `#[non_exhaustive]` for this? This is what `ErrorKind` in the standard library uses.

> Why not allow imago to auto-open images in any format then? (Note: Already does this with backing files.)
At that point we could, but it would be good to have the callback you mentioned for implementing an allow/deny list of backing files. I think in the case of libkrun consumers it should be reasonable for them to have an explicit allow list (probably through a callback is fine, so this could even be used by an interactive application).

XanClic: > Can't you use #[non_exhaustive] for this?

True!

> At that point we could, but it would be good to have the callback you mentioned for implementing an allow/deny list of backing files.

Yes. I can work on it.

jakecorrenti:
I just want to do a quick recap to make sure I'm getting everything:
- We still need to come to a conclusion about probing. Does imago do it? Libkrun? Or push it onto the libkrun consumer?
- To move this forward, keep `krun_add_disk` for adding Raw files. Use `krun_add_disk2` to add Qcow2 files. Don't do any probing right now (essentially getting rid of commit fd6203f955922605fb6f6cf84f68b1442470a78b)  and get rid of the `KRUN_DISK_FORMAT_{UNSPECIFIED, RAW}` constants.
- We should come up with a safe and sane API that will be ready for libkrun 2.0. This means investigating what should be done in libkrun and/or imago.

slp:
> I just want to do a quick recap to make sure I'm getting everything:
> 
> * We still need to come to a conclusion about probing. Does imago do it? Libkrun? Or push it onto the libkrun consumer?

Neither libkrun nor imago (on behalf of libkrun) should do any probing. There's no reasonable use case for that, and the risks are significant (this was a hard-learnt lesson in QEMU). It's up to the library consumers to explicitly tell us the image format.

> * To move this forward, keep `krun_add_disk` for adding Raw files. Use `krun_add_disk2` to add Qcow2 files. Don't do any probing right now (essentially getting rid of commit [fd6203f](https://github.com/containers/libkrun/commit/fd6203f955922605fb6f6cf84f68b1442470a78b))  and get rid of the `KRUN_DISK_FORMAT_{UNSPECIFIED, RAW}` constants.

Please remove UNSPECIFIED but keep RAW. There may be users preferring to always use `krun_add_disk2` for consistency, and also makes the transition to 2.0 easier for us.

(BTW, thanks for this work!)

include/libkrun.h
```diff
@@ -95,16 +95,23 @@ int32_t krun_set_root_disk(uint32_t ctx_id, const char *disk_path);
...
  *
  * This API is mutually exclusive with the deprecated krun_set_root_disk and
  * krun_set_data_disk methods and must not be used together.
  *
+ * SECURITY_NOTE:
+ * By all possible means, one should avoid probing the disk image that will be given to libkrun.
+ * If a raw file is probed, and a QCOW2 header is written into the raw file's header, the next time
+ * libkrun launches the VM it will boot as a QCOW2 image. QCOW2 files can, and will, open other files.
+ * The guest will most likely have access to the data in those files, essentially giving the guest
+ * access to any file on the host system the disk emulation has access to.
```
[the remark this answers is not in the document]
jakecorrenti: Nice catch! I just fixed it in my latest push of changes

src/devices/src/virtio/block/disk/mod.rs
jakecorrenti: @slp As of my latest push, `disk/mod.rs` just contains `QCOW_MAGIC` and `ImageType`. Does that still justify the copyright notice, or can I remove it?

include/libkrun.h
tylerfanelli: > I don't think that is a good idea, it's like probing but worse. Also, extensions means nothing in a unix world, for instance, I have many qcow2 and raw images sharing the same extension `.img`

If that's standard practice to have qcow2 images with `.img` then yes my suggestion is a very bad idea.

jakecorrenti:
Hi everyone, thank you for all of the comments. I pushed a set of changes that should address everything. PTAL when you get a chance.

src/devices/src/virtio/file_traits.rs
```diff
@@ -460,3 +465,54 @@ impl FileReadWriteAtVolatile for DiskProperties {
...
+    *head_buffer = Some(head.try_into_owned(image.mem_align())?);
+    let head_slice = head_buffer.as_ref().unwrap().as_ref().into_slice();
+
+    if head_slice.get(0..4) == Some(&disk::QCOW_MAGIC.to_be_bytes()) {
+        return Err(Error::new(
+            ErrorKind::InvalidData,
+            "writing QCOW2 header to disk image",
+        ));
+    }
+
```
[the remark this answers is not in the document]
jakecorrenti: Ok, noted. I wasn't sure if I should have kept it or not, and just erred on the side of caution. I'll remove it

src/devices/src/virtio/file_traits.rs
```diff
@@ -460,3 +465,54 @@ impl FileReadWriteAtVolatile for DiskProperties {
...
+) -> Result<IoVector<'a>> {
+    if (1..4).contains(&offset) || (offset + iovec.len() < 4) {
+        // Make sure the guest writes to the first four bytes as a whole. This is done to prevent a
+        // Raw disk image byte-by-byte, and in any order, writing the header of a formatted disk
+        // image into the first sector of the disk image.
+        return Err(Error::new(
+            ErrorKind::InvalidInput,
+            "partial write to first four bytes of disk image",
+        ));
+    }
```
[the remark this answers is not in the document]
jakecorrenti: If this is impossible, and the mitigation below is also irrelevant, then it seems like I'll just remove 46f797e0ed4b0d16308ac0810b3d9802a672faac altogether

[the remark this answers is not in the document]
jakecorrenti: Dropped the last 2 commits. Should hopefully be up to date now.

src/devices/src/virtio/block/disk/mod.rs
```diff
@@ -0,0 +1,13 @@
...
+/// QCOW magic constant
+pub const QCOW_MAGIC: u32 = 0x5146_49fb;
```
XanClic, line 6: This is now unused.

include/libkrun.h
```diff
@@ -95,16 +95,23 @@ int32_t krun_set_root_disk(uint32_t ctx_id, const char *disk_path);
...
+ * format is "raw".
  *
  * This API is mutually exclusive with the deprecated krun_set_root_disk and
  * krun_set_data_disk methods and must not be used together.
  *
+ * SECURITY NOTE:
+ * By all possible means, one should avoid probing the disk image that will be given to libkrun.
+ * If a raw file is probed, and a QCOW2 header is written into the raw file's header by the guest,
+ * the next time libkrun launches the VM it will boot as a QCOW2 image. QCOW2 files can, and will,
+ * open other files. The guest will most likely have access to the data in those files, essentially
```
XanClic, line 108: Given that libkrun will no longer probe, it’s not automatically true that “the next time libkrun […] will boot as a QCOW2 image”. This is only true if the user were to re-probe the image, and then tell libkrun that it’s a qcow2 image now.

Also, this function will only ever open the image as raw; I think a comment about image formats, probing, and why this doesn’t probe makes absolute sense, but this comment as-is no longer really fits it.

germag:
LGTM

XanClic (commented):
One general thing: I’d switch commits 3 and 4. Commit 3 introduces the external interface for qcow2 images, but internally, the format is actually ignored. So the internal capability to handle qcow2 images should come before.

I know, it technically doesn’t matter, as long as nobody looks between commits 3 and 4.

src/devices/src/virtio/file_traits.rs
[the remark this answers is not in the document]
XanClic: (By the way, by “impossible” I didn’t mean it’d be impossible to prevent byte-by-byte writes by the guest. Quite the opposite, I meant that it’s impossible for the guest to do byte-by-byte writes, so it wouldn’t be necessary for us to detect them.)

include/libkrun.h
jakecorrenti: Updated the comments. Let me know what you think!

jakecorrenti:
I still need to swap commits 2 and 3 which requires a couple changes, but I wanted to push everything else first.

include/libkrun.h
jakecorrenti: I missed some capitalization mistakes, which I will fix in the next push.

jakecorrenti:
Squashed a commit, reordered them, and updated the commit messages. Let me know if there's anything else :)

include/libkrun.h
```diff
@@ -113,6 +122,49 @@ int32_t krun_set_data_disk(uint32_t ctx_id, const char *disk_path);
...
+ * image formats are: "raw" and "qcow2".
+ *
+ * This API is mutually exclusive with the deprecated krun_set_root_disk and
+ * krun_set_data_disk methods and must not be used together.
+ *
+ * SECURITY NOTE:
+ * Libkrun is not responsible for probing the disk image provided, as probing a disk image should
+ * be avoided. However, if one must probe the disk image, it falls onto the user. The disk image
+ * must only be probed once, and the probe of the image must be done before the first boot of the
+ * microVM.
```
XanClic, line 139: I find it too easy to read the first part and get the idea that probing “‘is to be avoided’, but, well, avoiding doesn’t mean ‘must not’, right?”

Also, the concrete problem is buried below, and it reads a bit like qcow2 images referencing other files is only a problem if the guest modified a raw image to do so; but actually it’s a general security concern, for any image that isn’t fully trusted, so probably should be the very first thing mentioned.

(I didn’t really want to write a full proposal, but it kind of happened accidentally, so… What would you think about the following?)

___

“Non-Raw images can reference other files, which libkrun will automatically open, and to which the guest will have access. Libkrun should therefore never be asked to open an image in a non-Raw format when it does not come from a fully trustworthy source.

Consequently, probing an image’s format is quite dangerous and to be avoided if at all possible, which is why libkrun provides no facilities for doing so. If it is not clear what format an image has, it may also not be clear whether it can be trusted not to reference files to which the guest should not have access.

If probing absolutely cannot be avoided, it must only be done on images that are fully trusted, i.e. before a potentially untrusted guest had write access to it. Specifically, consider that a guest has full access to all of a Raw image, and can therefore turn it into a file in an arbitrary format, for example, into a Qcow2 image, referencing and granting a malicious guest access to arbitrary files. To hand a Raw image to an untrusted and potentially malicious guest, and then to re-probe it after the guest was able to write to it (when it can no longer be trusted), would therefore be a severe security vulnerability.

Therefore, after having probed a yet fully trusted image once, the result must be remembered so the image will from then on always be opened in the format that was detected originally. When adhering to this, a guest can write anything they want into a Raw image, it is always going to be opened as a Raw image, preventing the security vulnerability outlined above.

However, if at all possible, the image format should be explicitly selected based on knowledge obtained separately from the pure image data, for example by the user.”

XanClic (commented):
Besides the comments, looks good to me, thanks!

include/libkrun.h
XanClic: My suggestion may well be (much?) too long, but I think it is important to be specific about the exact problem (formatted images give guests access to other files) so that users always consider this in all of its implications (i.e. not only get format detection right, but also ensure that even a freshly downloaded image won’t contain malicious file references in the first place).

jakecorrenti: Just added your suggestions

src/devices/Cargo.toml
```diff
@@ -34,6 +34,8 @@ utils = { path = "../utils" }
 polly = { path = "../polly" }
 rutabaga_gfx = { path = "../rutabaga_gfx", features = ["virgl_renderer", "virgl_renderer_next"], optional = true }
 
+imago = { version = "0.1.0", features = ["sync-wrappers", "vm-memory"] }
+
```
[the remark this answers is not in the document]
tylerfanelli: Yea, it's compatible with any `0.1._`, I *think* it's as simple as running `cargo clean` and `cargo clippy` again to generate a new `Cargo.lock`.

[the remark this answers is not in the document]
germag: Nop, `cargo clean + clippy` will not recreate `Cargo.lock`, you can just delete the lock file, or `cargo update`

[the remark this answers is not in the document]
jakecorrenti: I could update the Cargo.toml to be ">= 0.1" so that for future bugfixes we don't need to keep updating the Cargo.toml file, just the Cargo.lock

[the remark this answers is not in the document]
jakecorrenti: Ah, you're right. Today's a Public Holiday in the states, but if I get a chance I'll push the changes.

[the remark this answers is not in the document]
jakecorrenti: All fixed!

jakecorrenti:
@slp  PTAL

jakecorrenti:
New Code Quality failures are interesting considering no changes to the code were made... Something I should worry about in this PR?

germag:
> New Code Quality failures are interesting considering no changes to the code were made... Something I should worry about in this PR?

It seems it's in a file you didn't touched, you can open a new PR to change this:
```
impl<'a> HvfVcpu<'a> {
```
to this
```
impl HvfVcpu<'_> {
```

in `src/hvf/src/lib.rs:250`

slp (approved):
@jakecorrenti Don't worry about the CI issues, I'll fix them in a different PR. Let's merge this one already.

slp:
@jakecorrenti @XanClic @tylerfanelli @germag Now, the one million question: who volunteers to package imago for Fedora? ;-)

germag:
> > Now, the one million question: who volunteers to package imago for Fedora? ;-)
> 
> 🙈 I’ll go figure it out.

I can help you with that

PR #238, opened 2024-11-13
Update SEV deps to latest versions

slp:


slp:
@tylerfanelli PTAL

tylerfanelli (approved):
LGTM, but the sev crate's 4.0.0 will not run on upstream kernels. We're looking to update to 5.0.0 now that 6.11 brought "stable" support. Perhaps we should wait to update libkrun until sev-5.0.0 is up? Shouldn't be much longer

slp:
> LGTM, but the sev crate's 4.0.0 will not run on upstream kernels. We're looking to update to 5.0.0 now that 6.11 brought "stable" support. Perhaps we should wait to update libkrun until sev-5.0.0 is up? Shouldn't be much longer

We need it to be in sync with the status of the packages in Fedora to be able to produce new releases. We can do another omnibus updates once 5.0.0 is released.

tylerfanelli:
SGTM.

PR #243, opened 2024-12-12
Update SEV-SNP implementation to run on upstream kernels

tylerfanelli:
Modify the SEV-SNP launch implementation to:

- Create VMs with the KVM_X86_SNP_VM type.
- Use KVM guest_memfd for memory regions.
- Use KVM_SET_USER_MEMORY_REGION2.
- Update the sev library dependency to 5.0.0.

With this, libkrun should be able to run on stock Linux kernels with version >= 6.11. Still completing some testing. Once tested, I will mark this PR as ready for review.

### DIFF
--- a/src/vmm/src/linux/vstate.rs
+++ b/src/vmm/src/linux/vstate.rs
@@ -49,7 +49,10 @@ use kvm_bindings::{
 };
 
 #[cfg(feature = "amd-sev")]
-use kvm_bindings::{kvm_create_guest_memfd, kvm_userspace_memory_region2, KVM_MEM_GUEST_MEMFD};
+use kvm_bindings::{
+    kvm_create_guest_memfd, kvm_memory_attributes, kvm_userspace_memory_region2,
+    KVM_MEM_GUEST_MEMFD,
+};
 
 use kvm_bindings::KVM_API_VERSION;
 use kvm_ioctls::*;
@@ -118,6 +121,9 @@ pub enum Error {
     SetUserMemoryRegion(kvm_ioctls::Error),
     /// Error creating memory map for SHM region.
     ShmMmap(io::Error),
+    #[cfg(feature = "amd-sev")]
+    /// Error setting KVM memory attributes of a region.
+    SetMemoryAttributes(kvm_ioctls::Error),
     #[cfg(feature = "amd-sev")]
     /// Error initializing the Secure Virtualization Backend (SEV).
     SevSecVirtInit(SevError),
@@ -278,6 +284,10 @@ impl Display for Error {
             ),
             SetUserMemoryRegion(e) => write!(f, "Cannot set the memory regions: {e}"),
             ShmMmap(e) => write!(f, "Error creating memory map for SHM region: {e}"),
+            #[cfg(feature = "amd-sev")]
+            SetMemoryAttributes(e) => {
+                write!(f, "Error setting KVM memory attributes of a region: {e}")
+            }
             #[cfg(feature = "tee")]
             SevSecVirtInit(e) => {
                 write!(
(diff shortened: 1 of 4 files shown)


tylerfanelli:
With these changes, I'm encountering the following unexpected error immediately once the VM is started:
```
Starting TEE/microVM.
[2024-12-12T06:43:33Z ERROR vmm::linux::vstate] Unexpected exit reason on vcpu run: MemoryFault { flags: 8, gpa: 4096, size: 4096 }
```
Will investigate.

tylerfanelli:
I'd like to modify this to create *one* guest_memfd and map all guest memory regions into it (using the `kvm_userspace_memory_region2.guest_memfd_offset` field) rather than creating a new guest_memfd for each region of guest memory.

src/vmm/src/linux/vstate.rs
```diff
@@ -579,6 +591,42 @@ impl Vm {
             self.next_mem_slot += 1;
         }
 
+        #[cfg(feature = "amd-sev")]
```
jakecorrenti, line 594: IMO we should feature flag this on `#[cfg(feature = "tee")]` instead. Both Mathias and I's PRs are doing this exact same thing. 

In hindsight it may have been better for us to do that as a separate PR to just get the functionality into main knowing all the CC architectures were moving in that direction.

jakecorrenti (changes requested):

jakecorrenti:
> With these changes, I'm encountering the following unexpected error immediately after the VM starts:
> 
> ```
> Starting TEE/microVM.
> [2024-12-12T06:43:33Z ERROR vmm::linux::vstate] Unexpected exit reason on vcpu run: MemoryFault { flags: 8, gpa: 4096, size: 4096 }
> ```
> 
> Will investigate.

I know in TDX you need to also set memory attributes after creating the guest memfd and setting the user memory region. I wonder if you have to do that here as well?

tylerfanelli:
> > With these changes, I'm encountering the following unexpected error immediately after the VM starts:
> > ```
> > Starting TEE/microVM.
> > [2024-12-12T06:43:33Z ERROR vmm::linux::vstate] Unexpected exit reason on vcpu run: MemoryFault { flags: 8, gpa: 4096, size: 4096 }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Will investigate.
> 
> I know in TDX you need to also set memory attributes after creating the guest memfd and setting the user memory region. I wonder if you have to do that here as well?

Do you? I wasn't aware of this. Will try it out.

src/vmm/src/linux/vstate.rs
tylerfanelli: Agreed, but I don't think it fits in this PR simply because there is no other TEE arch currently supported. PRs adding TEE arch support should rebase on this and change the `#[cfg(feature = "amd-sev")]` to
`#[cfg(feature = "tee")]`

jakecorrenti:
> > > With these changes, I'm encountering the following unexpected error immediately after the VM starts:
> > > ```
> > > Starting TEE/microVM.
> > > [2024-12-12T06:43:33Z ERROR vmm::linux::vstate] Unexpected exit reason on vcpu run: MemoryFault { flags: 8, gpa: 4096, size: 4096 }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Will investigate.
> > 
> > 
> > I know in TDX you need to also set memory attributes after creating the guest memfd and setting the user memory region. I wonder if you have to do that here as well?
> 
> Do you? I wasn't aware of this. Will try it out.

```rust
let attr = kvm_memory_attributes {
    address: region.start_addr().raw_value(),
    size: region.len() as u64,
    // KVM_MEMORY_ATTRIBUTE_PRIVATE,
    attributes: 1 << 3,
    flags: 0,
};

self.fd
    .set_memory_attributes(attr)
    .map_err(Error::SetMemoryAttributes)?;
```

Should do the job. Hopefully this saves you time and avoids you having to dig around for it.

tylerfanelli:
Closing this in favor of an upcoming PR that adds update SEV-SNP and KVM guest_memfd support.

PR #244, opened 2024-12-15
virtio/fs: Check capabilities & act accordingly

asahilina:
Two changes that make the passthrough layer check whether it runs with the proper capabilities to do something (i.e. root) and, if not, make it not do it.

Needed for some experiments I'm trying with the FEX RootFS stuff, but also generally helpful (I don't expect anything else to be needed in libkrun, so I'm sending this early).


slp (approved):
LGTM, thanks!

PR #246, opened 2024-12-31
net: Allow creating vsocks in listen mode.

WhatAmISupposedToPutHere:
Previously all vsocks were expected to be connected from the guest. This patch adds support for sockets that are listened to by the guest and are connected from the host.

### DIFF
--- a/src/libkrun/src/lib.rs
+++ b/src/libkrun/src/lib.rs
@@ -95,7 +95,7 @@ struct ContextConfig {
     data_block_cfg: Option<BlockDeviceConfig>,
     #[cfg(feature = "tee")]
     tee_config_file: Option<PathBuf>,
-    unix_ipc_port_map: Option<HashMap<u32, PathBuf>>,
+    unix_ipc_port_map: Option<HashMap<u32, (PathBuf, bool)>>,
     shutdown_efd: Option<EventFd>,
     gpu_virgl_flags: Option<u32>,
     gpu_shm_size: Option<usize>,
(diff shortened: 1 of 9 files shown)


tylerfanelli (commented):
Any readily-available testing methods?

WhatAmISupposedToPutHere:
Build the linked PR from muvm, and launch a second command in the same vm. The second one will use this feature to send the launch request.

src/libkrun/src/lib.rs
```diff
@@ -933,6 +933,17 @@ pub unsafe extern "C" fn krun_add_vsock_port(
...
+
+#[allow(clippy::missing_safety_doc)]
+#[no_mangle]
+pub unsafe extern "C" fn krun_add_vsock_port2(
+    ctx_id: u32,
+    port: u32,
+    c_filepath: *const c_char,
+    listen: bool,
 ) -> i32 {
     let filepath = match CStr::from_ptr(c_filepath).to_str() {
```
slp, line 948: When `listen == true`, please also check if the path exists, failing if that's the case. It'd be also nice checking if the dirname is okay.

This allows us to return an early error to the library user, as otherwise can be a bit confusing.

slp (changes requested):
This looks great, thanks. I only requested a minor change to force the library users to provide a healthy path for the unix socket.

src/libkrun/src/lib.rs
mtjhrc: I think `krun_add_vsock_port2` with `listen = true` could just `bind` the socket before returning. This would more robustly verify the path is in fact a socket and not something else and just save the socket `fd` in the `unix_ipc_port_map`.

mtjhrc:
Thanks, I'll try to test this later, but for now I realized something:

If I am not mistaken, this only starts listening on the socket once the vsock device is activated, so if you want to connect to the socket from host, you would have to wait until the guest boot process initializes the vsock device. 
Is this intended or useful for your use case? If so It should at least be documented, because the users of libkrun could easily run into a race condition where they sometimes try to connect to the socket before the guest boots. An alternative could be to also call `listen` on the socket before `krun_add_vsock_port2` returns.

WhatAmISupposedToPutHere:
Yes, the socket file appearing only after the vm is ready to process incoming connection is desired, that way the vm can wait for the file to appear before sending the requests.

slp:
@mtjhrc Do you want to take a second look or can we merge it already?

mtjhrc (approved):
Sorry for the delay. I managed to test this just by adding a listening vsock in `chroot_vm` and using it in the guest. 
Now I understand how it's meant to be used. Still I am unsure if we generally want to bind the sockets at the device initialization stage, but if It works for you this is fine for now, and we could always add another API later.

So for me this is good to merge! Thanks

slp (approved):

PR #258, opened 2025-01-27
Introduce an E2E testing framework

mtjhrc:
This PR adds a testing framework for writing (mostly) end-to-end tests using libkrun. 

The tests are E2E in the sense that they use the public API of libkrun to start a VM and run a test program in the guest. Note that currently this is very limited, because there are no other userspace executables or libraries (not even libc) in the guest apart from `guest-agent`. `guest-agent` is a statically linked Rust executable, that executes the guest part of each test.
    
In the future this can be extended to run tests with specific images which would allow the use of system libraries by the test program.
    
This also introduces 2 tests:
 - **`test-vm-config`** - asserts the VM is constructed with the correct number of CPUs and amount of memory. This also serves as an example how tests can be parameterized.
- **`vsock-guest-connect`** - This tests the guest connecting to a vsock port created by the `krun_add_vsock_port` API.

You can run these tests using `make test`. 

Note: Currently these tests won't be ran on CI, because we need a CI that supports nested virtualization.


Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [workspace]
-members = ["src/libkrun"]
+members = ["krun-sys", "src/libkrun"]
```
slp, line 2: `krun-sys` is published in crates.io and packaged in Fedora. Can we consume it from there?

mtjhrc: Oh, I didn't know that. But where is the repository for that? I think we should consume it locally and publish the version located here in `crates.io`. Is it the one in `muvm`? 

Wouldn't the crates.io version depend on system wide installation of libkrun? (something we want to avoid here).

slp: > Oh, I didn't know that. But where is the repository for that? Is it the one in `muvm`? I think we should consume it locally and publish the version located here in `crates.io`

Yes, it's that one. Perhaps it should be moved here, perhaps it should have its own repo in github.com/containers, perhaps it can stay in muvm but we should definitely consume it from crates.io.

> Wouldn't the crates.io version depend on system wide installation of libkrun? (something we want to avoid here).

In its current state, yes. I'm going to create a PR to use `pkg-config` to find out the location of libkrun. That should allow us to use the local installation from `test-prefix`.

tests/test_cases/src/test_vm_config.rs
```diff
@@ -0,0 +1,78 @@
...
+            }
+        }
+        panic!("MemTotal field not found");
+    }
+
+    impl Test for TestVmConfig {
+        fn in_guest(self: Box<Self>) {
+            assert_eq!(detect_num_cpus(), self.num_cpus as u32);
+
+            let ram_avalible = detect_ram_size_mib();
```
slp, line 71: Typo `available`.

tests/test_cases/src/test_vm_config.rs
```diff
@@ -0,0 +1,78 @@
...
+            // Check if ram within 5percent of specifed
+            assert!(self.ram_mib >= (ram_avalible as f64 * 0.95) as u32);
```
slp, line 73: This fails on aarch64 for the 256MiB case. We might need to relax the check to 15% or so for the 256MiB test.

slp:
This is looking good, nice work here. There a couple of very small things to fix, and consume `krun-sys` from crates.io (this requires a new version of `krun-sys`.

Cargo.toml
mtjhrc: What would be the advantage of using it from `crates.io` if it was in this repository here? In that case I think it makes more sense for the tests to use the potentially unpublished and potentially broken newest version of `krun-sys` (though admittedly since everything is auto-generated it is not that important).

Makefile
```diff
@@ -136,3 +136,10 @@ install: libkrun.pc
 clean:
 	rm -f $(INIT_BINARY)
 	cargo clean
+	rm -rf test-prefix
+	cd tests; cargo clean
+
+test: $(LIBRARY_RELEASE_$(OS))
+	mkdir -p test-prefix
+	PKG_CONFIG_PATH=test-prefix/lib64/pkgconfig/ PREFIX="$$(realpath test-prefix)" make install
```
slp, line 144: Please use `realpath` for `PKG_CONFIG_PATH` too.

Makefile
```diff
@@ -136,3 +136,10 @@ install: libkrun.pc
...
+	cd tests; LD_LIBRARY_PATH=../test-prefix/lib64/ PKG_CONFIG_PATH=../test-prefix/lib64/pkgconfig/ ./run.sh
```
slp, line 145: Use `realpath` for `../test-prefix` here too. This fixes a problem with non-system-wide libkrun installations.

slp:
I've just published a new version of krun-sys, 1.10.1, which combined with #270 should help us avoid vendoring the crate.

mtjhrc:
I've changed it to use the new version of krun-sys, all the env variables should now use realpath now.
Even though the pkg-config has the path to the library now, we still need `LD_LIBRARY_PATH` for runtime though (otherwise it doesn't find the `libkrun.so.1`). We could also specify `-rpath` in the pkg-config too, but I don't know if this is a good idea.

slp (approved):
LGTM, thanks!

PR #261, opened 2025-01-28
devices/console: Fix a bug which would cause libkrun to hang on exit

mtjhrc:
This fixes a hang observed when the VM is about to exit. The problem was the console device was joining a thread on vm shutdown (or device reset) which would never exit.

To replicate the hang, the stdin must not be a terminal (for example you can use a pipe).
Before this fix, this would hang:
`echo hello | ./chroot_vm rootfs_fedora /bin/cat`
(Btw libkrun used 100% (1 core) of CPU when it was stuck in this state - I don't know why exactly.)

This seems to be a regression introduced by the commit: `4076b7`: devices/console: implement reset method

### DIFF
--- a/src/devices/src/virtio/console/port_io.rs
+++ b/src/devices/src/virtio/console/port_io.rs
@@ -226,8 +226,13 @@ impl PortInput for PortInputSigInt {
         Ok(1)
     }
 
-    fn wait_until_readable(&self, _stopfd: Option<&EventFd>) {
-        let mut poll_fds = [PollFd::new(self.sigint_evt.as_raw_fd(), PollFlags::POLLIN)];
+    fn wait_until_readable(&self, stopfd: Option<&EventFd>) {
+        let mut poll_fds = Vec::with_capacity(2);
+        poll_fds.push(PollFd::new(self.sigint_evt.as_raw_fd(), PollFlags::POLLIN));
+        if let Some(stopfd) = stopfd {
+            poll_fds.push(PollFd::new(stopfd.as_raw_fd(), PollFlags::POLLIN));
+        }
+
         poll(&mut poll_fds, -1).expect("Failed to poll");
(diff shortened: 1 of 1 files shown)


src/devices/src/virtio/console/port_io.rs
```diff
@@ -226,8 +226,13 @@ impl PortInput for PortInputSigInt {
         Ok(1)
     }
 
-    fn wait_until_readable(&self, _stopfd: Option<&EventFd>) {
-        let mut poll_fds = [PollFd::new(self.sigint_evt.as_raw_fd(), PollFlags::POLLIN)];
+    fn wait_until_readable(&self, stopfd: Option<&EventFd>) {
+        let mut poll_fds = Vec::new();
```
bilelmoussaoui, line 230:
```suggestion
        let mut poll_fds = Vec::with_capacity(2);
```
maybe?

mtjhrc: Sure.

(Side note/rant: The performance doesn't really matter here, but it's a shame Rust has no _**easy**_ way to just stack allocate a fixed capacity vector using only the standard library, which would be the best choice in situations like this :disappointed: )

tylerfanelli: Does it need to be a vector? If the size is fixed, is `[PollFd; 2]` stack allocated?

tylerfanelli: That's fair, I'm mainly just commenting on the heap allocation of `Vec`. Sure, there'd be some extra logic involved (perhaps you could also try `[Option<PollFd>; 2]` but that also requires some logic in the handlers) but it may give you the stack allocation you're looking for.

slp (approved):
Confirmed to fix the issue here. Thanks!

PR #263, opened 2025-02-04
Connect to gvproxy socket

nirs:
Connecting to the socket set the remote address so we can use 
send()/recv() or write()/read() instead of sendto()/recvfrom().

Advantages:
- Simpler code, no need to keep the remote address.
- The server will want to connect to the client address to ensure it
  receives frames only from the connected client. Such server will want
  to remove the unix socket once the client connected[2], which doe snot
  work with current code.
- Once the socket is connected, the same backend can be used to handle
  passed file descriptor[1].
- iperf3 -R is 1.33 times faster (46.6 Gbits/s vs 35.0 Gbits/s).

Tested with:
- [x] gvproxy
- [x] vmnet-helper

For testing results see 
https://github.com/containers/libkrun/issues/264#issuecomment-2640338792

[1] https://github.com/containers/krunkit/issues/24
[2] https://github.com/nirs/vmnet-helper/blob/5c6a595ba3e76314e1d0bef2b0160388439d69ec/helper.c#L475

### DIFF
--- a/src/devices/src/virtio/net/gvproxy.rs
+++ b/src/devices/src/virtio/net/gvproxy.rs
@@ -1,7 +1,7 @@
 use nix::fcntl::{fcntl, FcntlArg, OFlag};
 use nix::sys::socket::{
-    bind, getsockopt, recv, sendto, setsockopt, socket, sockopt, AddressFamily, MsgFlags, SockFlag,
-    SockType, UnixAddr,
+    bind, connect, getsockopt, recv, send, setsockopt, socket, sockopt, AddressFamily, MsgFlags,
+    SockFlag, SockType, UnixAddr,
 };
 use nix::unistd::unlink;
 use std::os::fd::{AsRawFd, RawFd};
@@ -13,7 +13,6 @@ const VFKIT_MAGIC: [u8; 4] = *b"VFKT";
 
 pub struct Gvproxy {
     fd: RawFd,
-    peer_addr: UnixAddr,
 }
 
 impl Gvproxy {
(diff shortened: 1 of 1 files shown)


src/devices/src/virtio/net/gvproxy.rs
```diff
@@ -34,8 +33,11 @@ impl Gvproxy {
         }
         bind(fd, &local_addr).map_err(ConnectError::Binding)?;
 
-        sendto(fd, &VFKIT_MAGIC, &peer_addr, MsgFlags::empty())
-            .map_err(ConnectError::SendingMagic)?;
+        // connect so we dn't need to use peer address again. This also allows the
```
tylerfanelli, line 36:
```suggestion
        // connect so we don't need to use peer address again. This also allows the
```

Typo

tylerfanelli (changes requested):
Small nit. Otherwise LGTM.

src/devices/src/virtio/net/gvproxy.rs
nirs: Thandks, fixed

src/devices/src/virtio/net/gvproxy.rs
```diff
@@ -34,8 +33,11 @@ impl Gvproxy {
         }
         bind(fd, &local_addr).map_err(ConnectError::Binding)?;
 
-        sendto(fd, &VFKIT_MAGIC, &peer_addr, MsgFlags::empty())
-            .map_err(ConnectError::SendingMagic)?;
+        // Connect so we dnn't need to use the peer address again. This also
```
tylerfanelli, line 36:
```suggestion
        // Connect so we don't need to use the peer address again. This also
```

Think you meant to add an 'o' here ;)

tylerfanelli (changes requested):

src/devices/src/virtio/net/gvproxy.rs
nirs: Trying again :-)

MatiasVara (approved):

slp:
There was a reason why we couldn't connect and needed to keep the `peer_addr` around, but I can't remember why now. I'll take a deeper look next week.

tylerfanelli (approved):
No, LGTM. Thanks.

tylerfanelli:
@slp Is this ok to merge?

PR #266, opened 2025-02-12
demote WARN log to DEBUG

jakecorrenti:
Addresses: https://github.com/containers/krunkit/issues/32


jakecorrenti:
I checked the other `warn!()` messages but they use terminology like "invalid" or "unknown" so I couldn't tell if those should stay elevated or not.

tylerfanelli (approved):
LGTM. I think you can leave the other `warn!` messages.

slp (approved):
Thanks @jakecorrenti !

PR #267, opened 2025-02-17
vsock: TCP test + fix a bug leading to an error in TCP connect call in guest 

mtjhrc:
This PR adds a simple tests that attempts to connect to a TCP socket on host from the guest. However this would often fail, due to a race condition:

As far as I understand, this is what happens:
`libkrun` connects to the TCP socket on the host (this action is done upon the TSI_CONNECT request), but the guest is not connected to stream vsock socket corresponding to the actual transport for that connection. (We have not called `confirm_connect`).  The problem is that since we assume the guest is already connected we attempt to recv here:
https://github.com/containers/libkrun/blob/a84dc1ac44c20713deff11343647ee51f5fc7931/src/devices/src/virtio/vsock/tcp.rs#L700-L714
This returns `wait_credit = true`, which causes us to send `VSOCK_OP_CREDIT_REQUEST` before sending the `VSOCK_OP_RESPONSE` to the `VSOCK_OP_REQUEST` which the guest kernel sends.
The kernel is not expecting this and returns EPROTO from `connect` call in the guest userspace. (specifically EPROTO is returned from the `virtio_transport_recv_connecting` in `net/vmw_vsock/virtio_transport_common.c`)

The fix here introduces a new state ConnectedUnconfirmed which it uses for TCP Proxy code in vsock. This state is used as a transitional state between the host socket being connected to libkrun and the `confirm_connect` being called, meaning the connection is actually established to the guest and we can actually send/recv.

A similar issue cannot happen with UDP sockets, because `VSOCK_OP_REQUEST`/`VSOCK_OP_RESPONSE` is only used for stream sockets. This also cannot **currently** happen with UNIX stream sockets, because currently these are only ever connected on the host and `confirm_connect`-ed at the same time here:
https://github.com/containers/libkrun/blob/a84dc1ac44c20713deff11343647ee51f5fc7931/src/devices/src/virtio/vsock/muxer.rs#L533-L534
Though, should the unix stream sockets also use the `ConnectedUnconfirmed` state too? For consistency of the semantics of the `Proxy` trait methods it would make sense...  

Depends on: https://github.com/containers/libkrun/pull/258

### DIFF
--- a/src/devices/src/virtio/vsock/proxy.rs
+++ b/src/devices/src/virtio/vsock/proxy.rs
@@ -25,6 +25,7 @@ pub enum ProxyError {
 pub enum ProxyStatus {
     Idle,
     Connecting,
+    ConnectedUnconfirmed,
     Connected,
     Listening,
     Closed,
--- a/src/devices/src/virtio/vsock/tcp.rs
+++ b/src/devices/src/virtio/vsock/tcp.rs
@@ -340,8 +340,7 @@ impl TcpProxy {
         push_packet(self.cid, rx, &self.rxq, &self.queue, &self.mem);
     }
 
-    fn switch_to_connected(&mut self) {
-        self.status = ProxyStatus::Connected;
+    fn switch_to_blocking(&mut self) {
         match fcntl(self.fd, FcntlArg::F_GETFL) {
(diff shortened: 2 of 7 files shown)


slp (approved):
LGTM, thanks!

PR #268, opened 2025-02-18
Fix incremental builds using make

nirs:
Add .PHONY to to the release and debug library targets so running make does the right thing after changing the source.

It is possible to use cargo directly, but this means a developer need to also run the more commands manually for each build which is error prone.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,7 +74,7 @@ ifeq ($(PREFIX),)
     PREFIX := /usr/local
 endif
 
-.PHONY: install clean
+.PHONY: install clean $(LIBRARY_RELEASE_$(OS)) $(LIBRARY_DEBUG_$(OS))
 
 all: $(LIBRARY_RELEASE_$(OS)) libkrun.pc
 


mtjhrc:
This is annoying indeed! :smile:, but I have already fixed it as part of: https://github.com/containers/libkrun/pull/258, here: https://github.com/containers/libkrun/pull/258/commits/010e99d8668e6d67e34f518066a530c19f9e21b2 
I also ran into and fixed this: https://github.com/containers/libkrun/pull/258/commits/d88544c42556ea3b74b84dff2c787187f93fdaa2

Maybe I could have fixed it as a separate PR, but it seemed quite simple and related to adding a `make test` target....

nirs:
> This is annoying indeed! 😄, but I have already fixed it as part of: #258, here: [010e99d](https://github.com/containers/libkrun/commit/010e99d8668e6d67e34f518066a530c19f9e21b2) I also ran into and fixed this: [d88544c](https://github.com/containers/libkrun/commit/d88544c42556ea3b74b84dff2c787187f93fdaa2)

Both fixes looks good, separating to another PR and merging seems like the right way.

mtjhrc (approved):

PR #270, opened 2025-02-20
libkrun.pc.in: add libs and cflags directories

slp:
Add variables for the headers and libraries to Cflags and Libs. This fixes non-system-wide installations of libkrun referenced by using PKG_CONFIG_PATH.


slp:
cc/ @mtjhrc

PR #271, opened 2025-02-25
tee: Remove legacy AMD-SEV module

tylerfanelli:
To prepare for the updating of the Rust SEV library to 5.0.0, remove the legacy SEV/SEV-ES module. AMD has all but deprecated legacy SEV, and has instead opted to push all new development to SEV-SNP.

As such, the SEV library is intending to deprecate legacy SEV.


tylerfanelli:
The clippy warnings are handled by #272. Once that is merged, this can be rebased.

src/vmm/src/linux/vstate.rs
```diff
@@ -469,13 +439,10 @@ pub struct Vm {
     irqchip_handle: Option<Box<dyn GICDevice>>,
 
     #[cfg(feature = "amd-sev")]
-    sev: Option<AmdSev>,
+    tee: Option<AmdSnp>,
```
jakecorrenti, line 442: Not blocking, but I'm wondering if this should be named `snp: ...`? It's going to get re-named once TDX and CCA land anyways

jakecorrenti (commented):
Do you plan on removing the SEV-related code in init/ as well? 

Other than that, LGTM. Just a non-blocking question as well.

src/vmm/src/linux/vstate.rs
tylerfanelli: Yea, will rename.

slp (approved):

PR #272, opened 2025-02-25
rutabaga_gfx: Fix clippy warnings

tylerfanelli:


jakecorrenti (approved):
LGTM

mtjhrc (changes requested):
PR looks good, though I am not sure if I am personally a fan of this clippy warning with all of the `(` `)`. I am not gonna fight it though, we should probably use the recommendation by clippy.

Please take another look at the `map_or` -> `is_none_or` suggestion, I think we can use it.

tylerfanelli:
> PR looks good, though I am not sure if I am personally a fan of this clippy warning with all of the `(` `)`. I am not gonna fight it though, we should probably use the recommendation by clippy.
>

Personally I'm not a fan either, but also don't like to fight clippy. Perhaps we can relax clippy's rules a bit? Up to you.
 
> Please take another look at the `map_or` -> `is_none_or` suggestion, I think we can use it.

It's currently an unstable library feature (I believe), which was preventing me from using that.

mtjhrc:
> > Please take another look at the `map_or` -> `is_none_or` suggestion, I think we can use it.
> 
> It's currently an unstable library feature (I believe), which was preventing me from using that.
What version of rust are you using though? As of Rust 1.82 (released 17 October 2024) it should be stable. So I don't think it's unreasonable to use it, though still relatively new.

tylerfanelli:
Yep, I was on an older version. Fixing now.

tylerfanelli:
@mtjhrc Can you re-review?

mtjhrc (approved):
LGTM now, thanks!

Issue #276, opened 2025-03-10
krun and crun handle quotes differently


eriksjolund:
krun and crun handle quotes differently when running 

```
podman --runtime $i run -t docker.io/library/fedora echo '"' ""
```

| crun git commit | OCI runtime | output |
|  --  |  -- | -- |
| e67610c84f6b08fe3c7898fa99e529187b62e37f | crun | `"` |
| e67610c84f6b08fe3c7898fa99e529187b62e37f | krun | `"" "` |
| e9a27669c400d44f371252a50c77351e0eb5440c | crun | `"` |
| e9a27669c400d44f371252a50c77351e0eb5440c | krun | `\"` |

(e9a27669c400d44f371252a50c77351e0eb5440c is the latest commit in the _main_ branch as of today 10 March 2025)

### Details

<details><summary>Click me</summary>

Do the following steps. 
Steps 1-8 are just for installing crun/krun.
Step 9 is the interesting step.

1. Log in as root on a Fedora 41 system
2. Create directory
   ```
   mkdir /root/containerfiles
   ```
3. Create file _/root/containerfiles/Containerfile.crunbuilder_
   ```
   FROM docker.io/library/fedora
   RUN dnf install -y make \
     automake autoconf gettext criu-devel libtool gcc libcap-devel systemd-devel \
     libseccomp-devel python3 libtool git protobuf-c protobuf-c-devel xz \
     libkrun-devel libkrun libkrun-sev libkrun-sev-devel \
     libkrunfw-devel libkrun libkrunfw-sev-devel libkrunfw-sev libkrunfw \
     yajl-devel yajl
   ```
4. Build image
   ```
   podman build -t crunbuilder -f Containerfile.crunbuilder /root/containerfiles/
   ```
5. Create file _/root/build.bash_
   ```
   #!/bin/bash
   set -o errexit
   set -o nounset
   commit=$1
   dir=$2
   git switch --detach $commit
   installdir=$dir/$commit
   mkdir $installdir
   podman run --rm \
      -v ./:/src:O \
      -v $installdir:/usr/local:Z \
      -w /src localhost/crunbuilder bash -c \
      "./autogen.sh && ./configure --with-libkrun && make -j4 && make install" 
   chcon --reference /usr/bin/crun $installdir/bin/crun
   chcon --reference /usr/bin/krun $installdir/bin/krun
   ```
6. Clone git repo
   ```
   git clone https://github.com/containers/crun.git
   ```
7. Create installation directory
   ```
[...]

PR #277, opened 2025-03-10
init: escape json strings

giuseppe:
Closes: https://github.com/containers/libkrun/issues/276

### DIFF
--- a/init/init.c
+++ b/init/init.c
@@ -451,6 +451,137 @@ static int mount_filesystems()
 	return 0;
 }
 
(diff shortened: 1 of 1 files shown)


flouthoc (approved):
LGTM

slp:
Nice one, thanks @giuseppe !
 
> @slp Do changes like this result in the initrd bundle needing an update in libkrunfw?

Indeed. I'm going to it right now to avoid forgetting about it.

Issue #275, opened 2025-03-10
Question about physical block devices


nohajc:
Hi,

thank you for this great project! I've been experimenting with it to see if I could mount a macOS host partition in a microVM but so far without success.

My use case is basically to spin up a minimalistic VM to access data from Asahi Linux or an external drive formatted with a Linux filesystem such as ext4 or btrfs.

To that end, I compiled libkrun with BLK=1 on macOS. Then I configured a VM with a virtiofs root and a virtio-block disk backed by /dev/diskXsY partition. If I run the VM as root (which is probably required to access block devices directly), it fails with an I/O error (errno 5) when the guest attempts to mount the filesystem. I even tried to grant full disk access permission to my VM runner program which didn't help.

On the other hand, when I do this with a regular filesystem image, everything works as expected.

From what I could find online, it seems you can use qemu to attach physical disks or partitions as raw images. So I just assumed it might work with libkrun too. Am I doing something wrong or is there a specific limitation in libkrun which is causing this issue?

When I look at the Apple Virtualization Framework documentation (I know you use the low-level Hypervisor Framework instead), I can see there are two different classes for storage devices: **VZDiskImageStorageDeviceAttachment** and **VZDiskBlockDeviceStorageDeviceAttachment**. That leads me to believe there are probably some implementation differences between these too but I have no idea how much work it would be to add support for physical block devices to libkrun assuming the lack of support is the problem here.

nohajc:
I'm playing with this more and I found out I'm able to open the mapped virtio block device from within the guest, it looks like sequential read succeeds too but when I try seeking, I get invalid argument error (errno 22).

Any idea why seeking might be broken for virtual disks backed by real block devices?

slp:
Just a wild guess, have you tried pointing virtio-blk to the raw disk (`/dev/rdisk*`) variant instead? If that doesn't work either, we would need to debug a bit to find which macOS syscalls in libkrun are failing when operating the device.

nohajc:
I tried both disk and rdisk with the same result.

nohajc:
So, I have another update.

> The given file must be seekable and sizable.
Seeing this comment in the code, I decided to check if the file size of a block device is correctly determined when the virtio device is initialized. As I suspected, it is not. It doesn't fail either. It just silently sets the size as zero.

I suppose it should be first detected if the image is a regular file or a block device and then use an appropriate method to retrieve the correct size.

I should be able to come up with a pull request. :)

nohajc:
[...]

PR #278, opened 2025-03-13
virtio/block: allow using host OS block devices in addition to regular files

nohajc:
This PR fixes file size detection of virtio block devices in case the backing file is also a block device on the host OS.
See #275.

### DIFF
--- a/src/imago/src/file.rs
+++ b/src/imago/src/file.rs
@@ -5,9 +5,11 @@ use crate::storage::drivers::CommonStorageHelper;
 use crate::{Storage, StorageOpenOptions};
 use std::fmt::{self, Display, Formatter};
 use std::fs;
-use std::io::{self, Seek, SeekFrom, Write};
+use std::io::{self, Write};
 #[cfg(any(target_os = "linux", target_os = "macos"))]
 use std::os::fd::AsRawFd;
+#[cfg(unix)]
+use std::os::unix::fs::FileTypeExt;
 #[cfg(all(unix, not(target_os = "macos")))]
 use std::os::unix::fs::OpenOptionsExt;
 #[cfg(windows)]
@@ -52,8 +54,8 @@ impl TryFrom<fs::File> for File {
     ///
     /// When using this, the resulting object will not know its own filename.  That makes it
     /// impossible to auto-resolve relative paths to it, e.g. qcow2 backing file names.
-    fn try_from(mut file: fs::File) -> io::Result<Self> {
-        let size = file.seek(SeekFrom::End(0))?;
+    fn try_from(file: fs::File) -> io::Result<Self> {
+        let size = get_file_size(&file)?;
 
         Ok(File {
             file: RwLock::new(file),
@@ -389,9 +391,9 @@ impl File {
         }
 
         let filename_owned = filename.to_owned();
-        let mut file = file_opts.open(filename)?;
+        let file = file_opts.open(filename)?;
 
-        let size = file.seek(SeekFrom::End(0))?;
+        let size = get_file_size(&file)?;
 
         #[cfg(target_os = "macos")]
         if opts.direct {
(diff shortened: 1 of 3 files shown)


src/imago/src/file.rs
```diff
@@ -454,3 +456,58 @@ impl Display for File {
         }
     }
 }
+
+fn get_file_size(file: &fs::File) -> io::Result<u64> {
+    file.metadata().and_then(|m| {
+        #[cfg(windows)]
+        let is_block_device = false;
+        #[cfg(unix)]
+        let is_block_device = m.file_type().is_block_device();
```
nohajc, line 465: Ok, so actually this is not a complete fix. It works on macOS with the buffered `/dev/disk*` but not `/dev/rdisk*`.
For one thing, I'd have to change the condition here to `is_block_device() || is_char_device()` but when I do that, I get `ERROR devices::virtio::block::worker] error processing request: WritingToDescriptor(Os { code: 22, kind: InvalidInput, message: "Invalid argument" })`.

I also noticed my disk sector size is 4096 but there's a lot of hardcoded 512 in the code. Raw disks don't allow unaligned access so that could be it...

Anyway, I'm not sure if I should attempt to fix this or rather ignore the character device case for now. Maybe return an error here.

jakecorrenti:
Hi @nohajc, thank you for the contribution! The `imago` crate is actually vendored from https://gitlab.com/hreitz/imago, so any changes should be made to that repository.

We should actually remove the vendor-ed code from libkrun and just add the `imago`  dependency since it's packaged in Fedora now. 

cc @XanClic

nohajc:
Oh, I haven't noticed. Thanks for pointing it out.

slp:
@nohajc Thanks for working on this. I'm going to keep this one open until the MR is created in the imago repo.

nohajc:
I'm happy to contribute. :) I'll see if I can open the correct MR today.

src/imago/src/file.rs
XanClic: I don’t think imago prescribes any block size. It queries the underlying block size and ensures requests are aligned to it (with read-modify-write cycles if necessary).

Imago’s users are responsible for choosing an appropriate block size, and imago reports the minimal underlying alignment via [FormatAccess::req_align()](https://docs.rs/imago/latest/imago/format/access/struct.FormatAccess.html#method.req_align) and [FormatAccess::mem_align()](https://docs.rs/imago/latest/imago/format/access/struct.FormatAccess.html#method.mem_align).

I assume the 512 comes from libkrun’s virtio-blk code, because 512 is the virtio-blk sector size. It’s possible to report other block sizes to the guest and thus force it to use those (via `virtio_blk_config.blk_size` and `virtio_blk_config.topology.physical_block_exp`).

nohajc: Yeah, that's actually the code I was referring to. The virtio block worker in libkrun.

And I was wondering how it reports the virtualized block size into the guest. So thanks for pointers.

XanClic: Ah, I think the problem is that libkrun opens the disk image without `direct_io` set, so the 0.1 version of imago assumes an alignment of 1. So backporting https://gitlab.com/hreitz/imago/-/commit/6aa125b06dc497dffec658b54f80d58f302535e5 to 0.1 could fix it.

nohajc:
> @nohajc Thanks for working on this. I'm going to keep this one open until the MR is created in the imago repo.
MR created: https://gitlab.com/hreitz/imago/-/merge_requests/18

src/imago/src/file.rs
nohajc: @slp Another improvement might be to set `direct_io` to `true` in libkrun if the file path begins with `/dev/`. What do you think?

slp: @nohajc I think in libkrun we should use `direct_io` unconditionally to avoid double-caching. @jakecorrenti @XanClic WDYT? Any reason to favor not using `direct_io`?

XanClic: @nohajc Thanks a lot of opening the MR and for testing 6aa125b0. I can backport those to 0.1, no problem.

@slp I think `direct_io` is a good idea. As far as I understand, for QEMU, it is generally suggested to disable the host page cache on VM images to avoid double-caching, as you say. It may cause issues with images in locations that don’t support `O_DIRECT`, but the only case I know of was tmpfs, which now supports it, too.

XanClic:
I’ve released 0.1.4 to fix this. I’ll do the Fedora release tomorrow.

XanClic:
(By the way, the Fedora release is done (from my side), too: https://src.fedoraproject.org/rpms/rust-imago)

nohajc:
I suppose this can be closed now. I'll look forward to the next release. Thank you everyone for your help!

slp:
> I suppose this can be closed now. I'll look forward to the next release. Thank you everyone for your help!

Thanks you @nohajc !

PR #279, opened 2025-03-13
Remove vendor-ed imago crate

jakecorrenti:
Remove the vendor-ed `imago` crate. Now that `imago` is packaged in Fedora, we can just add the depency through Cargo.

Related to: https://github.com/containers/libkrun/pull/278


slp (approved):
Scratching this from my TO-DO list, thanks @jakecorrenti !

PR #282, opened 2025-03-19
upgrade imago to 0.1.4

### LINKED ISSUE #275: Question about physical block devices

nohajc:
Fixes #275

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -235,6 +235,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "baf1de4339761588bc0619e3cbc0120ee582ebb74b53b4efbf79117bd2da40fd"
 
 [[package]]
+name = "cfg_aliases"
+version = "0.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "613afe47fcd5fac7ccf1db93babcb082c5994d996f20b8b159f2ad1658eb5724"
+
+[[package]]
 name = "chrono"
 version = "0.4.38"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


jakecorrenti (approved):
LGTM but we still need to wait for 1.4 to hit Fedora first

tylerfanelli:
Why do we need to wait for a Fedora release?

jakecorrenti:
> Why do we need to wait for a Fedora release of imago?

I suppose I assumed that would be the case since we needed to vendor the crate before it was in Fedora. 

If that's not the case then everything looks good to me.

tylerfanelli:
We'll need it when we release for Fedora, but that's a downstream process requirement. We don't have to let upstream wait because of it.

jakecorrenti:
> We'll need it when we release for Fedora, but that's a downstream process requirement. We don't have to let upstream wait because of it.

Sounds good. Thanks for the clarification.

PR #283, opened 2025-03-21
libkrun: avoid failing when libkrunfw is missing

slp:
In the last iteration of PR#256 we introduced the use of LazyLock, which helps us load the library earlier. The implementation was wronging because we were unwrapping if the library is missing, which is exactly what we were trying to prevent.

Replace the value with an Option, which allows us to avoid the unwrap and actually achieves the purpose of making libkrunfw optional.


tylerfanelli (approved):
LGTM.

tylerfanelli:
Probably should have tested with a missing libkrunfw :) my mistake

PR #287, opened 2025-03-25
ci: consolidate tests on github runners

slp:
Last year GitHub added nested virt support on x86_64 runners, in addition to aarch64 and macOS runners. Let's make use of them to simplify and consolidate our testing infrastructure.


jakecorrenti (approved):
I can't comment much on the YAML, but tests are green so LGTM

PR #288, opened 2025-03-26
Implement support for HVF Nested Virtualization

slp:
Apple introduced Nested Virt (EL2) support in macOS Sequoia, available
    on Apple Silicon devices based on M3 and later SoCs.

    This commit introduces the infrastructure to enable Nested Virt on
    libkrun. The biggest change is setting up the vCPU reset registers to
    values that are acceptable in EL2 according. This isn't easy since HVF
    doesn't document its expectations, but the current implementation allows
    the guest to boot in EL2 and run a nested guest.

    Instead of linking directly against the new functions, we're using
    libloader again to find the new symbols, to avoid breaking the binaries
    in Sonoma.


slp:
This depends on #280, creating this as a draft until it's merged.

slp:
Rebased, working and ready to be reviewed!

jakecorrenti (commented):
Code LGTM but I can't test the nested virt since I don't have an M3.

src/hvf/src/lib.rs
```diff
@@ -156,11 +206,56 @@ pub fn vcpu_set_vtimer_mask(vcpuid: u64, masked: bool) -> Result<(), Error> {
...
+        libloading::Symbol<'static, unsafe extern "C" fn(hv_vm_config_t, *mut bool) -> hv_return_t>,
+}
+
 pub struct HvfVm {}
 
+static HVF: LazyLock<libloading::Library> = LazyLock::new(|| unsafe {
+    libloading::Library::new(
+        "/System/Library/Frameworks/Hypervisor.framework/Versions/A/Hypervisor",
+    )
+    .unwrap()
```
tylerfanelli, line 222: Does it make sense to fail gracefully here in the case that the `Hypervisor` library is either unavailable or of a different version?

tylerfanelli (approved):
One small nit, but overall LGTM! I also don't have an M3 to test however.

src/hvf/src/lib.rs
woachk: That'd be an ABI break on the macOS side, it's ~ the equivalent of .so versions on Linux

slp: Well, in the unlikely case Apple stops shipping version 'A' of Hypervisor.framework, we can't do much other than printing a different error message. ;-P

PR #296, opened 2025-04-02
tee/snp: Update SEV-SNP APIs, use KVM guest_memfd

tylerfanelli:
This series updates the SEV-SNP APIs (sev library dependency to 6.0.0) and adds KVM guest_memfd support for SEV-SNP guests. This is required for running SEV-SNP encrypted guests on upstream kernels.

Tested on linux `6.14.0-0.rc3`

cc @slp @jakecorrenti @MatiasVara


src/vmm/src/linux/vstate.rs
```diff
@@ -1099,6 +1240,16 @@ impl Vcpu {
                     self.io_bus.write(0, u64::from(addr), data);
                     Ok(VcpuEmulation::Handled)
                 }
+                VcpuExit::MemoryFault { gpa, size, flags } => {
+                    let private = (flags & (1 << 3)/* KVM_EXIT_MEMORY_FAULT */) != 0;
```
jakecorrenti, line 1244:
```suggestion
                    let private = (flags & kvm_bindings::KVM_EXIT_MEMORY_FAULT) != 0;
```

src/vmm/src/linux/vstate.rs
```diff
@@ -1132,7 +1283,7 @@ impl Vcpu {
                 r => {
                     // TODO: Are we sure we want to finish running a vcpu upon
                     // receiving a vm exit that is not necessarily an error?
-                    error!("Unexpected exit reason on vcpu run: {:?}", r);
+                    error!("Unexpected exit reason on vcpu run: {:x?}", r);
```
jakecorrenti, line 1286: Should this be prefixed with 0x so the user knows it's in hex not decimal?

src/vmm/src/linux/vstate.rs
```diff
@@ -455,13 +480,18 @@ impl Vm {
             supported_cpuid,
             #[cfg(target_arch = "x86_64")]
             supported_msrs,
+            guest_memfds: RangeMap::new(),
         })
     }
 
     #[cfg(feature = "amd-sev")]
     pub fn new(kvm: &Kvm, tee_config: &TeeConfig) -> Result<Self> {
+        use crate::vstate::Error::*;
```
jakecorrenti, line 489: You could avoid this import and just do `Error::HypercallExitEnable` below

src/vmm/src/linux/vstate.rs
```diff
@@ -97,6 +102,11 @@ pub enum Error {
     #[cfg(target_arch = "x86_64")]
     /// Error configuring the general purpose registers
     REGSConfiguration(arch::x86_64::regs::Error),
+    #[cfg(target_arch = "aarch64")]
+    /// Error setting up the global interrupt controller.
+    SetupGIC(arch::aarch64::gic::Error),
```
tylerfanelli, line 113: I must've accidentally included this as a result of a rebase. Will remove from this commit.

src/libkrun/src/lib.rs
```diff
@@ -1491,6 +1509,91 @@ pub extern "C" fn krun_start_enter(ctx_id: u32) -> i32 {
...
+                Ok(m) => {
+                    let (guest_memfd, region_start) = mapper_vmm
+                        .lock()
+                        .unwrap()
+                        .kvm_vm()
+                        .guest_memfd_get(m.gpa)
+                        .unwrap();
+
+                    let attributes: u64 = if m.private {
+                        1 << 3 /* KVM_MEMORY_ATTRIBUTE_PRIVATE */
```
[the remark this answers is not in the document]
tylerfanelli: Fixed in latest version.

src/libkrun/src/lib.rs
```diff
@@ -1491,6 +1509,91 @@ pub extern "C" fn krun_start_enter(ctx_id: u32) -> i32 {
...
+                        .lock()
+                        .unwrap()
+                        .kvm_vm()
+                        .fd()
+                        .set_memory_attributes(attr)
+                        .unwrap();
+
+                    let region = guest_mem.find_region(GuestAddress(m.gpa));
+                    if region.is_none() {
+                        error!("Region not found");
```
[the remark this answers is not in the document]
tylerfanelli: Thanks, modified the error message.

src/vmm/src/linux/vstate.rs
tylerfanelli: This was a typo, the actual flag I was referring to was `KVM_MEMORY_EXIT_FLAG_PRIVATE`, but your point still stands. Fixed in latest version.

src/vmm/src/linux/vstate.rs
tylerfanelli: I ultimately decided to drop the commit altogether, as I think it merely refers to a linux errno, without any addresses present.

src/vmm/src/linux/vstate.rs
tylerfanelli: Thanks, fixed in latest version.

src/libkrun/src/lib.rs
```diff
@@ -1454,12 +1465,19 @@ pub extern "C" fn krun_start_enter(ctx_id: u32) -> i32 {
     #[cfg(target_os = "macos")]
     let (sender, receiver) = unbounded();
 
+    #[cfg(feature = "tee")]
+    let (pm_sender, pm_receiver) = unbounded();
+    #[cfg(feature = "tee")]
+    let pm_efd = EventFd::new(libc::EFD_SEMAPHORE).unwrap();
```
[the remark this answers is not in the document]
tylerfanelli: The blocking feature is present unless `EFD_NONBLOCK` is specified. The `EFD_SEMAPHORE` flag indicates that the eventfd counter will only be decremented by 1 when a read of the file occurs, rather than setting the counter value to zero. This is used in the scenario that multiple vCPUs (that is, multiple threads) need to send `MemoryProperties` data to the handler.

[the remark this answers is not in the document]
jakecorrenti: Understood. Thanks

jakecorrenti:
Re-ran the Integration Tests but LGTM otherwise

MatiasVara:
> @jakecorrenti @slp @MatiasVara Can you (re-)review? I'm not sure why the integration test (starting a VM with 1 vCPU and 1GiB of memory) fails. I'm able to boot a (non-confidential) VM with as little as 64MiB of memory.

It is in my TODO list. I just had no time :'(

MatiasVara:
I'm trying to understand why the rebase wasn’t done on my branch, which already included the guestmemfd (see https://github.com/containers/libkrun/pull/211/commits/15ac9fcb4b9f68adf99a1b133becf60a7a30c38e) and KVM_EXIT_MEMORY_FAUL (see https://github.com/containers/libkrun/pull/211/commits/5c3f283fb553b7a5320a524eded1b27e5a33ad2f) changes since last year—correct me if I’m wrong. To speed things up, I could create a new PR with just those two changes, so you would only need to rebase on top of that. I’d like to avoid any duplication of effort.

tylerfanelli:
@MatiasVara I could do that if you prefer. The reason is because I needed to handle `KVM_EXIT_HYPERCALL` first, as on SEV-SNP that is what is encountered first. Therefore, I thought it was just be worthwhile to add `KVM_EXIT_MEMORY_FAULT` later.

tylerfanelli:
@MatiasVara I also did things a bit differently with the managing of guest_memfds, associating with a `RangeMap` rather than a standard `Vec`.

tylerfanelli:
@MatiasVara Can you rebase your `use-guest-memfd` branch on `main`?

tylerfanelli:
@MatiasVara PTAL. I've rebased on your branch.

tylerfanelli:
@mtjhrc Not sure why the integration test is failing, I'm able to run it locally just fine:

```
$ make test

.. snip ..

[configure-vm-1cpu-256MiB]: OK
[configure-vm-2cpu-1GiB]: OK
[vsock-guest-connect]: OK
[tsi-tcp-guest-connect]: OK
[tsi-tcp-guest-listen]: OK

OK (PASSED 5/5)
```

tylerfanelli:
@slp @MatiasVara @jakecorrenti Can you re-review? it seems the integration tests failure will need to be handled in a separate PR as its failing on non-TEE tests as well.

src/libkrun/src/lib.rs
```diff
@@ -1576,6 +1601,93 @@ pub extern "C" fn krun_start_enter(ctx_id: u32) -> i32 {
...
+                        flags: 0,
+                    };
+
+                    mapper_vmm
+                        .lock()
+                        .unwrap()
+                        .kvm_vm()
+                        .fd()
+                        .set_memory_attributes(attr)
+                        .unwrap_or_else(|_| panic!("unable to set memory attributes for memory region corresponding to guest address 0x{:x}", m.gpa));
```
jakecorrenti, line 1641: You could do
 ```rust
.set_memory_attributes(attr).expect(<error message>)
``` 
if you want to panic with a specific message if it's an Error result

File: src/vmm/src/linux/vstate.rs [hunk 9]
```diff
@@ -452,6 +458,8 @@ pub struct Vm {
 
     #[cfg(feature = "amd-sev")]
     pub tee_config: Tee,
+
+    pub guest_memfds: RangeMap<u64, (RawFd, u64)>,
```
jakecorrenti, line 458: IMO we can keep track of this without needing a dependency. In my TDX branch I used a Vec of a custom type:

```rust
#[cfg(feature = "intel-tdx")]
pub struct GuestMemfdProperties {
    pub gpa: vm_memory::GuestAddress,
    pub size: u64,
    pub memfd_id: u64,
}
```

see here: https://github.com/containers/libkrun/pull/313/commits/6a50fe18ec699fc1bf66e5c44d2a73eb3824f957

src/vmm/src/linux/vstate.rs
```diff
@@ -551,16 +561,31 @@ impl Vm {
...
+        let end = start + region.len();
+
+        let mut use_guest_memfd = false;
+
+        #[cfg(feature = "tee")]
+        if self.fd.check_extension(GuestMemfd) {
+            use_guest_memfd = true;
+        }
+
+        if !use_guest_memfd {
```
jakecorrenti, line 581: `check_extension()` already returns `bool`. Could we just do 
```rust
if !self.fd.check_extension(GuestMemfd) {
...
}
```

src/libkrun/src/lib.rs
```diff
@@ -1486,6 +1499,11 @@ pub extern "C" fn krun_start_enter(ctx_id: u32) -> i32 {
 
     #[cfg(target_arch = "x86_64")]
     let (irq_sender, irq_receiver) = crossbeam_channel::unbounded();
+    #[cfg(feature = "tee")]
+    let (pm_sender, pm_receiver) = unbounded();
+    #[cfg(feature = "tee")]
+    let pm_efd =
+        EventFd::new(EFD_SEMAPHORE).expect("unable to create TEE memory properties eventfd");
```
jakecorrenti, line 1506: I won't block on this, but another option here, which would match the macOS implementation, is to use a response sender and receiver.

i.e. 

the memory properties sender would have the signature of `crossbeam_channel::Sender<(crossbeam_channel::Sender<bool>, MemoryProperties)>`. 

before sending the message to the worker thread you would create a response channel:

```rust
let (response_sender, response_receiver) = unbounded();
memory_sender.send((response_sender, memory_properties)).unwrap();
if !response_receiver.recv().unwrap() {
    // there was some error in the worker thread
}
```

and in the worker thread, which would receive a tuple with the response sender and the memory properties, you would send a message back to the response receiver with the status of the task via a boolean.

This would ultimately allow us to avoid carrying an EventFd around through the codepath. Just food for thought. FWIW I did end up going down this route when consolidating the worker threads.

jakecorrenti (commented):
commits 6833eb7 and 6be8e02 could be squashed, as well as the last two commits.

src/vmm/src/linux/vstate.rs
tylerfanelli: But this data structure provides ranged keys, so you don't need to add logic of iterating over guest addresses and checking if their within some region in a `Vec`. You simply supply an address, and the map itself will ensure you get the correct region. It's simpler IMO.

tylerfanelli:
> Commits [6833eb7](https://github.com/containers/libkrun/commit/6833eb7e013e8190d993c1e9f740af3f7f5742e5) and [6be8e02](https://github.com/containers/libkrun/commit/6be8e0228019359443f645f7164643dadbb2cbe8) could be squashed, as well as the last two commits.
>

I don't believe they should, as they are from two different authors.

> Also... Are we positive that nothing in this PR is causing the CI failure? I've posted 2 PRs myself after this one and haven't run into the issue with Integration tests. I only ask because I'd like to avoid failing CI on main if possible

The integration test fails on non-TEE tests. **Perhaps** its the thread sleep in `KVM_RUN`. Will configure that to only on be present with the `tee` feature and test.

src/libkrun/src/lib.rs
tylerfanelli: Was there a reason for using another sender/receiver pair rather than an event fd? The device code uses event fd for interrupts, so I'd probably want to remain uniform.

jakecorrenti: I guess the question is _what_ do we be uniform with :) I'm not sure the reasoning behind the macOS implementation. Maybe @slp could speak on that

jakecorrenti:
> > Commits [6833eb7](https://github.com/containers/libkrun/commit/6833eb7e013e8190d993c1e9f740af3f7f5742e5) and [6be8e02](https://github.com/containers/libkrun/commit/6be8e0228019359443f645f7164643dadbb2cbe8) could be squashed, as well as the last two commits.
> 
> I don't believe they should, as they are from two different authors.

You can add `Co-authored by: <signature>` in the git commit to show you and someone else should get credit for the squashed commit. IMO it would result in a cleaner git history. 

> 
> > Also... Are we positive that nothing in this PR is causing the CI failure? I've posted 2 PRs myself after this one and haven't run into the issue with Integration tests. I only ask because I'd like to avoid failing CI on main if possible
> 
> The integration test fails on non-TEE tests. **Perhaps** its the thread sleep in `KVM_RUN`. Will configure that to only be present with the `tee` feature and test.

MatiasVara:
> @slp @MatiasVara @jakecorrenti Can you re-review? it seems the integration tests failure will need to be handled in a separate PR as its failing on non-TEE tests as well.

Apologies, I did not have time to see this. I will try next week.

src/libkrun/src/lib.rs
tylerfanelli: For some reason, clippy complains when I do that:

```
warning: use of `expect` followed by a function call
    --> src/libkrun/src/lib.rs:1620:26
     |
1620 |                         .expect(&format!("unable to find KVM guest_memfd for memory region corresponding to GPA 0x{:x}", m.gpa));
     |                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try: `unwrap_or_else(|| panic!("unable to find KVM guest_memfd for memory region corresponding to GPA 0x{:x}", m.gpa))`
     |
     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#expect_fun_call
     = note: `#[warn(clippy::expect_fun_call)]` on by default
```

I agree it's nicer with `expect`, but err towards clippy.

src/vmm/src/linux/vstate.rs
tylerfanelli: With this, `use_guest_memfd` will always be false in non-TEE guests. With your suggestion, guest_memfd would be used even on non-TEE guests if supported. Do we want to do that? I'm not sure. For now, I'd say we shouldn't use guest_memfd when it's not required.

jakecorrenti: The GuestMemfd capability is only enabled for SEV and KVM_SW_PROTECTED_VM types (for now). This should never be enabled on a non-TEE system. 

See: 
- https://github.com/torvalds/linux/blob/master/arch/x86/kvm/Kconfig
- https://github.com/torvalds/linux/blob/9d7a0577c9db35c4cc52db90bc415ea248446472/virt/kvm/kvm_main.c#L4845

src/libkrun/src/lib.rs
tylerfanelli: event fd is the standard way to deal with this, rather than another message channel just for notification that something is completed.

src/vmm/src/linux/vstate.rs
tylerfanelli: Good catch, then my `#[cfg(feature = "tee")]` is redundant.

tylerfanelli: Fixed.

jakecorrenti (approved):

jakecorrenti:
Ideally I would like a review from @MatiasVara as well since this impacts him as well, but if that can't happen that's ok too

MatiasVara:
> > > Commits [6833eb7](https://github.com/containers/libkrun/commit/6833eb7e013e8190d993c1e9f740af3f7f5742e5) and [6be8e02](https://github.com/containers/libkrun/commit/6be8e0228019359443f645f7164643dadbb2cbe8) could be squashed, as well as the last two commits.
> > 
> > 
> > I don't believe they should, as they are from two different authors.
> 
> You can add `Co-authored by: <signature>` in the git commit to show you and someone else should get credit for the squashed commit. IMO it would result in a cleaner git history, but I'm not going to block you on that.
> 
> > > Also... Are we positive that nothing in this PR is causing the CI failure? I've posted 2 PRs myself after this one and haven't run into the issue with Integration tests. I only ask because I'd like to avoid failing CI on main if possible
> > 
> > 
> > The integration test fails on non-TEE tests. **Perhaps** its the thread sleep in `KVM_RUN`. Will configure that to only be present with the `tee` feature and test.

I agree with Jake that those commits could be squashed, including `2f25d7d527ffda24a44291977f8b9da4fd68b035` and `81f46532f13fb44bbcac7c5a5606c2e676b5e0cc`. I understand this is a result of the rebasing. My comment [here](https://github.com/containers/libkrun/pull/296#issuecomment-2779227977) was meant to point out that these changes could have been addressed directly during the review of #211. It's something we should keep in mind for next time. The PR LGTM.

MatiasVara (approved):

tylerfanelli:
> > > > Commits [6833eb7](https://github.com/containers/libkrun/commit/6833eb7e013e8190d993c1e9f740af3f7f5742e5) and [6be8e02](https://github.com/containers/libkrun/commit/6be8e0228019359443f645f7164643dadbb2cbe8) could be squashed, as well as the last two commits.
> > > 
> > > 
> > > I don't believe they should, as they are from two different authors.
> > 
> > 
> > You can add `Co-authored by: <signature>` in the git commit to show you and someone else should get credit for the squashed commit. IMO it would result in a cleaner git history, but I'm not going to block you on that.
> > > > Also... Are we positive that nothing in this PR is causing the CI failure? I've posted 2 PRs myself after this one and haven't run into the issue with Integration tests. I only ask because I'd like to avoid failing CI on main if possible
> > > 
> > > 
> > > The integration test fails on non-TEE tests. **Perhaps** its the thread sleep in `KVM_RUN`. Will configure that to only be present with the `tee` feature and test.
> 
> I agree with Jake that those commits could be squashed, including `2f25d7d527ffda24a44291977f8b9da4fd68b035` and `81f46532f13fb44bbcac7c5a5606c2e676b5e0cc`. I understand this is a result of the rebasing.

I've generally avoided squashing commits that were from two separate authors. Does the `Co-authored-by:` line preserve authorship in git, or does authorship switch to the author of the squashed commit?

> My comment [here](https://github.com/containers/libkrun/pull/296#issuecomment-2779227977) was meant to point out that these changes could have been addressed directly during the review of #211. It's something we should keep in mind for next time. The PR LGTM.

I see. It was a bit of a conundrum because those changes weren't able to be merged yet, so I thought I'd address the changes in my PR. Under typical circumstances I completely agree.

Bit of a strange circumstance here, but I will definitely keep your comments in mind in the future.

Issue #293, opened 2025-03-29
Flush operations and block devices


nohajc:
When I configure a disk backed by macOS block device and try to use it in read/write mode, I end up with the following error:

> [2025-03-28T23:29:45Z ERROR devices::virtio::block::worker] error processing request: FlushingToDisk(Os { code: 25, kind: Uncategorized, message: "Inappropriate ioctl for device" })

It seems to come from here:

```
VIRTIO_BLK_T_FLUSH => match self.disk.cache_type() {
    CacheType::Writeback => {
        let diskfile = self.disk.file();
        diskfile.flush().map_err(RequestError::FlushingToDisk)?;
        diskfile.sync().map_err(RequestError::FlushingToDisk)?;
        Ok(0)
    }
    CacheType::Unsafe => Ok(0),
},
```

It happens with both `/dev/rdisk*` and `/dev/disk*` files.

I am able to make it work if I set `CacheType` to `Unsafe` but this is not exposed in the C API. Do you think it would be a good idea to do so, or is there any other way to address this?

Thank you.

slp:
I see two options:

1. Introduce an `krun_add_disk3` so users can set up `CacheType` manually.
2. On macOS, detect if the backend is `/dev/rdisk*` (I'm not sure if it would be safe doing it with `/dev/disk*` since it's buffered) and use `CacheType::Unsafe` automatically.

Personally, I'd prefer 2. Is there a way on macOS to detect if the backend is a raw disk beyond check the path name?

nohajc:
I thought it would work with the buffered `/dev/disk*` but it gave me that same error. I'll investigate more. It can fail on the flush operation or the subsequent sync. Both return the same thing so I don't actually know yet which one is being triggered here.

Besides checking the path, you can check the file type which is a block device for `/dev/disk*` and a character device for `/dev/rdisk*`.

slp:
In that case, I think the best approach is defaulting to `CacheType::Unsafe` if the path is `/dev/rdisk*` and the target is `macos`. Would you like to propose a PR doing that?

nohajc:
Yes, I'm going to open a PR. So, you'd prefer to match `/dev/rdisk*` path rather than any character device? Maybe there are other ways to detect this (I haven't really done any research yet) but I suppose the path should be a safe bet.

I could also find out how this works on Linux host to make a similar adjustment there as well.

PR #297, opened 2025-04-02
devices: set CacheType::Unsafe for macOS raw disks

nohajc:
See #293

### DIFF
--- a/src/devices/src/virtio/block/device.rs
+++ b/src/devices/src/virtio/block/device.rs
@@ -51,6 +51,18 @@ pub enum CacheType {
     Writeback,
 }
 
+impl CacheType {
+    /// Picks the appropriate cache type based on disk image or device path.
+    /// Special files like `/dev/rdisk*` on macOS do not support flush/sync.
+    pub fn auto(_path: &str) -> CacheType {
+        #[cfg(target_os = "macos")]
+        if _path.starts_with("/dev/rdisk") {
+            return CacheType::Unsafe;
+        }
+        CacheType::Writeback
+    }
+}
+
 /// Helper object for setting up all `Block` fields derived from its backing file.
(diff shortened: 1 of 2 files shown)


tylerfanelli (dismissed):
LGTM. You mentioned in the issue that this problem persists with `/dev/disk*` as well. Do you plan to address that in this PR as well?

nohajc:
@slp said
> I'm not sure if it would be safe doing it with /dev/disk* since it's buffered

Besides, there's no practical reason to use `/dev/disk*` when you have the raw variant which provides more efficient access anyway.

src/devices/src/virtio/block/device.rs
```diff
@@ -51,6 +51,18 @@ pub enum CacheType {
     Writeback,
 }
 
+impl CacheType {
+    /// Picks the appropriate cache type based on disk image or device path.
+    /// Special files like `/dev/rdisk*` on macOS do not support flush/sync.
+    pub fn auto(#[allow(unused_variables)] path: &str) -> CacheType {
```
slp, line 57: nit: could you please use `_path` instead `path` and remove the annotation. This approach is perfectly valid, but everywhere else we're using the underscore, so feels nicer for consistency.

nohajc: Sure, it's more concise too. Done.

nohajc:
Don't know what's with the failing checks now... they seem unrelated.

nohajc:
I also noticed a crash in unit tests...

slp:
> I also noticed a crash in unit tests...

Yes, and I can reproduce the failure here after updating to `1.86.0`.

slp:
@nohajc Tests should be fixed with #298, please rebase this PR.

nohajc:
Done

PR #298, opened 2025-04-03
Another round of clippy fixes and deprecated unit tests

slp:


src/arch_gen/src/x86/mpspec.rs
```diff
@@ -54,119 +54,6 @@ pub struct mpf_intel {
...
-        16usize,
-        concat!("Size of: ", stringify!(mpf_intel))
-    );
-    assert_eq!(
-        ::std::mem::align_of::<mpf_intel>(),
-        4usize,
-        concat!("Alignment of ", stringify!(mpf_intel))
-    );
-    assert_eq!(
-        unsafe { &(*(0 as *const mpf_intel)).signature as *const _ as usize },
```
mtjhrc, line 70: I'm not totally against removing these. But we could fix these tests relatively easily by using `offset_of!` instead of the sketchy null pointer de-reference here.  

`sed -i 's/unsafe { &(\*(0 as \*const \([^(]*\)))\.\([^(]*\) as \*const _ as usize }/::std::mem::offset_of!(\1, \2)/' src/arch_gen/src/x86/mpspec.rs`

slp: I've considered fixing it, but then I found `bindgen` no longer generates those tests (see https://gist.github.com/slp/1dfcd092ca3441632c0e10d426bae451), at least for the types in this header.

tylerfanelli (approved):

src/arch_gen/src/x86/mpspec.rs
mtjhrc: Ok then, makes sense

mtjhrc (approved):

src/arch_gen/src/x86/mpspec.rs
mtjhrc: Oh but actually it generates compile time assertions now:

```#[allow(clippy::unnecessary_operation, clippy::identity_op)]
const _: () = {
    ["Size of mpc_intsrc"][::std::mem::size_of::<mpc_intsrc>() - 8usize];
    ["Alignment of mpc_intsrc"][::std::mem::align_of::<mpc_intsrc>() - 2usize];
    ["Offset of field: mpc_intsrc::type_"][::std::mem::offset_of!(mpc_intsrc, type_) - 0usize];
};```

slp: That's a nice improvement. I've created #299 to track the regeneration of bindgen files.

PR #302, opened 2025-04-04
vmm/linux/vstate: support KVM SystemEvents

slp:
On aarch64, we receive the shutdown/reset signals as KVM SystemEvents. Let's identify them properly to avoid printing an error on exit.

### DIFF
--- a/src/vmm/src/linux/vstate.rs
+++ b/src/vmm/src/linux/vstate.rs
@@ -40,7 +40,9 @@ use kvm_bindings::{
     KVM_CLOCK_TSC_STABLE, KVM_IRQCHIP_IOAPIC, KVM_IRQCHIP_PIC_MASTER, KVM_IRQCHIP_PIC_SLAVE,
     KVM_MAX_CPUID_ENTRIES,
 };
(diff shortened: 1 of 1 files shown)


tylerfanelli (approved):

src/vmm/src/linux/vstate.rs
```diff
@@ -1129,6 +1131,14 @@ impl Vcpu {
                     error!("Received KVM_EXIT_INTERNAL_ERROR signal");
                     Err(Error::VcpuUnhandledKvmExit)
                 }
+                VcpuExit::SystemEvent(event, _reason) => {
```
tylerfanelli, line 1134: Should this be tagged behind `#[cfg(target_arch = "aarch64")]`?

tylerfanelli (approved):
One small nit.

src/vmm/src/linux/vstate.rs
slp: On x86_64 we don't hit that one right now, because we shut down with a triple fault, but it's valid a could eventually happen.

PR #303, opened 2025-04-04
init: support cwd as workdir

slp:
Support "Cwd" in addition to "WorkingDir" as workdir, as both seem to be used indistinctively.


tylerfanelli (approved):

jakecorrenti (approved):
LGTM

PR #311, opened 2025-04-09
include: Add #ifndef header guards

tylerfanelli:


nohajc:
This reminds me you could also add

```
#ifdef __cplusplus
extern "C" {
#endif
...
```

mtjhrc (approved):

slp (changes requested):
It's a good idea, but please address @nohajc and @mtjhrc comments before merging.

tylerfanelli:
@mtjhrc @nohajc Thanks for the suggestions, I've applied them. @slp Can you re-review?

slp (approved):
LGTM, thanks!

Issue #306, opened 2025-04-07
Exit status not propagated from the guest


nohajc:
There are use cases where you run a single command inside a VM and you'd like to preserve its exit code. At least to distinguish between a successful and a failed run. Currently, it seems no matter how the guest executable terminates, you just get orderly shutdown and exit code SUCCESS.

Is there any way to implement exit code propagation? I suppose it could be an optional feature. I could try to implement it myself, just don't know what the mechanism could be to communicate this from the guest to the host. Or if there are any other existing VMMs doing this.

slp:
This issue was raise some weeks ago here https://github.com/containers/crun/issues/1694

The trickiest part is transmitting the exit code from the guest to the host. Last week I've started implementing a solution for this based on virtio-fs, I should be able to have a PR ready by the end of the week.

nohajc:
Thanks, that's great to hear!

Out of curiosity, why virtio-fs and not vsock let's say?

jakecorrenti:
Closing as fixed. If this still needs more discussion, feel free to re-open it.

PR #312, opened 2025-04-09
Implement the ability to record the exit code from the entrypoint in userspace

slp:
For the container use case, we need to relay the exit code from
    userspace in the microVM all the way to the process using libkrun to
    launch the VMM.

    Since ioctls are passed mostly unmodified from userspace in the guest
    to the virtio-fs device, we can use them as a mechanism for transporting
    this information without requiring any specific support in the guest's
    kernel.

    Here we create an AtomicI32 and wire it up between virtio-fs and Vmm,
    using it as exit code if userspace has set it to some value other than
    i32::MAX. Otherwise, we keep using the vCPU exit code, as we did before.

### DIFF
--- a/init/init.c
+++ b/init/init.c
@@ -980,6 +980,7 @@ int main(int argc, char **argv)
     struct ifreq ifr;
     int sockfd;
     int status;
+    int saved_errno;
     char localhost[] = "localhost\0";
     char *hostname;
     char *krun_home;
@@ -1067,16 +1068,23 @@ int main(int argc, char **argv)
     int child = fork();
     if (child < 0) {
         perror("fork");
-        exit(-3);
+        set_exit_code(125);
+        exit(125);
     }
     if (child == 0) { // child
         if (setup_redirects() < 0) {
-            exit(-4);
+            exit(125);
         }
         if (execvp(exec_argv[0], exec_argv) < 0) {
+            saved_errno = errno;
             printf("Couldn't execute '%s' inside the vm: %s\n", exec_argv[0],
                    strerror(errno));
-            exit(-3);
+            // Use the same exit code as chroot and podman do.
+            if (saved_errno == ENOENT) {
+                exit(127);
+            } else {
+                exit(126);
+            }
         }
     } else { // parent
         // Wait until our first child has exited, ignoring any other children.
--- a/init/init.c
+++ b/init/init.c
@@ -28,6 +28,8 @@
 #include "tee/snp_attest.h"
 #endif
 
+#define KRUN_EXIT_CODE_IOCTL 0x7602
+
 #define KRUN_MAGIC "KRUN"
 #define KRUN_FOOTER_LEN 12
 #define CMDLINE_SECRET_PATH "/sfs/secrets/coco/cmdline"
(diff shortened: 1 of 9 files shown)


slp:
Fixes #306, https://github.com/containers/crun/issues/1694 and https://github.com/containers/crun/issues/1688

src/vmm/src/lib.rs
```diff
@@ -394,15 +396,23 @@ impl Subscriber for Vmm {
             // If the exit_code can't be found on any vcpu, it means that the exit signal
             // has been issued by the i8042 controller in which case we exit with
             // FC_EXIT_CODE_OK.
```
nohajc, line 398: Maybe this comment could be updated.

init/init.c
```diff
@@ -1057,11 +1079,17 @@ int main(int argc, char **argv)
             exit(-3);
         }
     } else { // parent
-        // tell the kernel we don't want to be notified on SIGCHLD so it'll reap
-        // our children for us
-        signal(SIGCHLD, SIG_IGN);
-        // wait for children since we can't exit init
-        waitpid(pid, NULL, 0);
+        // Wait until our first child has exited, ignoring any other children.
```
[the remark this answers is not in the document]
slp: I think you're right, perhaps `workload's entrypoint` would be more accurate?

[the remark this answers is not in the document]
mtjhrc: Sure, that sounds more accurate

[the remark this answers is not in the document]
slp: Done.

src/vmm/src/lib.rs
slp: Done.

slp:
I think I've addressed all the feedback.

include/libkrun.h
```diff
@@ -582,10 +582,16 @@ int32_t krun_split_irqchip(uint32_t ctx_id, bool enable);
...
  *
- * Returns:
+ * Notes:
  *  This function only returns if an error happens before starting the microVM. Otherwise, the
  *  VMM assumes it has full control of the process, and will call to exit() once the microVM shuts
  *  down.
+ *
+ * Returns:
+ *  -EINVAL - The VMM has detected an error in the microVM configuration.
+ *  125     - "init" cannot set up the environment inside the microVM.
```
tylerfanelli, line 592: Should we define these as enums available in `libkrun.h`?

slp: Both in podman and chroot they're just referred directly by their number, so I think us giving them a name would generate confusion more than anything else.

tylerfanelli: Ok sounds good to me.

tylerfanelli (dismissed):

mtjhrc (changes requested):

include/libkrun.h
```diff
@@ -582,10 +582,16 @@ int32_t krun_split_irqchip(uint32_t ctx_id, bool enable);
...
+ *  126     - "init" can find the executable to be run inside the microVM but cannot not execute it.
+ *  127     - "init" cannot find the executable to be run inside the microVM.
```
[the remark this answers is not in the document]
slp: Yes, this looks better. I've squashed this commit into the previous one and added you as co-author.

tylerfanelli (approved):

PR #315, opened 2025-04-16
Update KVM dependencies

tylerfanelli:


jakecorrenti:
I'm good with the change, but my TDX PR is going to switch the dependencies to use the virtee forks since the TDX bindings aren't in upstream (and I don't think they'd accept any considering the kernel code isn't upstream)

jakecorrenti:
That works for me. I guess we don't really have a choice either way :)

jakecorrenti (dismissed):

tylerfanelli:
@jakecorrenti Can you re-review?

jakecorrenti (approved):
LGTM

PR #319, opened 2025-04-23
Introduce a display subsystem for GPU output

mtjhrc:
This PR adds initial support for the more traditional display output from the VM by basically having virtual displays (until now we only had cross-domain/wayland).

This implementes 2 display backends:
- `DisplayBackendNoop` - used by default, pretty much returns an error for every operation
- `DisplayBackendGtk` - implementation using GTK (works on Linux, to be tested on macOS)

This introduces 2 new public API function: 
- `int32_t krun_set_display_backend_gtk(uint32_t ctx_id)`
- `int32_t krun_set_display(uint32_t ctx_id, uint32_t display_id, uint32_t width, uint32_t height)`

Current limitations to be addressed by future PRs:
- add a virtio-input device and forward the events from the window to the guest
- improve the performance by using dmabufs instead copying it to the CPU
- send display resize events to the guest

You can test this using the newly added `--display` option in `chroot_vm` and running something graphical in the guest - the simplest option is probably `kmscube`:
`$ ./chroot_vm --display=0:1920:1080 rootfs_fedora kmscube`

### DIFF
--- a/src/devices/Cargo.toml
+++ b/src/devices/Cargo.toml
@@ -12,6 +12,7 @@ blk = []
 efi = ["blk", "net"]
 gpu = ["rutabaga_gfx", "thiserror", "zerocopy", "zerocopy-derive"]
 snd = ["pw", "thiserror"]
+gtk_display = ["gpu", "gtk4"]
 virgl_resource_map2 = []
 
 [dependencies]
@@ -29,6 +30,7 @@ virtio-bindings = "0.2.0"
 vm-memory = { version = ">=0.13", features = ["backend-mmap"] }
 zerocopy = { version = "0.6.3", optional = true }
 zerocopy-derive = { version = "0.6.3", optional = true }
+gtk4 = { version = "0.9.6", features = ["v4_14"], optional = true }
 
 arch = { path = "../arch" }
 utils = { path = "../utils" }
--- a/src/devices/src/display/gtk/worker.rs
+++ b/src/devices/src/display/gtk/worker.rs
@@ -1,0 +1,276 @@
+use std::os::fd::AsRawFd;
+
+use crate::display::gtk::DisplayEvent;
+use crate::display::{DisplayInfo, DisplayInfoList, MAX_DISPLAYS};
+use utils::pollable_channel::PollableChannelReciever;
+
+use gtk4::{
+    gdk,
+    gio::{ActionEntry, Cancellable, SimpleActionGroup},
+    glib::{self, source, Bytes, ControlFlow, IOCondition},
+    prelude::*,
+    AlertDialog, Align, Button, EventControllerMotion, HeaderBar, Overlay, Picture,
+    Revealer, RevealerTransitionType, Window,
+};
+
+struct Scanout {
+    window: Window,
+    width: i32,
+    height: i32,
+    format: gdk::MemoryFormat,
+    picture: Picture,
+}
+
+impl Scanout {
+    fn new(
+        title: String,
+        display_info: DisplayInfo,
+        width: i32,
+        height: i32,
+        format: gdk::MemoryFormat,
+    ) -> Self {
+        let header_bar = HeaderBar::new();
+        let window = Window::builder()
+            .title(title)
+            // remove the close button, since it is unclear what it should do
+            .deletable(false)
+            // Enforce a minimum window size:
+            .height_request(64)
+            .titlebar(&header_bar)
+            .build();
+
+        let actions = SimpleActionGroup::new();
+        actions.add_action_entries([
+            ActionEntry::builder("kill")
+                .activate(glib::clone!(
+                    #[weak]
+                    window,
+                    move |_, _, _| {
+                        let dialog = AlertDialog::builder()
+                            .buttons(["Kill VM", "Cancel"].as_ref())
+                            .default_button(0)
+                            .cancel_button(1)
+                            .modal(true)
+                            .message("Do you want to kill the VM?")
+                            .detail("WARNING: This may lead to loss of data or corruption of the VM image.")
+                            .build();
+                        dialog.choose(Some(&window), None::<&Cancellable>, |b| {
+                            if b.is_ok_and(|b| b == 0) {
+                                // SAFETY: Safe because we are terminating the process anyway.
+                                // We also use _exit during normal VM exit, so we don't clean up
+                                // ever anyway.
+                                unsafe { libc::_exit(125) };
+                            }
+                        });
+                    }
+                ))
+                .build(),
+            ActionEntry::builder("fullscreen")
+                .activate(glib::clone!(
+                    #[weak]
+                    window,
+                    move |_, _, _| {
+                        window.fullscreen();
+                    }
+                ))
+                .build(),
+            ActionEntry::builder("unfullscreen")
+                .activate(glib::clone!(
+                    #[weak]
+                    window,
+                    move |_, _, _| {
+                        window.unfullscreen();
+                    }
+                ))
+                .build(),
+        ]);
+        window.insert_action_group("scanout", Some(&actions));
+
+        let fullscreen_btn = Button::builder()
+            .icon_name("view-fullscreen")
+            .tooltip_text("Enter fullscreen mode")
+            .action_name("scanout.fullscreen")
+            .build();
+
+        let picture = Picture::builder()
+            // Set picture dimension to the requested display size,
+            // this will make the created window have the appropriate size
+            .width_request(display_info.width as i32)
+            .height_request(display_info.height as i32)
+            .build();
+
+        window.set_titlebar(Some(&header_bar));
+        header_bar.pack_start(&build_kill_vm_btn());
+        header_bar.pack_end(&fullscreen_btn);
+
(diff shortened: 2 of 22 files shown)


slp:
On macOS I'm getting this:

```
thread 'gtk display' panicked at /Users/slopezpa/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/gtk4-0.9.6/src/rt.rs:106:9:
assertion `left != right` failed: Attempted to initialize GTK on OSX from non-main thread
  left: 0
 right: 0
```

I haven't investigated further yet.

jakecorrenti (commented):
I can't speak to the logic of the subsystem since I'm not familiar with what's best practices here, but the code itself LGTM

src/devices/src/display/gtk/worker.rs
```diff
@@ -0,0 +1,278 @@
...
+        .label("Kill VM")
+        .action_name("scanout.kill")
+        .build();
+    kill_vm_btn.set_css_classes(&["destructive-action"]);
+    kill_vm_btn
+}
+
+fn build_overlay(window: &Window) -> Overlay {
+    let overlay_bar = HeaderBar::builder()
+        .visible(true)
```
bilelmoussaoui, line 160:
```suggestion
```
all widgets are visible by default

src/devices/src/display/gtk/worker.rs
```diff
@@ -0,0 +1,278 @@
...
+        .transition_type(gtk4::RevealerTransitionType::SwingDown)
+        .transition_duration(300)
+        .reveal_child(false)
+        .build();
+    revealer.set_child(Some(&overlay_bar));
+    overlay.add_overlay(&revealer);
+
+    let overlay_unfullscreen_btn = Button::builder()
+        .tooltip_text("Exit fullscreen mode")
+        .icon_name("view-restore")
```
bilelmoussaoui, line 178: where this icon comes from?

src/devices/src/display/gtk/worker.rs
```diff
@@ -0,0 +1,278 @@
...
+    overlay_bar.pack_start(&build_kill_vm_btn());
+    overlay_bar.pack_end(&overlay_unfullscreen_btn);
+    overlay_bar.set_show_title_buttons(false);
+
+    overlay
+}
+
+pub fn gtk_display_main_loop(rx: PollableChannelReciever<DisplayEvent>, displays: DisplayInfoList) {
+    gtk4::init().expect("Failed to initialize GTK");
+    let main_loop = glib::MainLoop::new(None, false);
```
bilelmoussaoui, line 215: any reason for running a manual main loop instead of using gtk::Application?

src/devices/Cargo.toml
```diff
@@ -29,6 +30,8 @@ virtio-bindings = "0.2.0"
 vm-memory = { version = ">=0.13", features = ["backend-mmap"] }
 zerocopy = { version = "0.6.3", optional = true }
 zerocopy-derive = { version = "0.6.3", optional = true }
+gtk4 = { version = "0.9.6", features = ["v4_14"], optional = true }
+glib = { version = "0.20.9", optional = true }
```
bilelmoussaoui, line 34: it is already pulled by gtk4, no need for a specific dep.

src/devices/src/display/gtk/worker.rs
```diff
@@ -0,0 +1,278 @@
...
+
+impl Scanout {
+    fn new(
+        title: String,
+        display_info: DisplayInfo,
+        width: i32,
+        height: i32,
+        format: gdk::MemoryFormat,
+    ) -> Self {
+        let header_bar = HeaderBar::builder().build();
```
bilelmoussaoui, line 32:
```suggestion
        let header_bar = HeaderBar::default();
```
is enough i guess

src/devices/src/display/gtk/worker.rs
```diff
@@ -0,0 +1,278 @@
...
+        let window = Window::builder()
+            .title(title)
+            // remove the close button, since it is unclear what it should do
+            .deletable(false)
+            // Enforce a minimum window size:
+            .height_request(64)
+            .titlebar(&header_bar)
+            .visible(true)
```
bilelmoussaoui, line 40: same as other widgets

src/devices/src/display/gtk/worker.rs
```diff
@@ -0,0 +1,278 @@
...
+            .build();
+
+        let actions = SimpleActionGroup::new();
+        actions.add_action_entries([
```
bilelmoussaoui, line 44: why use a separate action group ? i would recommend using a gtk::Window which is an gio::ActionMap already

src/devices/src/display/gtk/worker.rs
```diff
@@ -0,0 +1,278 @@
...
+                    window,
+                    move |_, _, _| {
+                        window.unfullscreen();
+                    }
+                ))
+                .build(),
+        ]);
+        window.insert_action_group("scanout", Some(&actions));
+
+        let fullscreen_btn = gtk4::Button::builder()
```
bilelmoussaoui, line 90: weird that some types are imported, some are not ;-)

src/devices/src/display/gtk/worker.rs
```diff
@@ -0,0 +1,278 @@
...
+        header_bar.pack_end(&fullscreen_btn);
+
+        let overlay = build_overlay(&window);
+        overlay.set_child(Some(&picture));
+        window.set_child(Some(&overlay));
+        window.set_visible(true);
+
+        // Unset the width/height after the window is created to allow resizing the window to
+        // be smaller
+        picture.set_size_request(-1, -1);
```
bilelmoussaoui, line 114: that looks like a hack.

bilelmoussaoui (commented):
some comments here and there

src/devices/src/display/gtk/worker.rs
mtjhrc: Yes it is. Any idea how to do this properly, though? 
I wanted to set the default window width/height,  I see `Window.set_default_size` exits, but what to set the dimensions to?  The window height should be the height of the image + height of the header bar, which I don't know how to obtain, and even if I did calculating this manually feels even more hacky and error prone. 
I don't see a way to set the default, but not minimal dimensions of the Picture widget (which would automatically create the window with correct size).

src/devices/src/display/gtk/worker.rs
mtjhrc: Well, originally I also wanted to also use a custom `MainContext` and only use gtk from this worker thread. This is technically a library and theoretically other glib/gtk things could be running (but more of a theoretical than a real problem I guess). I feel like the Rust bindings always want you to use the global context anyway (for example the `source::unix_fd_add_local`), or maybe I just don't understand, how to use it properly. I'm not really familiar with glib/gtk C API anyway. 

But it looks like this might be a moot point anyway, because on macOS it seems gtk has to be initialized on the main thread anyway...

src/devices/src/display/gtk/worker.rs
mtjhrc: I looked at this list, and picked icon names: https://specifications.freedesktop.org/icon-naming-spec/latest/#names
(but looking at it again, it seems this is from 2007, so I'm not sure if I should be rely on this :smile: )
Is it a bad assumption, that these icons will always exist?

src/devices/src/display/gtk/worker.rs
bilelmoussaoui: You will have to create a custom paintable, see https://github.com/gtk-rs/gtk4-rs/tree/main/examples/custom_paintable or other custom paintables examples.

src/devices/src/display/gtk/worker.rs
```diff
@@ -0,0 +1,278 @@
...
+
+use crate::display::gtk::DisplayEvent;
+use crate::display::{DisplayInfo, DisplayInfoList, MAX_DISPLAYS};
+use utils::pollable_channel::PollableChannelReciever;
+
+use glib::{source, ControlFlow, IOCondition};
+use gtk4::{
+    gdk,
+    gio::{ActionEntry, Cancellable, SimpleActionGroup},
+    prelude::{ActionMapExtManual, GtkWindowExt, WidgetExt},
```
bilelmoussaoui, line 11:
```suggestion
    prelude::*,
```
as stuff can move from ExtManual to Ext with time, so avoid potential code fixing with version bumps

src/devices/src/display/gtk/worker.rs
```diff
@@ -0,0 +1,278 @@
...
+}
+
+fn build_kill_vm_btn() -> Button {
+    let kill_vm_btn = Button::builder()
+        .label("Kill VM")
+        .action_name("scanout.kill")
+        .build();
+    kill_vm_btn.set_css_classes(&["destructive-action"]);
+    kill_vm_btn
+}
```
bilelmoussaoui, line 156: seems too much to do this in a separate function. Also use `add_css_class`, the function you are using removes all the pre-existing CSS classes in the widget if any.

mtjhrc:
Superseded by #349

PR #320, opened 2025-04-23
hvf: Add API to verify Nested Virt is supported

jakecorrenti:
Add an API to check if the current system supports Nested Virt on macOS.

### DIFF
--- a/src/hvf/src/lib.rs
+++ b/src/hvf/src/lib.rs
@@ -9,6 +9,9 @@
 #[allow(deref_nullptr)]
 pub mod bindings;
 
+#[macro_use]
+extern crate log;
+
 use bindings::*;
 
 #[cfg(target_arch = "aarch64")]
@@ -206,11 +209,27 @@ pub fn vcpu_set_vtimer_mask(vcpuid: u64, masked: bool) -> Result<(), Error> {
     }
 }
 
-pub struct HvfNestedBindings {
-    hv_vm_config_get_el2_supported:
-        libloading::Symbol<'static, unsafe extern "C" fn(*mut bool) -> hv_return_t>,
-    hv_vm_config_set_el2_enabled:
-        libloading::Symbol<'static, unsafe extern "C" fn(hv_vm_config_t, bool) -> hv_return_t>,
+/// Checks if Nested Virtualization is supported on the current system. Only
+/// M3 or newer chips on macOS 15+ will satisfy the requirements.
+pub fn check_nested_virt() -> Result<bool, Error> {
+    type GetEL2Supported =
+        libloading::Symbol<'static, unsafe extern "C" fn(*mut bool) -> hv_return_t>;
+
+    let get_el2_supported: Result<GetEL2Supported, libloading::Error> =
+        unsafe { HVF.get(b"hv_vm_config_get_el2_supported") };
+    if get_el2_supported.is_err() {
+        info!("cannot find hv_vm_config_get_el2_supported symbol");
+        return Ok(false);
+    }
+
+    let mut el2_supported: bool = false;
+    let ret = unsafe { (get_el2_supported.unwrap())(&mut el2_supported) };
+    if ret != HV_SUCCESS {
+        error!("hv_vm_config_get_el2_supported failed: {:?}", ret);
+        return Err(Error::NestedCheck);
+    }
+
+    Ok(el2_supported)
 }
 
 pub struct HvfVm {}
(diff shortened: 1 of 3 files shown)


jakecorrenti:
Related to comments in https://github.com/containers/podman/pull/25922

jakecorrenti:
On the krunkit side, my thought was we could allow the user to still do `--nested`, but if `krun_check_nested_virt` returns <1 we would just ignore the argument

tylerfanelli:
If it's not supported, couldn't we just have `krun_set_nested_virt` return an error code?

jakecorrenti:
> If it's not supported, couldn't we just have `krun_set_nested_virt` return an error code?

That was the original behavior. At this moment, if the user wants to enable nested virt, they either have to already know the requirements for the feature and do some system queries, or they have to try and call `krun_set_nested_virt` then check if it fails. This provides an easy way for the consumer of the API to programmatically check if they meet all necessary requirements before hand. Think of it like calling `KVM_CHECK_EXTENSION` on a specific capability before using the associated ioctl.

tylerfanelli:
Your stated use-case in krunkit (the main consumer of this API):

> On the krunkit side, my thought was we could allow the user to still do `--nested`, but if `krun_check_nested_virt` returns <1 we would just ignore the argument and print something to stdout or log it describing as such

What is the difference between `krun_check_nested_virt` returning < 1 as opposed to `krun_set_nested_virt` returning -1. The outcome would be the same IMO.

tylerfanelli:
> For consumers of krunkit, like Podman, having the command fail is not a graceful way to handle the issue. It's also not ideal UX to have each consumer do the same manual checks when we can provide it for them.

We can define this failure as its own error code and have krunkit wrap/handle it accordingly for podman.

src/hvf/src/lib.rs
```diff
@@ -206,11 +209,39 @@ pub fn vcpu_set_vtimer_mask(vcpuid: u64, masked: bool) -> Result<(), Error> {
...
+    };
+    let mut el2_supported: bool = false;
+    let ret = unsafe { (get_el2_supported)(&mut el2_supported) };
+    if ret != HV_SUCCESS {
+        error!("processor does not support the nested virtualization functionality");
+        return Err(Error::NestedCheck);
+    }
+
+    // nested virtualization requires macOS 15.0+
+    match std::process::Command::new("sw_vers")
```
slp, line 230: Instead of checking the macOS version, we should simply return `false` above if we can't find the `hv_vm_config_get_el2_supported` symbol.

src/libkrun/src/lib.rs
```diff
@@ -1062,6 +1062,19 @@ pub unsafe extern "C" fn krun_set_nested_virt(ctx_id: u32, enabled: bool) -> i32
...
+#[no_mangle]
+pub unsafe extern "C" fn krun_check_nested_virt() -> i32 {
+    #[cfg(target_os = "macos")]
+    match hvf::check_nested_virt() {
+        Ok(supp) => supp as i32,
+        Err(_) => -libc::EINVAL,
+    }
+
+    #[cfg(not(target_os = "macos"))]
+    -libc::EINVAL
```
slp, line 1075: Perhaps `EOPNOTSUP` would fit better here?

src/hvf/src/lib.rs
jakecorrenti: Fixed, thanks!

src/libkrun/src/lib.rs
jakecorrenti: I agree. Fixed.

jakecorrenti:
> Your stated use-case in krunkit (the main consumer of this API):
> 
> > On the krunkit side, my thought was we could allow the user to still do `--nested`, but if `krun_check_nested_virt` returns <1 we would just ignore the argument and print something to stdout or log it describing as such
> 
> What is the difference between `krun_check_nested_virt` returning <1 as opposed to `krun_set_nested_virt` returning <1. The outcome would be the same IMO.

In this scenario, you're right and the outcome would be the same. In the grand scheme of things, in my opinion, check if nested virt is supported and enabling nested virt should be two separate operations

tylerfanelli:
I think the check itself is fine. I'm just not convinced it warrants its own krun API, when `krun_set_nested_virt` could do the check itself and return `EOPNOTSUP` if it fails. krunkit could then ignore the arg and print some log message saying its not supported.

I try to avoid adding new krun APIs unless absolutely necessary though. Perhaps @slp has different thoughts?

tylerfanelli:
> In this scenario, you're right and the outcome would be the same. In the grand scheme of things, in my opinion, checking if nested virt is supported and enabling nested virt should be two separate operations.

But every time you check for nested virt and it is supported, you would then enable, correct? This suggests to me it can be one operation.

tylerfanelli:
> This would perhaps be useful if you are creating a VM configuration to save to a file using some sort of CLI, but not starting it the VM yet - you can tell the user nested virt is not supported on their system.

In this scenario, that VM configuration would only apply to the system you're running on. You wouldn't be able to pick up a config file built on a M3 (that would indicate nested virt is supported) and use it on a M2 system (it would wrongly indicate that nested virt is supported).

Maybe I'm thinking too much into it, however. I just think that this check is best served when actually wanting to enable nested virt.

jakecorrenti:
I will just remove the UAPI and put the nested check at the start of `krun_set_nested_virt`. If it fails I'll return EOPNOTSUPP. 

I'd like to get this moving forward and get nested support in Podman

slp:
Sorry for chiming it late. Both options are reasonable, the problem is that `krun_set_nested_virt` was already part of a release, so we shouldn't change its behavior. It's true that, most likely, `krunkit` is the only user of this particular API, but changing it sets a bad precedent IMO.

jakecorrenti:
> Sorry for chiming it late. Both options are reasonable, the problem is that `krun_set_nested_virt` was already part of a release, so we shouldn't change its behavior. It's true that, most likely, `krunkit` is the only user of this particular API, but changing it sets a bad precedent IMO.

This is something I didn't think about, but that's a really good point and I agree. 
> 
> In addition to this, most of the configuration calls we have right now doesn't check if the parameters are viable (they only they're valid in the most basic sense). This is something that we might consider changing in v2, when we revamp the API.

src/hvf/src/lib.rs
```diff
@@ -206,11 +209,27 @@ pub fn vcpu_set_vtimer_mask(vcpuid: u64, masked: bool) -> Result<(), Error> {
...
+        unsafe { HVF.get(b"hv_vm_config_get_el2_supported") };
+    if get_el2_supported.is_err() {
+        info!("cannot find hv_vm_config_get_el2_supported symbol");
+        return Ok(false);
+    }
+
+    let mut el2_supported: bool = false;
+    let ret = unsafe { (get_el2_supported.unwrap())(&mut el2_supported) };
+    if ret != HV_SUCCESS {
+        error!("processor does not support the nested virtualization functionality");
```
slp, line 228: If `hv_vm_config_get_el2_supported` we don't really know if it's supported or not, just that the function call failed. I think we should change this message to reflect that.

slp (changes requested):

tylerfanelli:
> Sorry for chiming it late. Both options are reasonable, the problem is that `krun_set_nested_virt` was already part of a release, so we shouldn't change its behavior. It's true that, most likely, `krunkit` is the only user of this particular API, but changing it sets a bad precedent IMO.
> 
> In addition to this, most of the configuration calls we have right now doesn't check if the parameters are viable (they only they're valid in the most basic sense). This is something that we might consider changing in v2, when we revamp the API.

Sounds good to me. Perhaps we'll re-examine this in v2.

src/hvf/src/lib.rs
jakecorrenti: Should be fixed now.

tylerfanelli (approved):

slp:
Confirmed to work as expected in M1 (detects that nested virt is not available) and M3 (detects that nested virt is available).

PR #321, opened 2025-04-28
Fix clippy warnings

tylerfanelli:


jakecorrenti (approved):
LGTM. 

`std::ptr::eq(ret, libc::MAP_FAILED)` feels weird to use when comparing error codes, but I understand it's a bindings nuance.

tylerfanelli:
> LGTM.
> 
> `std::ptr::eq(ret, libc::MAP_FAILED)` feels weird to use when comparing error codes, but I understand it's a bindings nuance.

I agree. [The docs](https://doc.rust-lang.org/std/ptr/fn.eq.html) even say:

`This is the same as using the == operator, but less generic`

PR #325, opened 2025-05-02
Preliminary support for AWS Nitro Enclaves

tylerfanelli:
This PR adds a new flavor, `libkrun-nitro`. The nitro flavor is expected to be run on AWS EC2 instances that support [AWS Nitro Enclaves](https://aws.amazon.com/ec2/nitro/nitro-enclaves/).

AWS Nitro Enclaves are fully isolated and highly constrained virtual machines that run on the Nitro hypervisor. CPUs and memory of an enclave are completely isolated from a parent instance and other users of a system, significantly reducing the attack surface and helping to prevent the stealing/tampering of data.

### Try for yourself
To demo on an AWS EC2 instance with Nitro Enclaves enabled:

- Clone repo, checkout the `nitro` branch
```
$ git clone https://github.com/tylerfanelli/libkrun.git
$ cd libkrun
$ git checkout nitro
```

- Build the `libkrun-nitro` flavor
```
$ make NITRO=1
$ make install NITRO=1
$ ln /usr/local/lib64/libkrun-nitro.so /usr/local/lib64/libkrun.so.1
```
- Build the `nitro` example
```
$ cd examples
$ make nitro
```
- Run the nitro example with an example EIF file found in `examples/test_data`
```
$ LD_LIBRARY_PATH=/usr/local/lib64 ./nitro test_data/hello.eif
```

This will run the enclave in debug mode, where you can examine debug output from the nitro enclave.

This will include Linux kernel logs:
```
[    0.000000] Booting Linux on physical CPU 0x0                                                                                                                                                                                                                                
.. snip ..
```

As well as a hello message from the enclave:
```
.. snip ..
[    0.062913] random: crng init done
[    0.063151] NSM RNG: returning rand bytes = 128
[    0.063450] NSM RNG: returning rand bytes = 128
Hello (from a nitro enclave!)
[    0.066406] reboot: Restarting system
```

### DIFF
--- a/examples/nitro.c
+++ b/examples/nitro.c
@@ -1,0 +1,136 @@
+/*
+ * This is an example implementing running an example AWS nitro enclave with
+ * libkrun.
+ *
+ * Given a nitro enclave image, run the image in a nitro enclave with 1 vCPU and
+ * 256 MiB of memory allocated.
+ */
+
+#include <errno.h>
+#include <stdio.h>
+#include <sys/socket.h>
+#include <libkrun.h>
+#include <getopt.h>
+#include <stdbool.h>
+#include <assert.h>
+
+#define MAX_ARGS_LEN 4096
+#ifndef MAX_PATH
+#define MAX_PATH 4096
+#endif
+
+static void print_help(char *const name)
+{
+    fprintf(stderr,
+        "Usage: %s EIF_FILE [COMMAND_ARGS...]\n"
+        "OPTIONS: \n"
+        "        -h    --help                Show help\n"
+        "\n"
+        "ENCLAVE_IMAGE:     The enclave image to run\n",
+        name
+    );
+}
+
+static const struct option long_options[] = {
+    { "help", no_argument, NULL, 'h' },
+    { NULL, 0, NULL, 0 }
+};
+
+struct cmdline {
+    bool show_help;
+    const char *eif_path;
+};
+
+bool parse_cmdline(int argc, char *const argv[], struct cmdline *cmdline)
+{
+    int c, option_index = 0;
+
+    assert(cmdline != NULL);
+
+    // set the defaults
+    *cmdline = (struct cmdline){
+        .show_help = false,
+        .eif_path = NULL,
+    };
+
+    // the '+' in optstring is a GNU extension that disables permutating argv
+    while ((c = getopt_long(argc, argv, "+h", long_options, &option_index)) != -1) {
+        switch (c) {
+        case 'h':
+            cmdline->show_help = true;
+            return true;
+        case '?':
+            return false;
+        default:
+            fprintf(stderr, "internal argument parsing error (returned character code 0x%x)\n", c);
+            return false;
+        }
+    }
+
+    if (optind < argc) {
+        cmdline->eif_path = argv[optind];
+        return true;
+    } else
+        fprintf(stderr, "Missing EIF_FILE argument");
+
+    return false;
+}
+
+int main(int argc, char *const argv[])
+{
+    int ctx_id, err, i;
+    struct cmdline cmdline;
+
+    if (!parse_cmdline(argc, argv, &cmdline)) {
+        putchar('\n');
+        print_help(argv[0]);
+        return -1;
+    }
+
+    if (cmdline.show_help){
+        print_help(argv[0]);
+        return 0;
+    }
+
+    // Set the log level to "off".
+    err = krun_set_log_level(0);
+    if (err) {
+        errno = -err;
+        perror("Error configuring log level");
+        return -1;
+    }
+
+    // Create the configuration context.
+    ctx_id = krun_create_ctx();
+    if (ctx_id < 0) {
+        errno = -ctx_id;
+        perror("Error creating configuration context");
+        return -1;
+    }
+
+    // Configure the number of vCPUs (1) and the amount of RAM (512 MiB).
+    if (err = krun_set_vm_config(ctx_id, 1, 512)) {
+        errno = -err;
+        perror("Error configuring the number of vCPUs and/or the amount of RAM");
+        return -1;
+    }
+
+    // Set the nitro enclave image specified on the command line.
+    if (err = krun_nitro_set_image(ctx_id, cmdline.eif_path,
+                                   KRUN_NITRO_IMG_TYPE_EIF)) {
+        errno = -err;
+        perror("Error configuring nitro enclave image");
+        return -1;
+
+    }
+
+    // Start and enter the microVM. Unless there is some error while creating the microVM
+    // this function never returns.
+    if (err = krun_start_enter(ctx_id)) {
+        errno = -err;
+        perror("Error creating the microVM");
+        return -1;
+    }
+
+    return 0;
+}
--- a/examples/nitro.c
+++ b/examples/nitro.c
@@ -8,16 +8,22 @@
 
 #include <errno.h>
 #include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
 #include <sys/socket.h>
+#include <sys/un.h>
 #include <libkrun.h>
 #include <getopt.h>
 #include <stdbool.h>
 #include <assert.h>
+#include <pthread.h>
 
 #define MAX_ARGS_LEN 4096
 #ifndef MAX_PATH
 #define MAX_PATH 4096
 #endif
+
+#define IPC_SOCK_PATH "/tmp/krun_nitro_example_ipc.sock"
 
 static void print_help(char *const name)
 {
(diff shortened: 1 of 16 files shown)


tylerfanelli:
Unsure why the [macOS-aarch64 / Code Quality (fmt, clippy) (pull_request)](https://github.com/containers/libkrun/actions/runs/14787814475/job/41519297904?pr=325) test is failing.

It cites `clippy` warnings in the `nitro-enclaves` crate, yet the test is being run with the following features:
```
cargo clippy --features efi,gpu -- -D warnings
```
The `nitro-enclaves` crate is an optional dependency and only enabled with the `nitro` feature, which is disabled in this test.

tylerfanelli:
cc @slp

jakecorrenti (commented):
Great work! I haven't tested the code yet, but so far things are looking good. I added a couple nits here and there.

In general, I would prefer if the commits were ordered such that changes in src/ come first, then include/, then finally examples/. My thought process behind this is it prevents the use of the API before the functionality is fully baked. Then, once we have the "backend" code, we can add the UAPI, then finally the example when everything's in a state where a working solution is possible. I think that would be a pretty invasive change here so I'm not asking you to do that, but I'm not sure how others feel about commit ordering in that sense.

src/nitro/src/lib.rs
```diff
@@ -12,3 +42,126 @@ pub struct NitroEnclave {
...
+        let ret = read(socket_fd, &mut buf);
+        let Ok(sz) = ret else {
+            break;
+        };
+        if sz != 0 {
+            let msg = String::from_utf8(buf[..sz].to_vec()).unwrap();
+            print!("{}", msg);
+        } else {
+            break;
+        }
```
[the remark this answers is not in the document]
tylerfanelli: Thanks, I like this better. I updated https://github.com/containers/libkrun/pull/325/commits/9e5e00a5f7b8631d586e65048aee4e9e72e485ef to reflect the change.

tylerfanelli:
> In general, I would prefer if the commits were ordered such that changes in src/ come first, then include/, then finally examples/. My thought process behind this is it prevents the use of the API before the functionality is fully baked. Then, once we have the "backend" code, we can add the UAPI, then finally the example when everything's in a state where a working solution is possible. I think that would be a pretty invasive change here so I'm not asking you to do that, but I'm not sure how others feel about commit ordering in that sense.

I would say this is up to personal preference. You could include the `examples/` and `include/` modifications along with the `src/` modifications if it logically makes sense, or you could split them up. However, this is **not** true if you make a change that would break a caller's use of an API. In that scenario, the changes to the caller must be also included (in `examples/`, for instance), as to not break the example/test.

examples/nitro.c
```diff
@@ -132,5 +204,12 @@ int main(int argc, char *const argv[])
         return -1;
     }
 
+    ret = pthread_join(thread, NULL);
+    if (ret < 0) {
+        perror("unable to join listener thread");
+        close(sock_fd);
+        exit(1);
+    }
+
```
jakecorrenti, line 220: Is there ever a scenario where this gets reached? Shouldn't `krun_start_enter` either exit above or not at all?

examples/nitro.c
```diff
@@ -124,6 +154,48 @@ int main(int argc, char *const argv[])
...
+        perror("Error listening on socket");
+        close(sock_fd);
+        exit(1);
+    }
+
+    // Configure the IPC socket to read output from the enclave. The "port"
+    // argument is ignored.
+    if (err = krun_add_vsock_port(ctx_id, 0, IPC_SOCK_PATH)) {
+        errno = -err;
+        perror("Error configuring enclave vsock");
```
jakecorrenti, line 194: Would we be leaking the socket fd here and in `krun_start_enter`?

examples/nitro.c
tylerfanelli: Usually yes, but nitro enclaves is a special case in this scenario (`krun_start_enter` does return). Not sure if that will remain true, but for now it is.

src/devices/Cargo.toml
```diff
@@ -13,6 +13,7 @@ efi = ["blk", "net"]
 gpu = ["rutabaga_gfx", "thiserror", "zerocopy", "zerocopy-derive"]
 snd = ["pw", "thiserror"]
 virgl_resource_map2 = []
+nitro = []
```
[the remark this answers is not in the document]
tylerfanelli: We could. I prefer `sev` and tdx rather than `amd-sev` and `intel-tdx` FWIW.

examples/nitro.c
tylerfanelli: What do you mean by "leaking"?

jakecorrenti: On line 165 you create a socket, but if `krun_add_vsock_port`  or `krun_start_enter` fails the socket never gets closed via `close(sock_fd)` and the resources are "leaked".

jakecorrenti: It should be cleaned up by the OS iirc but I think it's best practice to clean them up manually regardless.

examples/nitro.c
jakecorrenti: Right, thanks for the clarification!

examples/nitro.c
tylerfanelli: I suppose it would be good practice to close the socket fd, but it will be closed anyway when the process exits.

tylerfanelli: > It should be cleaned up by the OS iirc but I think it's best practice to clean them up manually regardless.

Sure, will do so.

jakecorrenti (dismissed):
Code LGTM

tylerfanelli:
@slp Can you review when you have the chance?

tylerfanelli:
@jakecorrenti @slp After rebasing, I'm seeing a weird issue. Each test is failing with the following:
```

error[E0308]: mismatched types
   --> src/devices/src/legacy/ioapic.rs:141:28
    |
141 |         vm.set_gsi_routing(&irq_routing[0])?;
    |            --------------- ^^^^^^^^^^^^^^^ expected `&FamStructWrapper<kvm_irq_routing>`, found `&kvm_irq_routing`
    |            |
    |            arguments to this method are incorrect
    |
    = note: expected reference `&vmm_sys_util::fam::FamStructWrapper<kvm_bindings::kvm_irq_routing>`
               found reference `&kvm_bindings::kvm_irq_routing`
note: method defined here
   --> /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/kvm-ioctls-0.22.0/src/ioctls/vm.rs:633:12
    |
633 |     pub fn set_gsi_routing(&self, irq_routing: &KvmIrqRouting) -> Result<()> {
    |            ^^^^^^^^^^^^^^^
```

I didn't change any of the `kvm-bindings`, `kvm-ioctls`, or `vmm-sys-utils` dependencies, and this is not an issue in the `main` branch (with the same dependency versions). Do you have any idea what could cause this?

jakecorrenti:
> @jakecorrenti @slp After rebasing, I'm seeing a weird issue. Each test is failing with the following:
> 
> ```
> 
> error[E0308]: mismatched types
>    --> src/devices/src/legacy/ioapic.rs:141:28
>     |
> 141 |         vm.set_gsi_routing(&irq_routing[0])?;
>     |            --------------- ^^^^^^^^^^^^^^^ expected `&FamStructWrapper<kvm_irq_routing>`, found `&kvm_irq_routing`
>     |            |
>     |            arguments to this method are incorrect
>     |
>     = note: expected reference `&vmm_sys_util::fam::FamStructWrapper<kvm_bindings::kvm_irq_routing>`
>                found reference `&kvm_bindings::kvm_irq_routing`
> note: method defined here
>    --> /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/kvm-ioctls-0.22.0/src/ioctls/vm.rs:633:12
>     |
> 633 |     pub fn set_gsi_routing(&self, irq_routing: &KvmIrqRouting) -> Result<()> {
>     |            ^^^^^^^^^^^^^^^
> ```
> 
> I didn't change any of the `kvm-bindings`, `kvm-ioctls`, or `vmm-sys-utils` dependencies, and this is not an issue in the `main` branch (with the same dependency versions). Do you have any idea what could cause this?

upstream kvm-bindings wrapped `kvm_irq_routing` in a `FamStructWrapper` as of this PR https://github.com/rust-vmm/kvm/pull/325. I can work on a fix right now

tylerfanelli:
Yes, but I never updated those dependencies? Regardless, it seems that #348 fixed it. Thanks!

jakecorrenti:
> Yes, but I never updated those dependencies? Regardless, it seems that #348 fixed it. Thanks!

You changed src/utils/Cargo.toml to use kvm-bindings >= 0.10 instead of 0.11 which probably triggered the update if I had to guess

tylerfanelli:
> > Yes, but I never updated those dependencies? Regardless, it seems that #348 fixed it. Thanks!
> 
> You changed src/utils/Cargo.toml to use kvm-bindings >= 0.10 instead of 0.11 which probably triggered the update if I had to guess

Yea, that must've been a mistake. Removed that change.

PR #326, opened 2025-05-02
Introduce a new krun_init_log() API

mtjhrc:
Introduce a new krun_init_log public API function to allow for much detailed configuration of logging by the application. The main improvment is the ability to specify a file descriptor to write the log to.

This can be tested using the added `chroot_vm` flags: `--log=/path/to/a/pipe` or  `--color-log=/path/to/a/pipe` (which will also display a formatting).

### DIFF
--- a/include/libkrun.h
+++ b/include/libkrun.h
@@ -26,6 +26,44 @@ extern "C" {
  */
 int32_t krun_set_log_level(uint32_t level);
 
+
+#define KRUN_LOG_TARGET_DEFAULT -1
+
+#define KRUN_LOG_LEVEL_OFF 0
+#define KRUN_LOG_LEVEL_ERROR 1
+#define KRUN_LOG_LEVEL_WARN 2
+#define KRUN_LOG_LEVEL_INFO 3
+#define KRUN_LOG_LEVEL_DEBUG 4
+#define KRUN_LOG_LEVEL_TRACE 5
+
+#define KRUN_LOG_STYLE_AUTO 0
+#define KRUN_LOG_STYLE_ALWAYS 1
+#define KRUN_LOG_STYLE_NEVER 2
+
+#define KRUN_LOG_OPTION_NO_ENV 1
+
+/**
+ * Initializes logging for the library.
+ *
+ * Arguments:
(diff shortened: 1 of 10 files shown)


include/libkrun.h
```diff
@@ -26,6 +26,45 @@ extern "C" {
...
+ *  "style"     - Enable/disable usage of terminal escape sequences (to display colors)
+ *                One of: KRUN_LOG_STYLE_{AUTO, ALWAYS, NEVER}.
+ *
+ *  "options"   - Bitmask of logging options, use 0 for default options.
+ *                KRUN_LOG_OPTION_NO_ENV to disallow environment variables to override these settings.
+ *
+ * Returns:
+ *  Zero on success or a negative error number on failure.
+ */
+int32_t krun_init_log(int target_fd, const char *level, uint32_t style, uint32_t options);
```
jakecorrenti, line 66: I know this is still a draft, but is this new API meant to be used instead of `krun_set_log_level`? i.e. are we deprecating the old API

mtjhrc: I'm not sure, should we? Both can work at the same time, but this one is more flexible. If we get to having a v2 API, we should probably only keep this one.

One confusing thing about the existing `krun_set_log_level`, is that if you don't call it `RUST_LOG` env variable will not work - you have to call something like `krun_set_log_level(0)` to initialize the logging.

jakecorrenti: Like you said, I think deprecating `krun_set_log_level` would make sense for v2.

If for some reason we **don't** deprecate it, I would like to see `krun_init_log` use an integer for the log level instead of a string to be uniform.

slp: I agree, please use an integer for "level", and allow arbitrarily large values to be treated as "trace".

mtjhrc: Sure, I can make it an integer. But just to clarify the reason I made level a string - it is so that you can pass in the arbitrary RUST_LOG options. e.g. level = "devices::virtio::gpu=trace". Though I'm not sure if this is useful, since you could still use the env variable to set that...

jakecorrenti (commented):
haven't had a chance to test it yet, but code LGTM!

PR #328, opened 2025-05-12
Bump version to 1.12.0

slp:
Set up the stage for a new release. Bumping minor since we're extending the API with a function to check whether nested virt is supported.


jakecorrenti (approved):
lgtm

PR #333, opened 2025-05-19
vmm: drop use of rangemap crate

slp:
It's not packaged in Fedora and I don't think it adds enough value to justify its addition.


slp:
cc/ @jakecorrenti

tylerfanelli (approved):

PR #335, opened 2025-05-26
Devices: clean up interrupt sending code

mtjhrc:
This PR aims to remove some of the duplicate code we have in regards to sending interrupts. 

We had around 15 copies of:
```Rust
self.interrupt_status
    .fetch_or(VIRTIO_MMIO_INT_VRING as usize, Ordering::SeqCst);
if let Some(intc) = &self.intc {
    intc.lock()
        .unwrap()
        .set_irq(self.irq_line, Some(&self.interrupt_evt))?;
```
But also for each such a copy of the interrupt sending code, we had to pass around 4 variables `(interrupt_status, irq_line, intc, interrupt_evt)` to each worker thread, which is quite silly.

This simplifies the handling, by creating a struct `InterruptTransport` which holds the fields and has a `signal_used_queue` method. The  `InterruptTransport` is managed by the `MmioTransport` passing it to the device upon activation.


src/vmm/src/builder.rs
```diff
@@ -1720,12 +1714,7 @@ fn attach_console_devices(
         .map_err(RegisterFsSigwinch)?;
 
     // The device mutex mustn't be locked here otherwise it will deadlock.
-    attach_mmio_device(
-        vmm,
-        "hvc0".to_string(),
-        MmioTransport::new(vmm.guest_memory().clone(), intc.clone(), console),
-    )
-    .map_err(RegisterFsDevice)?;
+    attach_mmio_device(vmm, "hvc0".to_string(), intc, console).map_err(RegisterFsDevice)?;
```
jakecorrenti: non-blocking: I wonder if adding a `RegisterConsoleDevice` error variant would be more clear here? I know its pre-existing so you don't need to change it here.

jakecorrenti (approved):
LGTM

src/vmm/src/builder.rs
mtjhrc: Fixed

slp:
Good stuff, let's get it merged.

slp (approved):

PR #342, opened 2025-06-04
Tweak the Makefile to decouple EFI and GPU features

ggoodman:
This tweaks the `EFI=1` make option so that it doesn't automatically bring along the `gpu` feature. Since `EFI` resets `FEATURE_FLAGS`, the test for `GPU` and `SND` (not implied by `efi` rust feature) are evaluated after.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,9 +27,6 @@ ifeq ($(SEV),1)
     INIT_SRC += $(SNP_INIT_SRC)
 	BUILD_INIT = 0
 endif
-ifeq ($(GPU),1)
-    FEATURE_FLAGS += --features gpu
-endif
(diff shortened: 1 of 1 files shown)


tylerfanelli:
Thanks for the contribution @ggoodman, can you add a `Signed-off-by` line to the commit?

ggoodman:
@tylerfanelli 👍🏼

jakecorrenti (approved):
Confirmed everything works with Podman still.

This will require a change in the various homebrew formulas to make sure we don't break users expecting the GPU functionality. cc/ @slp

slp:
> This will require a change in the various homebrew formulas to make sure we don't break users expecting the GPU functionality. cc/ @slp

Yup, but it's fine, I'll keep an eye on it.

PR #345, opened 2025-06-05
libkrun: add krun_add_vsock_unix_tunnel API

ggoodman:
Introduce a new API to create Unix socket tunnels between host and
guest over vsock connections. This enables transparent communication
through Unix domain sockets across the VM boundary.

The API supports both listen and connect modes:
- Listen mode: guest listens on Unix socket, connects to host via vsock
- Connect mode: guest connects to Unix socket, accepts from host via vsock

Implementation includes proper thread synchronization to prevent race
conditions during tunnel setup, ensuring proxy threads are ready
before the target process starts.

### DIFF
--- a/include/libkrun.h
+++ b/include/libkrun.h
@@ -527,6 +527,24 @@ int32_t krun_add_vsock_port2(uint32_t ctx_id,
                              uint32_t port,
                              const char *c_filepath,
                              bool listen);
+
+/**
+ * Adds a Unix socket tunnel between host and guest that creates Unix socket tunnels over vsock.
+ *
+ * Arguments:
+ *  "ctx_id"          - the configuration context ID.
+ *  "c_guest_filepath"- a null-terminated string representing the path of the UNIX socket in the guest.
+ *  "c_host_filepath" - a null-terminated string representing the path of the UNIX socket in the host.
+ *  "listen"          - true if guest should listen on the socket, false if guest should connect
+ *
+ * Returns:
+ *  Zero on success or a negative error number on failure.
+ */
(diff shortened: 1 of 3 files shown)


tylerfanelli:
@ggoodman Thanks for the contribution.

We have something similar to this with the [krun_add_vsock_port](https://github.com/containers/libkrun/blob/b80a18ac0fc3975a62a69bd7e9e0a81425421e9a/include/libkrun.h#L503-L514) API. This allows a guest to communicate with a IPC socket (on the host) through an `AF_VSOCK` socket. Seems these two are trying to solve the same issue. 
But furthermore, I've been thinking about this issue lately and wonder if a more general solution is possible.

libkrun implements [Transparent Socket Impersonation (AF_TSI)](https://github.com/containers/libkrun?tab=readme-ov-file#networking) for guest networking. From the host’s perspective, all connections come from/go to the libkrun-enabled runtime, and are visible in the network namespace of the runtime’s context.

I wonder if we can extend this to IPC as well. That is, if we could enable a krun API like `krun_forward_unix_ipc`. This would extend the concept of `AF_TSI` but for `AF_UNIX` sockets. With this configured, every time a guest application attempts to open an `AF_UNIX` socket for IPC, the guest kernel will "hijack" it and create an `AF_VSOCK` transparently that would forward data to the host (and back). This could allow "transparent" IPC between libkrun guests and processes on the host within the krun VM's namespace.

@slp WDYT?

jakecorrenti:
What's the status on this one as well?

ggoodman:
I think this should be dropped. I reimplemented the guest side in go and am not solid enough in C to have confidence in this PR.

PR #347, opened 2025-06-06
feat: krun_set_rootfs_read_only

### LINKED ISSUE #343: Read-only root filesystem / root filesystem from image
I'm hoping to use `libkrun` to build a simple sandboxing tool. I'm aware of the (awesome) `virtio-fs` mechanism for using a root fs on the host. My concern with that is that code within the guest could exhaust the disk space or inodes of the host, causing a denial-of-service. To mitigate that:
1. I could look for platform-specific solutions to make that portion of the host filesystem `ro`; or 
2. I could remount `/` as `ro` from within the guest as part of bootstrapping.

I don't think either of these are ideal, since the first is not portable and the 2nd relies on the guest to do the right thing.

I was hoping to leverage `krun_add_disk` as a way to bring a disk image as the root filesystem. This way I could use something like `EROFS` that is natively read-only and is supported by the kernel.

Can the team offer some guidance on how we might go about supporting root filesystems from images like `.erofs` and/or ways of making the root filesystem read-only? I think doing disk usage quotas seems totally unrealistic, so read-only is where my mind is going to target this higher level of resource isolation.

ggoodman:
Introduce `krun_set_rootfs_read_only` which affects the `"ro"` or `"rw"` flags being injected into kernel command-line arguments.

This should be mutually exclusive from `krun_set_kernel`. I'm happy to refine this to add some validation for that situation. I think I'd have to make the flag an `Option<bool>` or it wouldn't be possible to distinguish someone actively setting the rootfs to rw (by passing `false`) from the default situation in a later call to `krun_set_kernel`.

Fixes #343

### DIFF
--- a/src/vmm/src/builder.rs
+++ b/src/vmm/src/builder.rs
@@ -531,7 +531,9 @@ pub fn build_microvm(
     } else if let Some(cmdline) = &vm_resources.boot_config.kernel_cmdline_prolog {
         kernel_cmdline.insert_str(cmdline).unwrap();
     } else {
-        kernel_cmdline.insert_str(DEFAULT_KERNEL_CMDLINE).unwrap();
+        kernel_cmdline
+            .insert_str(format!("{DEFAULT_KERNEL_CMDLINE} rw"))
+            .unwrap();
     }
 
     #[cfg(not(feature = "tee"))]
--- a/include/libkrun.h
+++ b/include/libkrun.h
@@ -475,6 +475,19 @@ int32_t krun_set_kernel(uint32_t ctx_id,
                         const char *cmdline);
 
 /**
+ * Sets the read-only flag for the kernel boot arguments. This results in the root
+ * filesystem being mounted read-only.
+ * Arguments:
+ *  "ctx_id"        - the configuration context ID.
+ *  "read_only"     - true when the root filesystem should be mounted read-only.
+ *
+ * Returns:
+ *  Zero on success or a negative error number on failure.
+*/
+int32_t krun_set_rootfs_read_only(uint32_t ctx_id,
+                                  bool read_only);
+
+/**
  * Sets environment variables to be configured in the context of the executable.
  *
  * Arguments:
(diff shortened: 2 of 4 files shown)


src/vmm/src/builder.rs
```diff
@@ -531,7 +531,9 @@ pub fn build_microvm(
     } else if let Some(cmdline) = &vm_resources.boot_config.kernel_cmdline_prolog {
         kernel_cmdline.insert_str(cmdline).unwrap();
     } else {
-        kernel_cmdline.insert_str(DEFAULT_KERNEL_CMDLINE).unwrap();
+        kernel_cmdline
+            .insert_str(format!("{DEFAULT_KERNEL_CMDLINE} rw"))
+            .unwrap();
```
ggoodman, line 536: Since I took `rw` out of both architectures' `DEFAULT_KERNEL_CMDLINE`, I figured I had to put it back in here. Not sure how this is consumed though.

slp:
Thanks for the PR! A couple comments:

- Instead of trusting the guest kernel to not write to the virtiofs-exposed filesystem, we should implement a read-only mode in virtiofs itself, as [virtiofsd does](https://gitlab.com/virtio-fs/virtiofsd/-/blob/main/src/passthrough/read_only.rs?ref_type=heads).
- For consistency with other changes in the API, instead of adding `krun_set_rootfs_read_only` we should add `krun_set_root2`, with an additional `bool read_only` argument.

jakecorrenti:
@ggoodman are you still working on this?

ggoodman:
I'm using it in my fork for the time being. I think @slp's suggestion represents a better solution overall, however I don't have time to tackle it ATM.

I believe that libkrun would benefit from a ro rootfs solution but am not bought into this approach.

I'm happy if you want to close the PR.

jakecorrenti:
@ggoodman understood. For the time being I'm going to close the PR, but if you want to re-open it you're more than welcome. I still think it's a good idea overall

PR #349, opened 2025-06-06
Introduce an API for hooking up GPU display (callbacks)

mtjhrc:
This is a v2 of #319.

This PR introduces an API for implementing a display backend for our GPU device. The structs and function pointer definitions related to the API are defined in a separate header `libkrun_display.h`. This also introduces Rust bindings for this display definitions. The bindings are also used internally to implement `NoopDisplay` backend built into libkrun and by the GPU to call the display methods.

This also introduces a `krun_gtk_display` crate written in rust, my idea would be to keep that as a "reference" implementation for the GTK display, which could also be used by libkrun consumers. And finally this introduces `examples/gui_vm` executable which uses the reference  display implementation.

Note that this currently this works only on Linux, macOS should be supported by the display implementation, but is not supported the gpu. This is because venus doesn't have an implementation for reading the GPU resources back to the CPU memory, which this currently requires.  I'll put some basic macOS support in a separate PR.


examples/krun_gtk_display/src/scanout_paintable/imp.rs
```diff
@@ -0,0 +1,74 @@
...
+    const NAME: &'static str = "ScanoutPaintable";
+    type Type = super::ScanoutPaintable;
+    type Interfaces = (Paintable,);
+}
+
+impl ObjectImpl for ScanoutPaintable {
+    fn dispose(&self) {
+        debug!("ScanoutPaintable::dispose");
+    }
+}
```
bilelmoussaoui:
```suggestion
#[derive(glib::Properties)]
#[properties(wrapper_type = super::ScanoutPaintable)]
pub struct ScanoutPaintable {
    // Store the texture that this paintable will draw.
    pub texture: RefCell<Option<Texture>>,
    #[property(get, set, min = -1)]
    pub default_width: Cell<i32>,
    #[property(get, set, min = -1)]
    pub default_height: Cell<i32>,
    pub rect: Rect,
}

// Implement the GObjectSubclass trait for our MyPaintable struct.
#[glib::object_subclass]
impl ObjectSubclass for ScanoutPaintable {
    const NAME: &'static str = "ScanoutPaintable";
    type Type = super::ScanoutPaintable;
    type Interfaces = (Paintable,);
}

#[glib::derived_properties]
impl ObjectImpl for ScanoutPaintable {
    fn dispose(&self) {
        debug!("ScanoutPaintable::dispose");
    }
}
```

examples/krun_gtk_display/src/scanout_paintable/mod.rs
```diff
@@ -0,0 +1,68 @@
...
+    pub fn new(default_width: i32, default_height: i32) -> Self {
+        let self_: Self = glib::Object::new();
+        self_.init(default_width, default_height);
+        self_
+    }
+
+    fn init(&self, default_width: i32, default_height: i32) {
+        self.imp().default_width.replace(default_width);
+        self.imp().default_height.replace(default_height);
+    }
```
bilelmoussaoui:
```suggestion
    pub fn new(default_width: i32, default_height: i32) -> Self {
        glib::Object::builder().property("default-width", default_width).property("default-height", default_height).build()
    }
```

examples/krun_gtk_display/src/scanout_paintable/mod.rs
```diff
@@ -0,0 +1,68 @@
...
+
+    pub fn update(
+        &mut self,
```
bilelmoussaoui:
```suggestion
        &self,
```

examples/krun_gtk_display/src/display_backend.rs
```diff
@@ -0,0 +1,213 @@
...
+                scanout_id,
+                buffer,
+                rect,
+            })
+            .unwrap();
+        Ok(())
+    }
+}
+
+fn resource_format_into_gdk(format: ResourceFormat) -> Result<MemoryFormat, DisplayBackendError> {
```
bilelmoussaoui: This always returns Ok, maybe it makes sense as From impl or so.

examples/krun_gtk_display/src/display_worker.rs
```diff
@@ -0,0 +1,294 @@
...
+        });
+
+        window.add_action_entries([
+            ActionEntry::builder("fullscreen")
+                .activate(glib::clone!(
+                    #[weak]
+                    window,
+                    move |_, _, _| {
+                        window.fullscreen();
+                    }
```
bilelmoussaoui:
```suggestion
                    move |window, _, _| {
                        window.fullscreen();
                    }
```
works too, no?

bilelmoussaoui (commented):
Various GTK related comments.

examples/krun_gtk_display/src/display_backend.rs
mtjhrc: Oops, yeah I'm not sure why this returns Result... 

It cannot be an From/Into trait impl because neither the gdk MemoryFormat nor the ResourceFormat is defined in this crate. (And I don't want to introduce a gtk dependency for the `krun_display` crate where it is defined)

mtjhrc:
@bilelmoussaoui Resolved your comments. Thanks for all the feedback on using GTK :smile:

slp (approved):
LGTM, nice work here, @mtjhrc. I'm a bit concerned about the possible impact of introducing llvm as a build dependency may have on packaging, but we can always think of an alternative later, if needed.

PR #352, opened 2025-06-08
Fix the library install name and path

afbjorklund:
The install_name is supposed to be an absolute path, and use the soname rather than the developer symlink.

Avoids using `DYLD_FALLBACK_LIBRARY_PATH=/usr/local/lib`. Homebrew changes the install_name downstream, in /opt.

The default is the absolute path of the build directory (not wanted)

Typical error: `dyld[36629]: Library not loaded: libkrun-efi.dylib`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -93,7 +93,7 @@ ifeq ($(SEV),1)
 endif
 ifeq ($(OS),Darwin)
 ifeq ($(EFI),1)
-	install_name_tool -id libkrun-efi.dylib target/release/libkrun.dylib
+	install_name_tool -id $(PREFIX)/$(LIBDIR_$(OS))/$(KRUN_SONAME_$(OS)) target/release/libkrun.dylib
 endif
 	mv target/release/libkrun.dylib target/release/$(KRUN_BASE_$(OS))
 endif


Makefile
nirs, line 1: > The install_name is supposed to be an absolute path,
> and use the soname rather than the developer symlink.

This may be right since the current configuration requires modifications in code using the library. But it is strange that we need to use install_name_tool by default, I would expect cargo to build it correctly. May be we don't configure cargo corectly?

afbjorklund: The versions are also wrong, so it seems that it doesn't set anything?

`/usr/local/lib/libkrun-efi.1.dylib (compatibility version 0.0.0, current version 0.0.0)`

They are supposed to match the versions of the soname and library:

`libkrun-efi.1.dylib -> libkrun-efi.1.13.0.dylib`

afbjorklund: Apparently it (cdylib support) has been in progress for 5+ years: https://github.com/rust-lang/cargo/issues/5045

There is a workaround called cargo-c: https://dev.to/luzero/building-crates-so-they-look-like-c-abi-libraries-1ibn

slp (changes requested):
Thanks for the PR! I think that we should fix this the same way [we did in Linux](https://github.com/containers/libkrun/blob/885d642c4396a9db4f77f36691b041ae25317c37/src/libkrun/build.rs#L4), by using `rustc-cdylib-link-arg` to pass `-install_name`, `-current_version`, and `-compatibility_version`.

afbjorklund:
Added the linker flags. We still need to rename and relocate the library, though.

src/libkrun/build.rs
```diff
@@ -5,5 +5,10 @@ fn main() {
         std::env::var("CARGO_PKG_VERSION_MAJOR").unwrap()
     );
     #[cfg(target_os = "macos")]
+    println!(
+        "cargo:rustc-cdylib-link-arg=-Wl,-install_name,libkrun.{}.dylib,-compatibility_version,1.0.0,-current_version,{}.0.0",
+        std::env::var("CARGO_PKG_VERSION_MAJOR").unwrap(), std::env::var("CARGO_PKG_VERSION_MINOR").unwrap()
```
slp, line 11: Reading `ld`'s manual, I'd say `compatibility_version` should be `MAJOR.0.0`, and `current_version` `MAJOR.MINOR.RELEASE`.

Issue #357, opened 2025-06-15
Can't build main branch for macOS EFI=1 due to Nitro errors


afbjorklund:
Running `make EFI=1` fails:

```
   Compiling nitro v0.1.0 (/Users/afb/libkrun/src/nitro)
error[E0433]: failed to resolve: use of unresolved module or unlinked crate `nitro_enclaves`
 --> src/nitro/src/lib.rs:6:5
  |
6 | use nitro_enclaves::{
  |     ^^^^^^^^^^^^^^ use of unresolved module or unlinked crate `nitro_enclaves`
  |
  = help: if you wanted to use a crate named `nitro_enclaves`, use `cargo add nitro_enclaves` to add it to your `Cargo.toml`

error[E0432]: unresolved import `nitro_enclaves`
 --> src/nitro/src/lib.rs:6:5
  |
6 | use nitro_enclaves::{
  |     ^^^^^^^^^^^^^^ use of unresolved module or unlinked crate `nitro_enclaves`
  |
  = help: if you wanted to use a crate named `nitro_enclaves`, use `cargo add nitro_enclaves` to add it to your `Cargo.toml`

error[E0433]: failed to resolve: use of unresolved module or unlinked crate `nitro_enclaves`
 --> src/nitro/src/error.rs:3:5
  |
3 | use nitro_enclaves::launch::LaunchError;
  |     ^^^^^^^^^^^^^^ use of unresolved module or unlinked crate `nitro_enclaves`
  |
  = help: if you wanted to use a crate named `nitro_enclaves`, use `cargo add nitro_enclaves` to add it to your `Cargo.toml`

error[E0432]: unresolved import `nix::sys::socket::VsockAddr`
  --> src/nitro/src/lib.rs:13:70
   |
13 |         socket::{connect, socket, AddressFamily, SockFlag, SockType, VsockAddr as NixVsockAddr},
   |                                                                      ---------^^^^^^^^^^^^^^^^
   |                                                                      |
   |                                                                      no `VsockAddr` in `sys::socket`
   |                                                                      help: a similar name exists in the module: `SockAddr`
   |
   = help: consider importing this struct instead:
           vsock::VsockAddr
note: found an item that was configured out
  --> /Users/afb/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/nix-0.26.4/src/sys/socket/mod.rs:70:42
   |
70 | pub use crate::sys::socket::addr::vsock::VsockAddr;
[...]

PR #359, opened 2025-06-16
Fix main breakage

jakecorrenti:
Currently, the main branch cannot build on macOS with `EFI=1` due to nitro-related code being compile when it should not be.

Resolves: #357


jakecorrenti:
@tylerfanelli can you double check this to make sure nothing is broken on nitro

tylerfanelli (approved):
LGTM, thanks for addressing this. I probably shouldn't have silenced that clippy test on macOS, as that caught the underlying issue.

PR #361, opened 2025-06-20
arch: aarch64: Provide `initrd_size` to `arch_memory_regions`

RuoqingHe:
`arch_memory_regions` requires two arguments now while our tests still try to invoke it with only `size`, provide `initrd_size` to make tests work correctly.

Signed-off-by: Ruoqing He <heruoqing@iscas.ac.cn>


slp (approved):
LGTM, thanks!

PR #363, opened 2025-06-25
Add granular console APIs

jakecorrenti:
Currently, libkrun forces the user to manually change the source code if they are not building with EFI=1 and want to use the serial console instead of the virtio-console.

Provide an API, `krun_set_serial_console`, allowing the user to specify whether the serial console should be enabled. The serial console remains the default on EFI=1 use cases.

### DIFF
--- a/src/devices/src/virtio/block/device.rs
+++ b/src/devices/src/virtio/block/device.rs
@@ -79,7 +79,7 @@ impl DiskProperties {
 
         // We only support disk size, which uses the first two words of the configuration space.
         // If the image is not a multiple of the sector size, the tail bits are not exposed.
-        if disk_size % SECTOR_SIZE != 0 {
+        if !disk_size.is_multiple_of(SECTOR_SIZE) {
             warn!(
                 "Disk size {disk_size} is not a multiple of sector size {SECTOR_SIZE}; \
                  the remainder will not be visible to the guest."
--- a/src/devices/src/virtio/block/worker.rs
+++ b/src/devices/src/virtio/block/worker.rs
@@ -208,7 +208,7 @@ impl BlockWorker {
         match request_header.request_type {
             VIRTIO_BLK_T_IN => {
                 let data_len = writer.available_bytes() - 1;
-                if data_len % 512 != 0 {
+                if !data_len.is_multiple_of(512) {
                     Err(RequestError::InvalidDataLength)
                 } else {
                     writer
@@ -218,7 +218,7 @@ impl BlockWorker {
             }
             VIRTIO_BLK_T_OUT => {
                 let data_len = reader.available_bytes();
-                if data_len % 512 != 0 {
+                if !data_len.is_multiple_of(512) {
                     Err(RequestError::InvalidDataLength)
                 } else {
                     reader
--- a/src/devices/src/virtio/console/port_queue_mapping.rs
+++ b/src/devices/src/virtio/console/port_queue_mapping.rs
@@ -24,7 +24,7 @@ pub(crate) fn queue_idx_to_port_id(queue_index: usize) -> (QueueDirection, usize
         _ => queue_index / 2 - 1,
     };
 
-    let direction = if queue_index % 2 == 0 {
+    let direction = if queue_index.is_multiple_of(2) {
         QueueDirection::Rx
     } else {
         QueueDirection::Tx
--- a/src/rutabaga_gfx/src/virgl_renderer.rs
+++ b/src/rutabaga_gfx/src/virgl_renderer.rs
@@ -91,7 +91,7 @@ impl RutabagaContext for VirglRendererContext {
         if !fence_ids.is_empty() {
             return Err(RutabagaError::Unsupported);
         }
-        if commands.len() % size_of::<u32>() != 0 {
+        if !commands.len().is_multiple_of(size_of::<u32>()) {
             return Err(RutabagaError::InvalidCommandSize(commands.len()));
         }
         let dword_count = (commands.len() / size_of::<u32>()) as i32;


tylerfanelli (approved):
LGTM, though I wonder if it makes sense for this to be `krun_set_console` and the input to be an enum with two supported options:

```
KRUN_CONSOLE_SERIAL
KRUN_CONSOLE_VIRTIO
```

Leaves open the possibility of adding other options. Just something to consider.

slp:
> LGTM, though I wonder if it makes sense for this to be `krun_set_console` and the input to be an enum with two supported options:
> 
> ```
> KRUN_CONSOLE_SERIAL
> KRUN_CONSOLE_VIRTIO
> ```
> 
> Leaves open the possibility of adding other options. Just something to consider.

It's even more complex. You might want to enable one of them, or both. But, in addition of that, for the non-EFI case and when the kernel command line is not specified with `krun_set_kernel`, we need to specify which one is going to be put along with `console=` in the kernel command line.

Perhaps something like...

 ```
KRUN_CONSOLE_SERIAL
KRUN_CONSOLE_VIRTIO
KRUN_CONSOLE_BOTH_DEFAULT_SERIAL
KRUN_CONSOLE_BOTH_DEFAULT_VIRTIO
```

...?

**Comment by mtjhrc** [comment]
> > LGTM, though I wonder if it makes sense for this to be `krun_set_console` and the input to be an enum with two supported options:
> > ```
> > KRUN_CONSOLE_SERIAL
> > KRUN_CONSOLE_VIRTIO
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Leaves open the possibility of adding other options. Just something to consider.
> 
> It's even more complex. You might want to enable one of them, or both. But, in addition of that, for the non-EFI case and when the kernel command line is not specified with `krun_set_kernel`, we need to specify which one is going to be put along with `console=` in the kernel command line.
> 
> Perhaps something like...
> 
> ```
> KRUN_CONSOLE_SERIAL
> KRUN_CONSOLE_VIRTIO
> KRUN_CONSOLE_BOTH_DEFAULT_SERIAL
> KRUN_CONSOLE_BOTH_DEFAULT_VIRTIO
> ```
> 
Yeah I agree we should be more flexible whether you want both, but I'm not really a fan of having an enum like this.

I feel like ideally we should not have a console at all and let the user attach SERIAL/VIRTIO console(s) explicitly. 

Perhaps we could retrofit it to the current API... Can we maybe add an `krun_no_implicit_consoles(ctx)`? And then add functions for explicitly attaching the consoles? In the v2 API the `krun_no_implicit_consoles(ctx)` could then be the default.
Perhaps something like this:
```C
int32_t ctx = krun_create_ctx();
krun_no_implicit_consoles(ctx);
krun_add_serial_console(/*console id*/ 1, ...);
krun_add_virtio_console(/*console id*/ 2, ...);
krun_set_kernel_console(/*console id*/ 1); // console= in the kernel command line.
```
This would also make future API for attaching multiple  virtio-console(s) nicer.

Though this requires bigger changes in libkrun, I feel like it makes the API much more sane for the users (and extensible for us!).

include/libkrun.h
```diff
@@ -646,6 +646,16 @@ int32_t krun_nitro_set_image(uint32_t ctx_id, const char *image_path,
...
 
+/**
+ * Specify whether to enable the serial console
+ *
+ * Arguments:
+ *  "ctx_id" - the configuration context ID.
+ *  "enable" - whether to enable the serial console.
+ *  "debug"  - whether to redirect serial console output to stdout.
+ */
+int32_t krun_set_serial_console(uint32_t ctx_id, bool enable, bool debug);
```
mtjhrc: Instead of `bool debug` for stdout, It would be nicer to pass a file descriptor here, where `-1` is the default value. This shouldn't be hard to implement and it would be much more flexible.

mtjhrc (changes requested):

slp:
> Perhaps we could retrofit it to the current API... Can we maybe add an `krun_no_implicit_consoles(ctx)`? And then add functions for explicitly attaching the consoles? In the v2 API the `krun_no_implicit_consoles(ctx)` could then be the default. Perhaps something like this:
> 
> ```c
> int32_t ctx = krun_create_ctx();
> krun_no_implicit_consoles(ctx);
> krun_add_serial_console(/*console id*/ 1, ...);
> krun_add_virtio_console(/*console id*/ 2, ...);
> krun_set_kernel_console(/*console id*/ 1); // console= in the kernel command line.
> ```
> 
> This would also make future API for attaching multiple virtio-console(s) nicer.
> 
> Though this requires bigger changes in libkrun, I feel like it makes the API much more sane for the users (and extensible for us!).

Yes, this is probably the way. We *know* we're going to need a much more flexible console configuration for Android support (Android is a heavy user of virtio-console to communicate the guests with various processes in the host), and this would be a step in the right direction.

That said, perhaps it'd be safer to wait a bit to see exactly what the requirements are on the Android side and then propose an API?

mtjhrc:
> That said, perhaps it'd be safer to wait a bit to see exactly what the requirements are on the Android side and then propose an API?

Yeah maybe, but I don't think it will really be a problem. For now we could just have the same functionality as we already do just make it more explicit. We can also have multiple `krun_add_virtio_console_*` functions with different arguments.  The current console auto-configuration (which automatically configures the stdout/stdin/stderr as separate ports) could still exist but just be e.g. `krun_add_virtio_console_default(int console_id)`. 

Details of how to configure the console(s) for Android can still be figured out later. We could add a function like `krun_add_virtio_serial(..)` (I don't like the misleading QEMU name, but not sure how to call it). If we need to configure a multiport console with specific ports we can just have e.g.  `krun_add_virtio_console_multiport(int port_id)` and configure the ports using multiple calls to `krun_add_virtio_console_port(int console_id, const char* name, int port_fd)`...

jakecorrenti:
This **should** work. I did my best to test it on macOS and x86. I spent more time than I'd like to admit going back and forth on how this should all look, but I tried making a sane starting point that allows for future expansion (for Android, for example). If it doesn't work for someone, or I completely over looked something, let me know :)

include/libkrun.h
```diff
@@ -800,6 +800,26 @@ int32_t krun_set_kernel_console(uint32_t ctx_id, const char *console_id);
...
+ *
+ * Arguments:
+ *  "ctx_id"    - the configuration context ID.
+ *  "input_fd"  - file descriptor to use as input for console.
+ *  "output_fd" - file descriptor to use as output for console.
+ *
+ * Returns:
+ *  Zero on success or a negative error number on failure.
+ */
+int32_t krun_add_virtio_console_inout(uint32_t ctx_id, uint32_t input_fd, uint32_t output_fd);
```
mtjhrc: I think the file descriptor arguments should be the canonical `int` type (`c_int` in Rust), we already use this type for file descriptors in our API.

mtjhrc: This should really also accept a file descriptor for `stderr`, because this still keeps the logic that automatically creates different console ports for `stdout`/`stdin`/`stderr`. 

 Imagine if the stderr of the libkrun process is a pipe (`input_fd`/`output_fd` are a tty file descriptors provided by the user), in that case calling the `krun_add_virtio_console_inout`, would also add a console with 2 ports, the first one is the console the user wanted, but the second one is a "krun-stderr" with the global stderr, this seems confusing and unexpected for the user.

You can make the user provide `-1` as a file descriptor to disable the stderr. 

Btw the reason for creating the special `"krun-stdin"`/`"krun-stdout"`/`"krun-stderr"` console ports is that it allows for them to be seamlessly piped to/from the libkrun process (it interacts with our init process and redirects the application in the guest to the ports instead of the console).  They are separate ports so there is a way to signal EOF on just e.g. stdin without also closing stdout. (But if stdin/stdout is a console it makes sense to close the whole console).

jakecorrenti: I think that I addressed your comments. If I misunderstood something, let me know

src/devices/src/virtio/console/port.rs
```diff
@@ -55,8 +55,14 @@ impl Port {
...
-                input: Some(Arc::new(Mutex::new(input.unwrap()))),
-                output: Some(Arc::new(Mutex::new(output.unwrap()))),
+                input: match input {
+                    Some(i) => Some(Arc::new(Mutex::new(i))),
+                    None => None,
+                },
+                output: match output {
+                    Some(o) => Some(Arc::new(Mutex::new(o))),
+                    None => None,
+                },
```
mtjhrc:
```suggestion
                input: input.map(|input| Arc::new(Mutex::new(input))),
                output: output.map(|output| Arc::new(Mutex::new(output))),
```

Sooner or later clippy would probably get smart enough to compain here too :smile:

src/libkrun/src/lib.rs
```diff
@@ -1759,6 +1759,104 @@ pub unsafe extern "C" fn krun_nitro_set_start_flags(ctx_id: u32, start_flags: u6
...
+
+#[allow(clippy::missing_safety_doc)]
+#[no_mangle]
+pub unsafe extern "C" fn krun_add_virtio_console_default(ctx_id: u32) -> i32 {
+    krun_add_virtio_console_inout(ctx_id, STDIN_FILENO, STDOUT_FILENO, STDERR_FILENO)
+}
+
+#[allow(clippy::missing_safety_doc)]
+#[no_mangle]
+pub unsafe extern "C" fn krun_add_virtio_console_inout(
```
mtjhrc: I keep thinking about this, but I don't think this is a good name. The name makes it seem like this just adds a simple virtio console with a single port, but in fact this makes more assumptions, and depends on whether the file descriptors are tty or not. These defaults are only really good for our custom `init`.  The doc comment also really needs more details (I can add that in a separate PR though).

How do you and @slp feel about  naming it `krun_add_virtio_console_default_inout`? Or maybe we can just get rid of these `_inout` functions and just always provide the file descriptors to the `_default` functions.
So a user would do:
`krun_add_virtio_console_default(STDIN_FILENO, STDOUT_FILENO, STDERR_FILENO)`?

jakecorrenti: I think that `*_console_default(in, out, err)` is good. I'm not particularly attached to the name. 

In that case I'll squash the `*_console_default` and `*_console_inout` commits.

src/devices/src/virtio/console/port.rs
jakecorrenti: That's so much cleaner... Thanks :)

src/libkrun/src/lib.rs
jakecorrenti: Should be fixed and I also updated the doc comments

include/libkrun.h
```diff
@@ -853,6 +859,81 @@ int32_t krun_nitro_set_image(uint32_t ctx_id, const char *image_path,
...
+ * The function can be called multiple times for adding multiple virtio-console devices.
+ * In the guest, the consoles will appear in the same order as they are added (that is,
+ * the first added console will be "hvc0", the second "hvc1", ...). However, if the
+ * implicit console is not disabled via `krun_disable_implicit_console`, the first
+ * console created with the function will occupy the "hvc1" ID.
+ *
+ * When calling this function, a port is added for each stream (input, output, error).
+ * However, if the standard stream file descriptors are in fact not TTYs, an additional port
+ * will be created that interacts with the init process and redirects the application in the
+ * guest to the port instead of the console.
```
mtjhrc:
```suggestion
 * This function attaches a multi port virtio-console to the guest. If the input, output and error
 * file descriptors are TTYs, the device will be created with just a single console port (`err_fd`
 * is ignored in this case, because error output just goes to the TTY). For each of the non-TTY file
 * descriptors an additional non-console port is created ("krun-stdin"/"krun-stdout"/"krun-stderr").
 * The libkrun init process in the guest detects the existence of the additional ports and redirects
 * the stdin/stdout/stderr of the application in the guest appropriately.
```

jakecorrenti: Updated! Thanks for the help.

mtjhrc (approved):

jakecorrenti:
@slp @tylerfanelli @MatiasVara PTAL whenever you get the chance

slp (approved):
LGTM, thanks @jakecorrenti 

I've added a small commit fixing clippy errors to get the CI green.

jakecorrenti:
Thanks @slp!

jeromegn:
What settings are required for the file descriptors being passed in? Should they be non-blocking (or specifically not non-blocking?) For example, could it be a socket?

PR #365, opened 2025-06-30
clippy: Fix clippy `uninlined_format_args`

RuoqingHe:
Our workflow automatically bumps Rust toolchain to 1.88.0 as it was recently released, causing new clippy warnings showed up.


slp (approved):
LGTM too, thanks @RuoqingHe !

PR #367, opened 2025-07-01
net: support multiple network interfaces

slp:
So far, we were assuming the microVM will have a single network interface. But we have at least one new usecase (namely, running Android), that requires having multiple interfaces.

Refactor the API to support adding mulitple interfaces, while we keep compatibility with the old API.

While there, also allow the user to set up the network device flags.

### DIFF
--- a/examples/boot_efi.c
+++ b/examples/boot_efi.c
@@ -90,27 +90,6 @@ bool parse_cmdline(int argc, char *const argv[], struct cmdline *cmdline)
     return false;
 }
 
-int connect_to_passt(char const *socket_path)
-{
-    struct sockaddr_un addr;
-    int socket_fd = socket(AF_UNIX, SOCK_STREAM, 0);
-    if (socket_fd < 0) {
-        perror("Failed to create passt socket fd");
-        return -1;
-    }
-
-    memset(&addr, 0, sizeof(addr));
-    addr.sun_family = AF_UNIX;
-    strncpy(addr.sun_path, socket_path, sizeof(addr.sun_path) - 1);
-
-    if (connect(socket_fd, (const struct sockaddr *) &addr, sizeof(addr)) < 0) {
-        perror("Failed to bind passt socket");
-        return -1;
-    }
-
-    return socket_fd;
-}
-
 #define SHUTDOWN_SOCK_PATH  "/tmp/krun_shutdown.sock"
 
 void *listen_shutdown_request(void *opaque)
@@ -214,16 +193,11 @@ int main(int argc, char *const argv[])
         return -1;
     }
 
-    int passt_fd = connect_to_passt(cmdline.passt_socket_path);
-
-    if (passt_fd < 0) {
-      return -1;
-    }
-
-    if (err = krun_set_passt_fd(ctx_id, passt_fd)) {
-      errno = -err;
-      perror("Error configuring net mode");
-      return -1;
+    uint8_t mac[] = {0x5a, 0x94, 0xef, 0xe4, 0x0c, 0xee};
+    if (err = krun_add_net_unixgram(ctx_id, cmdline.passt_socket_path, -1, &mac[0], COMPAT_NET_FEATURES, NET_FLAG_VFKIT)) {
+        errno = -err;
+        perror("Error configuring net mode");
+        return -1;
     }
 
     int efd = krun_get_shutdown_eventfd(ctx_id);
--- a/include/libkrun.h
+++ b/include/libkrun.h
@@ -265,7 +265,114 @@ int32_t krun_add_virtiofs2(uint32_t ctx_id,
                            const char *c_path,
                            uint64_t shm_size);
 
+/* Send the VFKIT magic after establishing the connection,
+   as required by gvproxy in vfkit mode. */
+#define NET_FLAG_VFKIT 1 << 0
+
+/* Taken from uapi/linux/virtio_net.h */
+#define NET_FEATURE_CSUM 1 << 0
+#define NET_FEATURE_GUEST_CSUM 1 << 1
+#define NET_FEATURE_GUEST_TSO4 1 << 7
+#define NET_FEATURE_GUEST_TSO6 1 << 8
+#define NET_FEATURE_GUEST_UFO 1 << 10
+#define NET_FEATURE_HOST_TSO4 1 << 11
+#define NET_FEATURE_HOST_TSO6 1 << 12
+#define NET_FEATURE_HOST_UFO 1 << 14
+
+/* These are the features enabled by krun_set_passt_fd and krun_set_gvproxy_path. */
+#define COMPAT_NET_FEATURES NET_FEATURE_CSUM | NET_FEATURE_GUEST_CSUM | \
+                            NET_FEATURE_GUEST_TSO4 | NET_FEATURE_GUEST_UFO | \
+                            NET_FEATURE_HOST_TSO4 | NET_FEATURE_HOST_UFO
 /**
+ * Adds an independent virtio-net device connected to a
+ * unixstream-based userspace network proxy, such as passt or
+ * socket_vmnet.
+ *
+ * The "krun_add_net_*" functions can be called multiple times for
+ * adding multiple virtio-net devices. In the guest the interfaces
+ * will appear in the same order as they are added (that is, the
+ * first added interface will be "eth0", the second "eth1"...)
+ *
+ * If no network interface is added, libkrun will automatically
+ * enable the TSI backend.
+ *
+ * Arguments:
+ *  "ctx_id"   - the configuration context ID.
+ *  "c_path"   - a null-terminated string representing the path
+ *               for the unixstream socket where the userspace
+ *               network proxy is listening. Must be NULL if "fd"
+ *               is not -1.
+ *  "fd"       - a file descriptor for an already open unixstream
+ *               connection to the userspace network proxy. Must
+ *               be -1 if "c_path" is not NULL.
+ *  "c_mac"    - MAC address as an array of 6 uint8_t entries.
+ *  "features" - virtio-net features for the network interface.
+ *  "flags"    - generic flags for the network interface.
+ *
+ * Notes:
+ * The arguments "c_path" and "fd" are mutually exclusive. If using
+ * "fd", the socket must be already initialized and configured as
+ * the userspace network proxy requires.
+ * If no network devices are added, networking uses the TSI backend.
+ * This function should be called before krun_set_port_map.
+ *
+ * Returns:
+ *  Zero on success or a negative error number on failure.
+ */
+int32_t krun_add_net_unixstream(uint32_t ctx_id,
+                                const char *c_path,
+                                int fd,
+                                uint8_t *const c_mac,
+                                uint32_t features,
+                                uint32_t flags);
+
+/**
+ * Adds an independent virtio-net device with a unixgram-based
+ * backend, such as gvproxy or vmnet-helper.
+ *
+ * The "krun_add_net_*" functions can be called multiple times for
+ * adding multiple virtio-net devices. In the guest the interfaces
+ * will appear in the same order as they are added (that is, the
+ * first added interface will be "eth0", the second "eth1"...)
+ *
+ * If no network interface is added, libkrun will automatically
+ * enable the TSI backend.
+ *
+ * Arguments:
+ *  "ctx_id"   - the configuration context ID.
+ *  "c_path"   - a null-terminated string representing the path
+ *               for the unixstream socket where the userspace
+ *               network proxy is listening. Must be NULL if "fd"
+ *               is not -1.
+ *  "fd"       - a file descriptor for an already open unixstream
+ *               connection to the userspace network proxy. Must
+ *               be -1 if "c_path" is not NULL.
+ *  "c_mac"    - MAC address as an array of 6 uint8_t entries.
+ *  "features" - virtio-net features for the network interface.
+ *  "flags"    - generic flags for the network interface.
+ *
+ * Notes:
+ * The arguments "c_path" and "fd" are mutually exclusive. If using
+ * "fd", the socket must be already initialized and configured as
+ * the userspace network proxy requires.
+ * If no network devices are added, networking uses the TSI backend.
+ * This function should be called before krun_set_port_map.
+ * If using gvproxy in vfkit mode, NET_FLAG_VFKIT must be passed in
+ * "flags" when using "c_path" to indicate the connection endpoint.
+ *
+ * Returns:
+ *  Zero on success or a negative error number on failure.
+ */
+int32_t krun_add_net_unixgram(uint32_t ctx_id,
+                              const char *c_path,
+                              int fd,
+                              uint8_t *const c_mac,
+                              uint32_t features,
+                              uint32_t flags);
+
+/**
+ * DEPRECATED. Use krun_add_net_unixstream instead.
+ *
  * Configures the networking to use passt.
  * Call to this function disables TSI backend to use passt instead.
  *
@@ -283,6 +390,8 @@ int32_t krun_add_virtiofs2(uint32_t ctx_id,
 int32_t krun_set_passt_fd(uint32_t ctx_id, int fd);
 
 /**
+ * DEPRECATED. Use krun_add_net_unixgram instead.
+ *
  * Configures the networking to use gvproxy in vfkit mode.
  * Call to this function disables TSI backend to use gvproxy instead.
  *
@@ -659,9 +768,9 @@ int32_t krun_nitro_set_start_flags(uint32_t ctx_id, uint64_t start_flags);
  * Notes:
  *  This function only returns if an error happens before starting the microVM. Otherwise, the
  *  VMM assumes it has full control of the process, and will call to exit() with the workload's exit
- *  code once the microVM shuts down. If an error occurred before running the workload the process 
+ *  code once the microVM shuts down. If an error occurred before running the workload the process
  *  will exit() with an error exit code.
- * 
+ *
  * Error exit codes:
  *  125     - "init" cannot set up the environment inside the microVM.
  *  126     - "init" can find the executable to be run inside the microVM but cannot execute it.
--- a/src/devices/src/virtio/net/device.rs
+++ b/src/devices/src/virtio/net/device.rs
@@ -21,10 +21,7 @@ use std::path::PathBuf;
 use std::sync::atomic::AtomicUsize;
 use std::sync::Arc;
 use utils::eventfd::{EventFd, EFD_NONBLOCK};
-use virtio_bindings::virtio_net::{
-    VIRTIO_NET_F_CSUM, VIRTIO_NET_F_GUEST_CSUM, VIRTIO_NET_F_GUEST_TSO4, VIRTIO_NET_F_GUEST_UFO,
-    VIRTIO_NET_F_HOST_TSO4, VIRTIO_NET_F_HOST_UFO, VIRTIO_NET_F_MAC,
-};
+use virtio_bindings::virtio_net::VIRTIO_NET_F_MAC;
 use virtio_bindings::virtio_ring::VIRTIO_RING_F_EVENT_IDX;
 use vm_memory::{ByteValued, GuestMemoryError, GuestMemoryMmap};
 
@@ -65,8 +62,10 @@ unsafe impl ByteValued for VirtioNetConfig {}
 
 #[derive(Clone)]
 pub enum VirtioNetBackend {
-    Passt(RawFd),
-    Gvproxy(PathBuf),
+    UnixstreamFd(RawFd),
+    UnixstreamPath(PathBuf),
+    UnixgramFd(RawFd),
+    UnixgramPath(PathBuf, bool),
 }
 
 pub struct Net {
@@ -92,13 +91,13 @@ pub struct Net {
 
 impl Net {
     /// Create a new virtio network device using the backend
-    pub fn new(id: String, cfg_backend: VirtioNetBackend, mac: [u8; 6]) -> Result<Self> {
-        let avail_features = (1 << VIRTIO_NET_F_GUEST_CSUM)
-            | (1 << VIRTIO_NET_F_CSUM)
-            | (1 << VIRTIO_NET_F_GUEST_TSO4)
-            | (1 << VIRTIO_NET_F_HOST_TSO4)
-            | (1 << VIRTIO_NET_F_GUEST_UFO)
-            | (1 << VIRTIO_NET_F_HOST_UFO)
+    pub fn new(
+        id: String,
+        cfg_backend: VirtioNetBackend,
+        mac: [u8; 6],
+        features: u32,
+    ) -> Result<Self> {
+        let avail_features = features as u64
             | (1 << VIRTIO_NET_F_MAC)
             | (1 << VIRTIO_RING_F_EVENT_IDX)
             | (1 << VIRTIO_F_VERSION_1);
--- a/src/devices/src/virtio/net/unixgram.rs
+++ b/src/devices/src/virtio/net/unixgram.rs
@@ -11,13 +11,18 @@ use super::backend::{ConnectError, NetBackend, ReadError, WriteError};
 
 const VFKIT_MAGIC: [u8; 4] = *b"VFKT";
 
-pub struct Gvproxy {
+pub struct Unixgram {
     fd: RawFd,
 }
 
-impl Gvproxy {
-    /// Connect to a running gvproxy instance, given a socket file descriptor
-    pub fn new(path: PathBuf) -> Result<Self, ConnectError> {
+impl Unixgram {
+    /// Create the backend with a pre-established connection to the userspace network proxy.
+    pub fn new(fd: RawFd) -> Self {
+        Self { fd }
+    }
+
+    /// Create the backend opening a connection to the userspace network proxy.
+    pub fn open(path: PathBuf, send_vfkit_magic: bool) -> Result<Self, ConnectError> {
         let fd = socket(
             AddressFamily::Unix,
             SockType::Datagram,
@@ -37,7 +42,9 @@ impl Gvproxy {
         // allows the server to remove the socket after the connection.
         connect(fd, &peer_addr).map_err(ConnectError::Binding)?;
 
-        send(fd, &VFKIT_MAGIC, MsgFlags::empty()).map_err(ConnectError::SendingMagic)?;
+        if send_vfkit_magic {
+            send(fd, &VFKIT_MAGIC, MsgFlags::empty()).map_err(ConnectError::SendingMagic)?;
+        }
 
         // macOS forces us to do this here instead of just using SockFlag::SOCK_NONBLOCK above.
         match fcntl(fd, FcntlArg::F_GETFL) {
@@ -75,17 +82,17 @@ impl Gvproxy {
         }
 
         log::debug!(
-            "passt socket (fd {fd}) buffer sizes: SndBuf={:?} RcvBuf={:?}",
+            "network proxy socket (fd {fd}) buffer sizes: SndBuf={:?} RcvBuf={:?}",
             getsockopt(fd, sockopt::SndBuf),
             getsockopt(fd, sockopt::RcvBuf)
         );
 
-        Ok(Self { fd })
+        Ok(Self::new(fd))
     }
 }
 
-impl NetBackend for Gvproxy {
-    /// Try to read a frame from passt. If no bytes are available reports ReadError::NothingRead
+impl NetBackend for Unixgram {
+    /// Try to read a frame the proxy. If no bytes are available reports ReadError::NothingRead
     fn read_frame(&mut self, buf: &mut [u8]) -> Result<usize, ReadError> {
         let frame_length = match recv(self.fd, buf, MsgFlags::empty()) {
             Ok(f) => f,
@@ -97,19 +104,11 @@ impl NetBackend for Gvproxy {
                 return Err(ReadError::Internal(e));
             }
         };
-        debug!("Read eth frame from passt: {frame_length} bytes");
+        debug!("Read eth frame from proxy: {frame_length} bytes");
         Ok(frame_length)
     }
 
-    /// Try to write a frame to passt.
-    /// (Will mutate and override parts of buf, with a passt header!)
-    ///
-    /// * `hdr_len` - specifies the size of any existing headers encapsulating the ethernet frame,
-    ///   (such as vnet header), that can be overwritten. Must be >= PASST_HEADER_LEN.
-    /// * `buf` - the buffer to write to passt, `buf[..hdr_len]` may be overwritten
-    ///
-    /// If this function returns WriteError::PartialWrite, you have to finish the write using
-    /// try_finish_write.
+    /// Try to write a frame to the proxy.
     fn write_frame(&mut self, hdr_len: usize, buf: &mut [u8]) -> Result<(), WriteError> {
         let ret =
             send(self.fd, &buf[hdr_len..], MsgFlags::empty()).map_err(WriteError::Internal)?;
@@ -126,7 +125,7 @@ impl NetBackend for Gvproxy {
     }
 
     fn try_finish_write(&mut self, _hdr_len: usize, _buf: &[u8]) -> Result<(), WriteError> {
-        // The gvproxy backend doesn't do partial writes.
+        // The unixgram backend doesn't do partial writes.
         Ok(())
     }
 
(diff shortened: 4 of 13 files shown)


include/libkrun.h
```diff
@@ -265,7 +265,63 @@ int32_t krun_add_virtiofs2(uint32_t ctx_id,
...
 
+
+#define NET_CSUM 1 << 0
+#define NET_GUEST_CSUM 1 << 1
+#define NET_GUEST_TSO4 1 << 7
+#define NET_GUEST_TSO6 1 << 8
+#define NET_GUEST_UFO 1 << 10
+#define NET_HOST_TSO4 1 << 11
+#define NET_HOST_TSO6 1 << 12
+#define NET_HOST_UFO 1 << 14
```
nirs: Seems that we replicate the virtio features flags, would be useful to comment about it here.

include/libkrun.h
```diff
@@ -265,7 +265,63 @@ int32_t krun_add_virtiofs2(uint32_t ctx_id,
...
+ * If no network devices are added, networking uses the TSI backend.
+ * This function should be called before krun_set_port_map.
+ *
+ * Returns:
+ *  Zero on success or a negative error number on failure.
+ */
+int32_t krun_add_net_gvproxy(uint32_t ctx_id,
+                             char *c_path,
+                             uint8_t *const c_mac,
+                             uint32_t flags);
```
nirs: If we add new functions, maybe eliminate the gvproxy part? This is generic way that works with both gvproxy and vent. Also it will be useful to support also fd like vfkit. Maybe there can be another function using fd.

include/libkrun.h
```diff
@@ -265,7 +265,63 @@ int32_t krun_add_virtiofs2(uint32_t ctx_id,
...
+#define NET_HOST_TSO4 1 << 11
+#define NET_HOST_TSO6 1 << 12
+#define NET_HOST_UFO 1 << 14
+/**
+ * Adds an independent virtio-net device with the passt backend.
+ * Call to this function disables TSI backend.
+ *
+ * Arguments:
+ *  "ctx_id"  - the configuration context ID.
+ *  "fd"      - a file descriptor to communicate with passt.
```
nirs: Is this a socket fd? stream/datagram?

include/libkrun.h
```diff
@@ -265,7 +265,63 @@ int32_t krun_add_virtiofs2(uint32_t ctx_id,
...
+#define NET_GUEST_CSUM 1 << 1
+#define NET_GUEST_TSO4 1 << 7
+#define NET_GUEST_TSO6 1 << 8
+#define NET_GUEST_UFO 1 << 10
+#define NET_HOST_TSO4 1 << 11
+#define NET_HOST_TSO6 1 << 12
+#define NET_HOST_UFO 1 << 14
+/**
+ * Adds an independent virtio-net device with the passt backend.
+ * Call to this function disables TSI backend.
```
nirs: Can you call this twice to add 2 virtio-net drives? How do you identify the device after it was added?

src/devices/src/virtio/net/device.rs
```diff
@@ -92,13 +89,13 @@ pub struct Net {
...
-            | (1 << VIRTIO_NET_F_HOST_TSO4)
-            | (1 << VIRTIO_NET_F_GUEST_UFO)
-            | (1 << VIRTIO_NET_F_HOST_UFO)
+    pub fn new(
+        id: String,
+        cfg_backend: VirtioNetBackend,
+        mac: [u8; 6],
+        flags: u32,
+    ) -> Result<Self> {
+        let avail_features = flags as u64
```
nirs: This means we should not add new flags which are not virtio feature flags, or maybe we can add a mask to filter out anything which is not virtio feature flags here.

src/libkrun/src/lib.rs
```diff
@@ -752,6 +729,83 @@ pub unsafe extern "C" fn krun_set_data_disk(ctx_id: u32, c_disk_path: *const c_c
...
+    }
+    KRUN_SUCCESS
+}
+
+const NET_CSUM: u32 = 1 << 0;
+const NET_GUEST_CSUM: u32 = 1 << 1;
+const NET_GUEST_TSO4: u32 = 1 << 7;
+const NET_GUEST_UFO: u32 = 1 << 10;
+const NET_HOST_TSO4: u32 = 1 << 11;
+const NET_HOST_UFO: u32 = 1 << 14;
```
nirs: Can we define the flags once and make them available for both rust and the C header?

include/libkrun.h
```diff
@@ -265,7 +265,63 @@ int32_t krun_add_virtiofs2(uint32_t ctx_id,
...
+ *  "flags"   - network interface flags.
+ *
+ * Notes:
+ * If no network devices are added, networking uses the TSI backend.
+ * This function should be called before krun_set_port_map.
+ *
+ * Returns:
+ *  Zero on success or a negative error number on failure.
+ */
+int32_t krun_add_net_passt(uint32_t ctx_id,
```
mtjhrc: Just a thought: 
Instead of returning zero on success, we could return the `cfg.net_index` (but don't refer to it as index in the docs, it would just be an `id`). This wouldn't cost us anything implementation wise now, but later it opens up a door for us - if we wanted to change more parameters (e.g. the interface name ("eth0")), we could just add a function `krun_set_net_iface_name(uint32_t krun_ctx_id, uint32_t interface_id, const char* name)`.

src/libkrun/src/lib.rs
mtjhrc: For now we don't have a mechanism for doing so, so I'd say this is out of scope for this PR. But eventually we should find a way to do this, as we add more constants.

include/libkrun.h
slp: This poses an interesting question. While it's true that QEMU uses generic terms like `socket` for referring to the backend of a `chardev`, applications such as `passt` or `gvproxy` (in vfkit mode) behave in significantly different ways. Does `vnet` behave in exactly the same was as `gvproxy` in vfkit mode? If so, we could perhaps rename this function (and the backend) as `*_vfkit`.

include/libkrun.h
slp: The nature of the `fd` is defined by the userspace application we're connecting to (in this case, passt). This function doesn't aim to be generic, as realistically this isn't possible as exposed in the previous comment.

include/libkrun.h
slp: Yes, the first one will be `eth0`, the second `eth1`, etc... I'll document this in both functions.

include/libkrun.h
slp: I like the idea, but I think we should leave this for libkrun-v2.x.x, as we don't this behavior for any other function in the API.

include/libkrun.h
nirs: The same datagram socket works with vfkit and krunkit and the vmnet framework. Any frame sent to the socket is written to vment, and any frame read from vmnet is written to the socket.

vmnet-helper is basically a stupid proxy between the socket and the vment framework:
- https://github.com/nirs/vmnet-helper/blob/2d26ed25c87fbc0ded832211a04e2f4c13949313/helper.c#L719
- https://github.com/nirs/vmnet-helper/blob/2d26ed25c87fbc0ded832211a04e2f4c13949313/helper.c#L632

vment-helper also work with qemu using `-netdev drgram`:
https://github.com/nirs/vmnet-helper/blob/2d26ed25c87fbc0ded832211a04e2f4c13949313/vmnet/vm.py#L192

There is one little hack vment-helper does to enable vfkit and krunkit with unix socket, it drops the invalid VFKT packet used as a "handshake":
https://github.com/nirs/vmnet-helper/blob/2d26ed25c87fbc0ded832211a04e2f4c13949313/helper.c#L459

The handshake packet is not really needed and we can drop it from gvproxy and vfkit.

Using krun_add_net_vfkit() is also not great since vfkit is just a wrapper for the virtualization framework.

So I would go with `krun_add_net_unixgram()` or `krun_add_net_socket()`, document the semantics (proxy frames between the guest and the network stack) and mention gvproxy and vment-helper as programs that can work with this socket.

slp: > The handshake packet is not really needed and we can drop it from gvproxy and vfkit.

Do you mean that we can stop sending the magic from libkrun or that we should modify `gvproxy` to stop expecting for it?

nirs: > > The handshake packet is not really needed and we can drop it from gvproxy and vfkit.
> 
> Do you mean that we can stop sending the magic from libkrun or that we should modify `gvproxy` to stop waiting for it?

In vment we it is not needed - it waits for the first packet with MSG_PEEK to get the peer address, and then start the forward loop which will consume the packet from socket. gvproxy may need little work to remove the "handshake".

@cfergeau what do you think?

cfergeau: I'd be happy to drop the handshake packet, but at the time I did not manage to get something similar to `MSG_PEEK` to work.

slp: We still need to keep compatibility with vfkit, so instead of proposing a change to `gvproxy` I'm going to rename the function to `*_unixgram` and add a non-virtio flag to indicate we need to send the VFKIT magic. @nirs Does this sound reasonable to you?

cfergeau: https://github.com/cfergeau/gvisor-tap-vsock/commit/a13a3dca8c380c7f12325ea78c4150227c69e29c This is what I tried at the time.

nirs: > We still need to keep compatibility with vfkit, so instead of proposing a change to `gvproxy` I'm going to rename the function to `*_unixgram` and add a non-virtio flag to indicate we need to send the VFKIT magic. @nirs Does this sound reasonable to you?

Sounds good, krunkit can add an option to enable the flag or to disable the flag if we want to keep it compatible by default.

nirs: > [cfergeau/gvisor-tap-vsock@a13a3dc](https://github.com/cfergeau/gvisor-tap-vsock/commit/a13a3dca8c380c7f12325ea78c4150227c69e29c) This is what I tried at the time.

I think the issue is reading with empty buffer:

```
syscall.Recvfrom(int(fd), []byte{}, syscall.MSG_PEEK|syscall.MSG_TRUNC)
```
vmnet-helper use 64 bytes buffer, since this is the minimal frame size. If the frame is shorter we drop the packet since writing it to vment will fail anyway.

If you want to look only for the VFKT you can read 4 bytes.

mtjhrc: A way to solve this could also be to accept a file descriptor to a connected socket like we do with passt and let the user send or not send any magic handshake. 

If we do this we really should also rename `krun_add_net_passt_fd` to `krun_add_net_unixstream_fd` or something for consistency  - it corresponds to  `-netdev stream` in QEMU. (https://www.qemu.org/docs/master/system/devices/net.html#using-passt-as-the-user-mode-network-stack)

nirs: > A way to solve this could also be to accept a file descriptor to a connected socket like we do with passt and let the user send or not send any magic handshake.
> 
> If we do this we really should also rename `krun_add_net_passt_fd` to `krun_add_net_unixstream_fd` or something for consistency - it corresponds to `-netdev stream` in QEMU. (https://www.qemu.org/docs/master/system/devices/net.html#using-passt-as-the-user-mode-network-stack)

We should keep support for unix socket, since fd passing with vment requires sudo option `closefrom_override`, which is not possible to configure for some users due to limitations on sudoers configuration.

In minikube we switched the vfkit driver to use unixSocketPath so users that cannot configure passwordless sudo or `closefrom_overdide` can start the driver with a password. This is not great but better than not being able to use the vfkit driver with the vmnet network.

This is also required to keep backward compatibility with older gvproxy.

mtjhrc: > We should keep support for unix socket, since fd passing with vment requires sudo option closefrom_override

Sure but that's not really a problem of libkrun directly, we don't care how the file descriptor is obtained.  If necessary, it could just be the responsibility of the libkrun consumer to bind and connect to the socket on the filesystem and pass that file descriptor to libkrun...  Though admittedly this is a bit annoying for simple use cases but makes the API more flexible.

nirs: I think supporting both fd and unix socket in libkrun is the way to keep the library easy to use. At the backend level we already use a fd, obtained from the unix socket. We can add 2 "constructors" one for unix socket and one for fd to support both.

slp: Actually, while working on this PR, I've considered removing the socket initialization logic from `Gvproxy` and forcing users to do that work for us. But that would be a user-breaking change we can't do in v1.x.x, and keeping both mechanisms for compatibility seems like a waste of code. I'm inclined to keep the behavior of `krun_set_gvproxy_path`, with the rename and making the magic optional, and do something more generic in v2.x.x.

mtjhrc: > I think supporting both fd and unix socket in libkrun is the way to keep the library easy to use. 

Yes this is also a good argument, so I'm fine with this either way.

I was just throwing it out there that supporting only the file descriptor case would remove the need for conditionally sending the "VFKT" magic by libkrun (by making the library user do it). Is it a good idea? I'm not sure.

cfergeau: > > [cfergeau/gvisor-tap-vsock@a13a3dc](https://github.com/cfergeau/gvisor-tap-vsock/commit/a13a3dca8c380c7f12325ea78c4150227c69e29c) This is what I tried at the time.
> 
> I think the issue is reading with empty buffer:
> 
> ```
> syscall.Recvfrom(int(fd), []byte{}, syscall.MSG_PEEK|syscall.MSG_TRUNC)
> ```
> 

I tried it again, and this seems to work this time https://github.com/cfergeau/gvisor-tap-vsock/commit/8aace499d80ced197a90dd7f4b5fbfea86331bb0 
Maybe I missed the retry loop the first time.
I filed https://github.com/containers/gvisor-tap-vsock/issues/520 so that we get this into shape.

include/libkrun.h
slp: Added a comment referencing `uapi/linux/virtio_net.h`.

include/libkrun.h
slp: Renamed the function and the backend as `unixgram` and made sending the VFKIT magic optional.

include/libkrun.h
slp: Added to each function docs.

src/devices/src/virtio/net/device.rs
slp: The code is now sanitizing the flags.

include/libkrun.h
```diff
@@ -283,6 +364,8 @@ int32_t krun_add_virtiofs2(uint32_t ctx_id,
 int32_t krun_set_passt_fd(uint32_t ctx_id, int fd);
 
 /**
+ * DEPRECATED. Use krun_add_net_gvproxy instead.
```
mtjhrc: You forgot to rename the reference to function in the doc:
```suggestion
 * DEPRECATED. Use krun_add_net_unixgram instead.
```

include/libkrun.h
```diff
@@ -265,7 +265,88 @@ int32_t krun_add_virtiofs2(uint32_t ctx_id,
...
+ *  "vflags"  - virtio-net flags for the network interface.
+ *
+ * Notes:
+ * If no network devices are added, networking uses the TSI backend.
+ * This function should be called before krun_set_port_map.
+ *
+ * Returns:
+ *  Zero on success or a negative error number on failure.
+ */
+int32_t krun_add_net_passt(uint32_t ctx_id,
```
mtjhrc: Since we are also renaming the `krun_add_gvproxy_path` to be more general, we should do that here too.
QEMU uses the paramter `-netdev stream` (https://www.qemu.org/docs/master/system/devices/net.html#to-use-socket-based-passt-interface). 

So maybe `krun_add_net_unixstream_fd()`? I don't know... naming is hard... I don't care about the specific name, but for consistency we don't have to name it as `passt` specific.

include/libkrun.h
slp: Ouch, fixed, thanks!

include/libkrun.h
slp: But, while there are at least two userspace applications that can be used with the `Unixgram` backend, this one only works with `passt`, doesn't it? Reading the implementation I see it's reading the frame length from the wire, which I think it's pretty much specific of this process.

nirs: qemu -netdev socket also prefix the packet size to the packet. [socket_vmnet](https://github.com/lima-vm/socket_vmnet) is another application that can work with this format. It works with qemu -netdev socket and with lima vms, converting the prefix+payload to payload for apple virtualization datagram socket.

lima forwarding: https://github.com/lima-vm/lima/blob/73c5f9761bff67d9cb746e14a20a8d58bae34b39/pkg/vz/network_darwin.go#L67

If passt use the same format as qemu, than krunkit can also work with socket_vment, and we can use a generic name.

slp: I've confirmed that our backend works with `socket_vmnet` too, so I went ahead and renamed the backend as `Unixstream`. I also extended the API to support pointing to the network proxy with either the path or a file descriptor. I think this should cover all our use cases in a reasonable way.

include/libkrun.h
```diff
@@ -265,7 +265,107 @@ int32_t krun_add_virtiofs2(uint32_t ctx_id,
...
+ * Arguments:
+ *  "ctx_id"  - the configuration context ID.
+ *  "c_path"  - a null-terminated string representing the path
+ *              for the unixstream socket where the userspace
+ *              network proxy is listening.
+ *  "fd"      - a file descriptor for an already open unixstream
+ *              connection to the userspace network proxy.
+ *  "c_mac"   - MAC address as an array of 6 uint8_t entries.
+ *  "flags"   - generic flags for the network interface.
+ *  "vflags"  - virtio-net flags for the network interface.
```
nirs: 2 flags are not great. We can have one, and make sure that our flags do no clash with virtio flags. We can use a mask to pick our flags or virtio flags. Based on the value above, we can have:

    #define NET_VIRTIO_FLAGS_MASK 0x0000ffff
    #define NET_INTERNAL_FLAGS_MASK 0xffff0000

include/libkrun.h
```diff
@@ -265,7 +265,107 @@ int32_t krun_add_virtiofs2(uint32_t ctx_id,
...
+ * first added interface will be "eth0", the second "eth1"...)
+ *
+ * If no network interface is added, libkrun will automatically
+ * enable the TSI backend.
+ *
+ * Arguments:
+ *  "ctx_id"  - the configuration context ID.
+ *  "c_path"  - a null-terminated string representing the path
+ *              for the unixstream socket where the userspace
+ *              network proxy is listening.
```
nirs: Must be NULL if fd is not -1

include/libkrun.h
```diff
@@ -265,7 +265,107 @@ int32_t krun_add_virtiofs2(uint32_t ctx_id,
...
+ *  "fd"      - a file descriptor for an already open unixstream
+ *              connection to the userspace network proxy.
```
nirs: Must be -1 if c_path is not NULL

include/libkrun.h
```diff
@@ -265,7 +265,107 @@ int32_t krun_add_virtiofs2(uint32_t ctx_id,
...
+#define NET_VFLAG_GUEST_TSO4 1 << 7
+#define NET_VFLAG_GUEST_TSO6 1 << 8
+#define NET_VFLAG_GUEST_UFO 1 << 10
+#define NET_VFLAG_HOST_TSO4 1 << 11
+#define NET_VFLAG_HOST_TSO6 1 << 12
+#define NET_VFLAG_HOST_UFO 1 << 14
+/**
+ * Adds an independent virtio-net device connected to a
+ * unixstream-based userspace network proxy, such as passt or
+ * socket_vmnet.
```
nirs, line 289: Nice! I don't recommend using socket_vment, but having more options is great.

include/libkrun.h
```diff
@@ -265,7 +265,107 @@ int32_t krun_add_virtiofs2(uint32_t ctx_id,
...
+ *              network proxy is listening.
+ *  "fd"      - a file descriptor for an already open unixstream
+ *              connection to the userspace network proxy.
+ *  "c_mac"   - MAC address as an array of 6 uint8_t entries.
+ *  "flags"   - generic flags for the network interface.
+ *  "vflags"  - virtio-net flags for the network interface.
+ *
+ * Notes:
+ * The arguments "c_path" and "fd" are mutually exclusive. If using
+ * "fd" to point to the endpoint, the socket must be already
```
nirs: "if using fd to point to the endpoint" is not clear, maybe "if using fd".

include/libkrun.h
```diff
@@ -265,7 +265,107 @@ int32_t krun_add_virtiofs2(uint32_t ctx_id,
...
+ * Returns:
+ *  Zero on success or a negative error number on failure.
+ */
+int32_t krun_add_net_unixgram(uint32_t ctx_id,
+                              const char *c_path,
+                              int fd,
+                              uint8_t *const c_mac,
+                              uint32_t flags,
+                              uint32_t vflags);
+
```
nirs, line 372: Same issues as previous function.

Nice that the function are the same, if learn to use one you know how to use the other.

src/devices/src/virtio/net/device.rs
```diff
@@ -65,8 +62,10 @@ unsafe impl ByteValued for VirtioNetConfig {}
 
 #[derive(Clone)]
 pub enum VirtioNetBackend {
-    Passt(RawFd),
-    Gvproxy(PathBuf),
+    UnixstreamFd(RawFd),
+    UnixstreamPath(PathBuf),
+    UnixgramFd(RawFd),
+    UnixgramPath(PathBuf, bool),
```
nirs, line 68: Nice

src/devices/src/virtio/net/unixgram.rs
```diff
@@ -37,7 +42,9 @@ impl Gvproxy {
         // allows the server to remove the socket after the connection.
         connect(fd, &peer_addr).map_err(ConnectError::Binding)?;
 
-        send(fd, &VFKIT_MAGIC, MsgFlags::empty()).map_err(ConnectError::SendingMagic)?;
+        if vfkit_magic {
+            send(fd, &VFKIT_MAGIC, MsgFlags::empty()).map_err(ConnectError::SendingMagic)?;
+        }
```
nirs, line 47: Nice. Maybe rename to send_vfkit_magic to make it more clear.

src/devices/src/virtio/net/unixgram.rs
```diff
@@ -80,11 +87,11 @@ impl Gvproxy {
...
         );
 
-        Ok(Self { fd })
+        Ok(Self::new(fd))
     }
 }
 
-impl NetBackend for Gvproxy {
+impl NetBackend for Unixgram {
     /// Try to read a frame from passt. If no bytes are available reports ReadError::NothingRead
```
nirs: We can replace the leftover "passt" with "peer"

src/devices/src/virtio/net/unixgram.rs
```diff
@@ -126,7 +133,7 @@ impl NetBackend for Gvproxy {
     }
```
nirs, line 126: another leftover passt in line 111. We can also clean this up later to keep this change minimal, these leftovers were already there.

src/devices/src/virtio/net/unixstream.rs
```diff
@@ -1,40 +1,77 @@
...
+        .map_err(ConnectError::CreateSocket)?;
+        let peer_addr = UnixAddr::new(&path).map_err(ConnectError::InvalidAddress)?;
+        connect(fd, &peer_addr).map_err(ConnectError::Binding)?;
+
+        if let Err(e) = setsockopt(fd, sockopt::SndBuf, &(16 * 1024 * 1024)) {
+            log::warn!("Failed to increase SO_SNDBUF (performance may be decreased): {e}");
+        }
+
+        log::debug!(
+            "passt socket (fd {fd}) buffer sizes: SndBuf={:?} RcvBuf={:?}",
```
nirs: passt -> unixstream?

src/devices/src/virtio/net/unixstream.rs
```diff
@@ -1,40 +1,77 @@
...
+    pub fn new(fd: RawFd) -> Self {
+        if let Err(e) = setsockopt(fd, sockopt::SndBuf, &(16 * 1024 * 1024)) {
             log::warn!("Failed to increase SO_SNDBUF (performance may be decreased): {e}");
         }
 
         log::debug!(
-            "passt socket (fd {passt_fd}) buffer sizes: SndBuf={:?} RcvBuf={:?}",
-            getsockopt(passt_fd, sockopt::SndBuf),
-            getsockopt(passt_fd, sockopt::RcvBuf)
+            "passt socket (fd {fd}) buffer sizes: SndBuf={:?} RcvBuf={:?}",
```
nirs: passt -> unixstream

src/libkrun/src/lib.rs
```diff
@@ -752,8 +731,172 @@ pub unsafe extern "C" fn krun_set_data_disk(ctx_id: u32, c_disk_path: *const c_c
...
+    | NET_VFLAG_HOST_UFO;
+#[cfg(feature = "net")]
+const NET_ALL_VFLAGS: u32 = NET_VFLAG_CSUM
+    | NET_VFLAG_GUEST_CSUM
+    | NET_VFLAG_GUEST_TSO4
+    | NET_VFLAG_GUEST_TSO6
+    | NET_VFLAG_GUEST_UFO
+    | NET_VFLAG_HOST_TSO4
+    | NET_VFLAG_HOST_TSO6
+    | NET_VFLAG_HOST_UFO;
```
nirs: It is not clear what are the compat flags - maybe add a comment? Seems that the difference are NET_VFLAG_GUEST_TSO6 and NET_VFLAG_HOST_TSO6.

We can use this to make this more clear:

```rust
const NET_ALL_VFLAGS: u32 = NET_COMPAT_VFLAGS
    | NET_VFLAG_GUEST_TSO6
    | NET_VFLAG_HOST_TSO6;
```

src/libkrun/src/lib.rs
```diff
@@ -752,8 +731,172 @@ pub unsafe extern "C" fn krun_set_data_disk(ctx_id: u32, c_disk_path: *const c_c
...
+        match CStr::from_ptr(c_path).to_str() {
+            Ok(path) => Some(PathBuf::from(path)),
+            Err(_) => None,
+        }
+    } else {
+        None
+    };
+
+    if fd >= 0 && path.is_some() {
+        return -libc::EINVAL;
```
nirs, line 806: EINVAL without an error message is hard to work with. We have 4 different EINVAL errors for this function. How the user can tell what is the issue?

Can we have libkrun_error() to get the last error message? Probably not in the scope of this change, and I did not check the rest of the code.

src/vmm/src/vmm_config/net.rs
```diff
@@ -41,51 +44,28 @@ type Result<T> = result::Result<T, NetworkInterfaceError>;
...
+    pub list: VecDeque<Arc<Mutex<Net>>>,
 }
 
 impl NetBuilder {
     /// Creates an empty list of Network Devices.
     pub fn new() -> Self {
         NetBuilder {
             // List of built network devices.
-            net_devices: Vec::new(),
+            list: VecDeque::new(),
```
nirs, line 55: Why do we need a deque? we only append to the end.

nirs (commented):
So much nicer with generic names!

include/libkrun.h
mtjhrc: Seems fine to me.
I'm not totally opposed to having single flags argument, but I'd say I prefer having two arguments like Sergio has done. The `vflags` change which features the guest sees are supported, whereas the `flags` are just flags that can also be used for arbitrary things to tweak ("quirks"), like the `VFKIT` magic header.

src/libkrun/src/lib.rs
mtjhrc: > Can we have libkrun_error() to get the last error message? Probably not in the scope of this change, and I did not check the rest of the code. 

Yeah, so our error reporting is not great and unfortunately we don't have anything like `libkrun_error()`, nor do we define our own error codes which would help here. Currently the only way to see what went wrong for example with `krun_start_enter` is to check the log (which has to be enabled). Not great I know, but that is really something we need to change in v2 API more globally (otherwise it would get really confusing how errors work).

include/libkrun.h
slp: I also prefer having two separate arguments. It's easier to maintain and less error-prone.

include/libkrun.h
slp: That's nicer, added.

include/libkrun.h
slp: That's nicer, added.

include/libkrun.h
slp: Changed, thanks.

include/libkrun.h
slp: Also fixed for this one.

src/devices/src/virtio/net/unixgram.rs
slp: Renamed, thanks.

src/devices/src/virtio/net/unixgram.rs
slp: I've replaced them with "network proxy" or "proxy".

src/devices/src/virtio/net/unixgram.rs
slp: As the commit has grown substantially, I also updated this ones too.

src/devices/src/virtio/net/unixstream.rs
slp: I've replaced them with "network proxy".

src/devices/src/virtio/net/unixstream.rs
slp: Ditto.

src/libkrun/src/lib.rs
slp: Added a comment clarifying why that's there, thanks!

src/vmm/src/vmm_config/net.rs
slp: Did that for consistency with other `vmm_config` device lists.

include/libkrun.h
nirs: 2 options is more error prone for users of the library. The main issue is
using NET_FLAG_VFKIT with vflags, enabling NET_VFLAG_CSUM by mistake silently, or using NET_VFLAG_CSUM for flags, enabling vfkit_magic silently. The way to prevent this to make NET_FLAG_VFKIT unique, for example (1 << 16). With this we can detect if the flags are used incorectly.

But, if vflags and flags do not clash, there is no reason to use 2 variables and not need to check for incorrect usage. Users do not need to think which flags belongs to which variable and they cannot use the function in the wrong way.

The price for making usage of the library safer, masking flags before using them in the library.

include/libkrun.h
```diff
@@ -265,7 +265,109 @@ int32_t krun_add_virtiofs2(uint32_t ctx_id,
                            const char *c_path,
                            uint64_t shm_size);
 
+/* Send the VFKIT magic after establishing the connection,
+   as required by gvproxy in vfkit mode. */
+#define NET_FLAG_VFKIT 1 << 0
+
+/* Taken from uapi/linux/virtio_net.h */
+#define NET_VFLAG_CSUM 1 << 0
```
nirs: If we go with 2 flag arguments let's make the constants and the flags names more clear using "features" and "flags".

```
#define NET_FEATURE_CSUM 1<<0
...
 *  "features"  - virtio-net features for the network interface.
 *  "flags"     - generic flags for the network interface.
```

Switching the order because the virtio features are more important and flags are usually the last argument.

slp: Sounds reasonable to me, I've just changed it.

slp:
@mtjhrc could you please take another look?

examples/boot_efi.c
```diff
@@ -23,6 +23,8 @@
 #define MAX_PATH 4096
 #endif
 
+#define COMPAT_NET_FLAGS NET_CSUM | NET_GUEST_CSUM | NET_GUEST_TSO4 | NET_GUEST_UFO | NET_HOST_TSO4 | NET_HOST_UFO
```
mtjhrc: You forgot to update this to the new constant names

examples/boot_efi.c
```diff
@@ -220,10 +222,11 @@ int main(int argc, char *const argv[])
       return -1;
     }
 
-    if (err = krun_set_passt_fd(ctx_id, passt_fd)) {
-      errno = -err;
-      perror("Error configuring net mode");
-      return -1;
+    uint8_t mac[] = {0x5a, 0x94, 0xef, 0xe4, 0x0c, 0xee};
+    if (err = krun_add_net_passt(ctx_id, passt_fd, &mac[0], COMPAT_NET_FLAGS)) {
```
mtjhrc: You forgot to update this to the new function name and signature

slp:
> Please fix the `boot_efi` example to not refer to non-existant functions/constants. Otherwise everything looks good.

Good catch, thanks. I've fixed that and also moved COMPAT_NET_FEATURES to `libkrun.h`, which seems like a better place for it.

mtjhrc (approved):

PR #368, opened 2025-07-01
net: implement a backend for tap devices

slp:
Implement a new backend that allows network interfaces to use a tap device on the host to read/write frames.

### DIFF
--- a/src/devices/src/virtio/net/tap.rs
+++ b/src/devices/src/virtio/net/tap.rs
@@ -1,0 +1,128 @@
+use libc::{
+    c_char, c_int, ifreq, IFF_NO_PI, IFF_TAP, IFF_VNET_HDR, TUN_F_CSUM, TUN_F_TSO4, TUN_F_TSO6,
+    TUN_F_UFO,
+};
+use nix::fcntl::{fcntl, open, FcntlArg, OFlag};
+use nix::sys::stat::Mode;
+use nix::unistd::{read, write};
+use nix::{ioctl_write_int, ioctl_write_ptr};
+use std::os::fd::{AsRawFd, RawFd};
+use std::{io, mem, ptr};
+use virtio_bindings::virtio_net::{
+    VIRTIO_NET_F_GUEST_CSUM, VIRTIO_NET_F_GUEST_TSO4, VIRTIO_NET_F_GUEST_TSO6,
+    VIRTIO_NET_F_GUEST_UFO,
+};
+
+use super::backend::{ConnectError, NetBackend, ReadError, WriteError};
+
+ioctl_write_ptr!(tunsetiff, b'T', 202, c_int);
+ioctl_write_int!(tunsetoffload, b'T', 208);
+ioctl_write_ptr!(tunsetvnethdrsz, b'T', 216, c_int);
+
+pub struct Tap {
+    fd: RawFd,
+}
+
+impl Tap {
+    /// Create an endpoint using the file descriptor of a tap device
+    pub fn new(tap_name: String, vnet_features: u64) -> Result<Self, ConnectError> {
+        let fd = match open("/dev/net/tun", OFlag::O_RDWR, Mode::empty()) {
+            Ok(fd) => fd,
+            Err(err) => return Err(ConnectError::OpenNetTun(err)),
+        };
+
+        let mut req: ifreq = unsafe { mem::zeroed() };
+
+        unsafe {
+            ptr::copy_nonoverlapping(
+                tap_name.as_ptr() as *const c_char,
+                req.ifr_name.as_mut_ptr(),
+                tap_name.len(),
+            );
+        }
+
+        req.ifr_ifru.ifru_flags = IFF_TAP as i16 | IFF_NO_PI as i16 | IFF_VNET_HDR as i16;
+
+        let mut offload_flags: u64 = 0;
+        if (vnet_features & (1 << VIRTIO_NET_F_GUEST_CSUM)) != 0 {
+            offload_flags |= TUN_F_CSUM as u64;
+        }
+        if (vnet_features & (1 << VIRTIO_NET_F_GUEST_TSO4)) != 0 {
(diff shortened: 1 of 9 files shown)


slp:
@MatiasVara @mtjhrc @jakecorrenti @tylerfanelli PTAL

src/devices/src/virtio/net/tap.rs
```diff
@@ -0,0 +1,110 @@
...
+            }
+
+            // TODO(slp): replace hardcoded vnet size with cons
+            if let Err(err) = tunsetvnethdrsz(fd, &12) {
+                return Err(ConnectError::TunSetVnetHdrSz(io::Error::from(err)));
+            }
+
+            if let Err(err) = tunsetoffload(fd, (TUN_F_CSUM | TUN_F_TSO4 | TUN_F_TSO6) as u64) {
+                return Err(ConnectError::TunSetOffload(io::Error::from(err)));
+            }
```
mtjhrc: So this is interesting. Here you hardcore the actual offload flags, for the tap. But you still let the user specify the offload flags visible to the guest irrespective of this.  In the case of passt/gvproxy/etc... that makes sense, because we don't control what they offload or how they are configured.

But maybe here we can make this dependent of `features` arg of `krun_add_net_tap`?  
`VIRTIO_NET_F_CSUM` => `TUN_F_CSUM` , 
`VIRTIO_NET_F_HOST_TSO4` => `TUN_F_TSO4`
`VIRTIO_NET_F_HOST_TSO6` => `TUN_F_TSO6`

src/libkrun/src/lib.rs
```diff
@@ -878,6 +878,52 @@ pub unsafe extern "C" fn krun_add_net_unixgram(
...
+pub unsafe extern "C" fn krun_add_net_tap(
+    ctx_id: u32,
+    c_tap_name: *const c_char,
+    c_mac: *const u8,
+    features: u32,
+    flags: u32,
+) -> i32 {
+    if cfg!(not(feature = "net")) {
+        return -libc::ENOTSUP;
+    }
```
mtjhrc: `return -libc::ENOTSUP;` is unreachable. The whole function is only compiled when "net" is enabled.

I think you should just repeat the whole function and also return -libc::ENOTSUP on macOS:

```Rust
#[cfg(not(all(target_os = "linux", feature = "net")))]
pub unsafe extern "C" fn krun_add_net_tap(
    ctx_id: u32,
    c_tap_name: *const c_char,
    c_mac: *const u8,
    features: u32,
    flags: u32,
) -> i32 {
        -libc::ENOTSUP;
}
```

mtjhrc (changes requested):
Looks good, please fix the `cfg!` comment. 
I'm not sure how we want to handle enabling the features/offload, if we hardcode them then why do we let the user specify the virtio features.

src/devices/src/virtio/net/tap.rs
```diff
@@ -0,0 +1,110 @@
...
+ioctl_write_ptr!(tunsetvnethdrsz, b'T', 216, c_int);
+
+pub struct Tap {
+    fd: RawFd,
+}
+
+impl Tap {
+    /// Create an endpoint using the file descriptor of a tap device
+    pub fn new(tap_name: String) -> Result<Self, ConnectError> {
+        let fd = unsafe { libc::open(c"/dev/net/tun".as_ptr() as *const _, O_RDWR) };
```
jakecorrenti: purely out of curiosity, why couldn't we use `File::open` and then do `<file>.as_raw_fd()` which is safe Rust?

jakecorrenti (commented):
Other than the comments that Matej already made, this looks good to me.

Apologies for the delay, this one fell through the cracks.

src/devices/src/virtio/net/tap.rs
mtjhrc: Personally, I think I'd rather use `open` from `nix` (which is also safe rust) and we already have that crate, even in this file. It seems more fitting when you want a file descriptor.
(But not that it really matters here).

src/devices/src/virtio/net/tap.rs
slp: What we set with `tunsetoffload` is the features that the tap backend in libkrun can support. As it doesn't involved any third component (neither the guest nor another application, such as passt), it only needs to be aligned with us.

src/libkrun/src/lib.rs
slp: Good point. I've also removed the check from other methods were we also had redundant checks.

src/devices/src/virtio/net/tap.rs
slp: Nice suggestion, I've replaced `libc::open` with `nix::open`.

src/devices/src/virtio/net/tap.rs
mtjhrc: I'm not sure if I understand all the offload flags correctly, but to me it seems this **must** involve the guest. Like if we pass the `VIRTIO_NET_F_GUEST_CSUM` to the guest, it can give libkrun unchecksummed packets, in which case libkrun better enable `TUN_F_CSUM` on the tap, no? We also directly pass the 12 byte header between the guest and the tap...

My point is that to me it seems that if we hardcode the `tunsetoffload` flags we should also hardcode the virtio features that correspond the the tap offload or make both configurable or alternatively derive  the `tunsetoffload` from the virtio features.

They do the latter in firecracker:
https://github.com/firecracker-microvm/firecracker/blob/7ae9f570c019664166b5addb9a872c49586f9fda/src/vmm/src/devices/virtio/net/device.rs#L768-L804

(In my previous comment I mentioned the wrong virtio features that effect this, we actually want the GUEST_* features not HOST_*)

slp: Makes sense and works fine. I've updated the PR to do that. Thanks!

slp:
@mtjhrc PTAL

slp:
@mtjhrc do we need any more changes to this PR?

PR #369, opened 2025-07-02
fdt: Gate aarch64 fdt behind a module

RuoqingHe:
fdt should be a generic module which allows fdt for different architectures, gate aarch64 fdt behind fdt module to avoid possible conflicts.

Signed-off-by: Ruoqing He <heruoqing@iscas.ac.cn>


RuoqingHe:
Could we allow rebase merge, our Integration Tests are kind of flaky (which means I'll need to force push several times to get a green Integration test result) 🫠

slp:
LGTM, thanks!

> Could we allow rebase merge, our Integration Tests are kind of flaky (which means I'll need to force push several times to get a green Integration test result) 🫠

It's allowed. I'm also worried about the Integration Tests, but can't reproduce the issue locally. I'm starting to suspect it's a problem with nested virt on the github runners...

slp (approved):

Issue #318, opened 2025-04-18
Support RISC-V 64-bit architecture


alicedreamed:
Hi libkrun team, do we have a plan for RISC-V arch support, or are we willing to extend architecture support to RISC-V?

jakecorrenti:
Hi @alicedreamed! 

AFAIK there's no roadmap for adding RISC-V arch support at this moment. I'm not opposed to extending architecture support, though. 

@slp @tylerfanelli thoughts? I'm not sure if this has been a discussion previously

tylerfanelli:
I don't think any RISC-V development boards support the Hypervisor extension at the moment, correct?

jakecorrenti:
Apparently the Hypervisor Extension just makes Virtualization more efficient. Supposedly RISC-V cores with M or U modes support virt by running the Guest in a special mode and trapping privileged instructions.

tylerfanelli:
I'm certainly interested in RISC-V support, but it's a matter of getting the hardware and testing KVM first. It seems kvmtool has a riscv version. @alicedreamed Do you know of the level of support right now?

alicedreamed:
> I'm certainly interested in RISC-V support, but it's a matter of getting the hardware and testing KVM first. It seems kvmtool has a riscv version. [@alicedreamed](https://github.com/alicedreamed) Do you know of the level of support right now?

I haven't dig into the status of kvmtool's RISC-V support. AFAIK the hardware side is not ready for this, while I saw rust-vmm and cloud-hypervisor have now supported RISC-V, and I happen to know the author of those works, maybe he could help us with our CI infrastructure.

If we are willing to accept RISC-V support, I could do the coding work here

davidlt:
There is no hardware currently available that support hypervisor extension. The recent SiFive HiFive Premier P550 and Milk-V Megrez (the same SoC from ESWIN) has non-ratified v0.6.1 version of Hypervisor extension. I think, vendor is emulating some missing bits in lower firmware to have it working. For the full high-performance VM you probably also need IOMMU, a new advanced interrupt controller, etc. There is no SoC with all of that.

RuoqingHe:
Thanks @alicedreamed for mentioning, if there's any problem with RISC-V CI, free feel to engage me.

> For the full high-performance VM you probably also need IOMMU, a new advanced interrupt controller, etc. There is no SoC with all of that.

[...]

PR #370, opened 2025-07-03
Enable libkrun on RISC-V 64-bit architecture

alicedreamed:
As discussed in #318, this patch set enables libkrun to work on RISC-V 64-bit platforms.


tylerfanelli:
This is a great contribution, thanks @alicedreamed! How would I go about testing?

src/arch/src/riscv64/linux/regs.rs
```diff
@@ -0,0 +1,180 @@
...
+pub enum Error {
+    /// Failed to set core register (PC, A0, A1 or general purpose ones).
+    SetCoreRegister(kvm_ioctls::Error),
+}
+type Result<T> = result::Result<T, Error>;
+
+// This macro gets the offset of a structure (i.e `str`) member (i.e `field`) without having
+// an instance of that structure.
+#[macro_export]
+macro_rules! _offset_of {
```
jakecorrenti: how is this different from `std::mem::offset_of!` ?

src/devices/src/legacy/kvmaia.rs
```diff
@@ -0,0 +1,180 @@
...
+use crate::Error as DeviceError;
+
+use kvm_ioctls::{DeviceFd, VmFd};
+use utils::eventfd::EventFd;
+
+pub struct KvmAia {
+    _device_fd: DeviceFd,
+
+    // /// AIA device properties, to be used for setting up the fdt entry
+    // properties: [u64; 4],
```
jakecorrenti: Can we remove this?

src/devices/src/legacy/kvmaia.rs
```diff
@@ -0,0 +1,180 @@
...
+    /// Number of CPUs handled by the device
+    vcpu_count: u32,
+}
+
+impl KvmAia {
+    pub fn new(vm: &VmFd, vcpu_count: u32) -> Self {
```
jakecorrenti: Can you return `Result<Self>` here to can handle the various `.unwrap()`s? This would follow suit with other implementations like `kvmioapic.rs`

src/devices/src/legacy/kvmaia.rs
```diff
@@ -0,0 +1,180 @@
...
+            flags: 0,
+        };
+        device_fd.set_device_attr(&attr).unwrap();
+
+        // Helpers to calculate address and attribute of IMSIC of each vCPU
+        let riscv_imsic_addr_of = |cpu_index: u32| -> u64 {
+            arch::riscv64::layout::IMSIC_START
+                + (cpu_index * kvm_bindings::KVM_DEV_RISCV_IMSIC_SIZE) as u64
+        };
+        let riscv_imsic_attr_of = |cpu_index: u32| -> u64 { cpu_index as u64 + 1 };
```
jakecorrenti: I don't think these are necessary considering they're only used in one place

src/devices/src/legacy/kvmaia.rs
```diff
@@ -0,0 +1,180 @@
...
+            group: kvm_bindings::KVM_DEV_RISCV_AIA_GRP_CTRL,
+            attr: u64::from(kvm_bindings::KVM_DEV_RISCV_AIA_CTRL_INIT),
+            addr: 0,
+            flags: 0,
+        };
+        device_fd.set_device_attr(&attr).unwrap();
+
+        Self {
+            _device_fd: device_fd,
+            // properties: [dist_addr, dist_size, redists_addr, redists_size],
```
jakecorrenti: Can we remove this as well?

src/arch/src/riscv64/linux/regs.rs
```diff
@@ -0,0 +1,180 @@
...
+// including config registers, core registers and timer registers.
+//
+// The register of core registers are wrapped in the `user_regs_struct` structure. See:
+// https://elixir.bootlin.com/linux/v6.10/source/arch/riscv/include/uapi/asm/kvm.h#L62
+
+// Get the ID of a register
+#[macro_export]
+macro_rules! riscv64_core_reg {
+    ($reg_type: tt, $offset: tt) => {
+        // The core registers of an riscv64 machine are represented
```
jakecorrenti: can we get rid of all this?

jakecorrenti (commented):
Thanks for the PR @alicedreamed! Just a couple comments.

Additionally, since there's no riscv64 runner in GitHub is it possible to do something with QEMU on an x86 runner to try and run code quality checks on each PR?

src/arch/src/riscv64/linux/regs.rs
```diff
@@ -0,0 +1,113 @@
...
+type Result<T> = result::Result<T, Error>;
+
+#[macro_export]
+macro_rules! riscv_offset_of {
+    ($reg_struct:ty, $field:ident) => {
+        std::mem::offset_of!($reg_struct, $field)
+    };
+    ($outer_reg_struct:ty, $outer_field:ident, $($inner_reg_struct:ty, $inner_field:ident), +) => {
+        std::mem::offset_of!($outer_reg_struct, $outer_field) + riscv_offset_of!($($inner_reg_struct, $inner_field), +)
+    };
```
mtjhrc: Do we need this? `std::mem::offset_of!` supports nested fields (provided you have the correct definition of the struct).
So instead of `riscv_offset_of!(kvm_riscv_core, regs, user_regs_struct, a0);` you should just be able to do `std::mem::offset_of!(kvm_riscv_core, regs.a0)`

alicedreamed: Thanks for pointing out, removed now

alicedreamed:
> This is a great contribution, thanks @alicedreamed! How would I go about testing?

I use qemu-system-riscv64 v10.0.0 for developing, I think H ext and AIA are the hard limits here

slp:
@alicedreamed Thanks a lot for the PR! Before digging into the code aesthetics, I'm doing some functional testing.

I'm using QEMU `10.0.2` with this command line `qemu-system-riscv64 -M virt,aia=aplic-imsic -smp 8 -m 8g ...`. The emulated host is running `6.13.0-0.rc4.36.0.riscv64.fc42.riscv64`, and I'm using a booting a libkrunfw kernel with `chroot_vm` using this PR.

For me, the guest kernel is always hanging here:

```
riscv-aplic 0.aplic: 96 interrupts forwarded to MSI base 0x0000000004000000
Free page reporting enabled
```

I suspect the interrupts are not properly routed to the vCPUs, which would mean `KvmAia` needs some tweaking. I've also tried running the emulated host with `qemu-system-riscv64 -M virt,aia=aplic-imsic,aia-guests=2 -smp 8 -m 8g ...` (note the additional `aia-guests=2`) and then `KvmAia` fails when setting `KVM_DEV_RISCV_AIA_CONFIG_SRCS`.

Is it booting for you to userspace? If so, could you please describe your test environment (emulated host kernel, qemu command line, guest kernel...)?

alicedreamed:
> Is it booting for you to userspace? If so, could you please describe your test environment (emulated host kernel, qemu command line, guest kernel...)?

In my case the kernel hangs at different place I think, I'm still investigating it. Would you mind sharing how did you get the kernel boot log while running chroot_vm?

slp:
> > Is it booting for you to userspace? If so, could you please describe your test environment (emulated host kernel, qemu command line, guest kernel...)?
> 
> In my case the kernel hangs at different place I think, I'm still investigating it. Would you mind sharing how did you get the kernel boot log while running chroot_vm?

Sure, this should get you serial output: https://gist.github.com/slp/1a27f8eb736de892dd66d344530308b4

I'm fairly convinced the problem lays in the MSI/GSI interrupt routing configuration in `KvmAia`.

jakecorrenti:
@alicedreamed Just out of curiosity, did you have a specific libkrun use case on RISCV that prompted this work? I'm interested :)

alicedreamed:
> @alicedreamed Just out of curiosity, did you have a specific libkrun use case on RISCV that prompted this work? I'm interested :)

Yes, I was exploring CoVE so I decided to start from scratch with some VMM, so I chose libkrun. Now it turns out there are many works to be done in ISA, hardware and kernel level at the time being

jakecorrenti:
> > @alicedreamed Just out of curiosity, did you have a specific libkrun use case on RISCV that prompted this work? I'm interested :)
> 
> Yes, I was exploring CoVE so I decided to start from scratch with some VMM, so I chose libkrun. Now it turns out there are many works to be done in ISA, hardware and kernel level at the time being

Very cool!

src/vmm/src/linux/vstate.rs
```diff
@@ -624,13 +636,15 @@ impl Vm {
                     .map_err(Error::SetUserMemoryRegion)?;
             };
 
+            #[cfg(not(target_arch = "riscv64"))]
```
valpackett, line 639: aarch64 busy-loops with guest memfd currently, could that be what you've been seeing on riscv64 as well?

PR #371, opened 2025-07-07
README: add libkrun logo

slp:


mtjhrc (approved):
Looks nice! 

Btw do you a like a source vector image for this? If so it would be nice to also upload an .svg to the repository somewhere.

PR #380, opened 2025-07-25
init.c: only use KRUN_EXIT_CODE_IOCTL if root is really virtiofs

nohajc:
I was playing with different VM configurations and was able to run a system from a raw disk image instead of virtiofs.
It basically works, I just need a custom cmdline and an actual copy of init.krun on the image filesystem.

The only issue is that init assumes root is always virtiofs and uses the KRUN_EXIT_CODE_IOCTL which will fail.
To avoid the error on conventional filesystems, we can first call statfs and check `f_type`.

### DIFF
--- a/init/init.c
+++ b/init/init.c
@@ -16,6 +16,7 @@
 #include <sys/resource.h>
 #include <sys/socket.h>
 #include <sys/stat.h>
+#include <sys/statfs.h>
 #include <sys/time.h>
 #include <sys/types.h>
 #include <sys/wait.h>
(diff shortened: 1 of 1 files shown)


ggoodman:
@nohajc not related to reviewing this PR, but if you have any examples of your setup to boot from an image it would be a valuable resource to share with the community 😄.

nohajc:
@ggoodman Sure, my setup is very crude at this point. Just a couple of shell scripts to prepare the root filesystem and a lua script to interact with the C API. I also have a project built on top of libkrun but that's in Rust. Now I'm basically trying to make the microVMs scriptable but without using krunvm. I was definitely planning to share my results.

nohajc:
@ggoodman I've seen your `krun_set_rootfs_kernel_args` pull request. The reason it works for me is that I already use a custom kernel image (`krun_set_kernel`) which lets me specify the entire kernel command line too.

Could be something like this
```
reboot=k panic=-1 panic_print=0 nomodule console=hvc0 root=/dev/vda rw quiet no-kvmapf init=/init.krun KRUN_INIT=/bin/bash KRUN_WORKDIR=/
```

nohajc:
@ggoodman You can check out my setup here:
https://github.com/nohajc/anylinuxfs/tree/main/kernel-builder

slp (approved):
LGTM, thanks!

PR #381, opened 2025-07-25
Add support for EDID display information

mtjhrc:
Adds support for EDID (Extended Display Identification Data) for our `virtio-gpu` device and extend the public API in order to specify the more detailed display information.

 Clippy warnings are fixed in https://github.com/containers/libkrun/pull/386.


slp:
LGTM, thanks!

PR #346, opened 2025-06-05
Add krun_set_rootfs_kernel_args API

ggoodman:
This commit introduces a new API function `krun_set_rootfs_kernel_args` that allows
callers to override kernel arguments related to the root filesystem, replacing the
default "rootfstype=virtiofs rw" parameters in the kernel command line.

New features:
- Added RootfsConfig struct with builder pattern for type-safe configuration
- Supports customizing root device, filesystem type, mount flags, and read-only mode
- Falls back to sensible defaults: /dev/vda1, virtiofs, rw when parameters are NULL
- Integrates with existing kernel command line construction in get_effective_kernel_cmdline()
- Follows established patterns used by other config types like BlockDeviceConfig

API signature:
```c
int32_t krun_set_rootfs_kernel_args(uint32_t ctx_id,
                                    const char *device,
                                    const char *fs_type,
                                    const char *mount_flags,
                                    bool read_only);
```

PR #384, opened 2025-08-05
Simplify booting from block devices

nohajc:
I had an idea how to simplify booting from block devices. Since `init.krun` is a virtual file implemented in the virtiofs `passthrough` module, we'd normally have to manually copy it to the device from which we want to boot and also adjust kernel cmdline to override the default root settings.

Instead, if `krun_set_root()` was never called but there is a block device, we can still create a dummy rootfs just for executing init and then switch to the real root with a similar mechanism initramfs would use.

This relies on setting `KRUN_BLOCK_ROOT` environment variable (internally) as well as reading `/proc/filesystems` so we can implement the behavior of `mount -t auto` for the real root.

I also added another custom ioctl which I use for cleaning up the temporary root when no longer needed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -947,6 +947,7 @@ dependencies = [
  "nitro-enclaves",
  "once_cell",
  "polly",
+ "rand 0.9.2",
  "utils",
  "vm-memory",
  "vmm",
@@ -1103,7 +1104,7 @@ dependencies = [
  "bitflags 2.9.1",
  "libc",
  "nix 0.26.4",
- "rand 0.9.1",
+ "rand 0.9.2",
  "vsock",
 ]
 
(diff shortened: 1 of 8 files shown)


nohajc:
Hmm, just realized `/dev/vda` might not be the best default and we'll need a way to configure it anyway. Also @ggoodman is working on something similar (#346).

mtjhrc:
Really cool idea!  I'm not sure if we want this enabled implicitly though.   Perhaps you could add another API function, something like `krun_set_root_disk_remount(ctx_id, "/dev/vda")` that enables this functionality and also allows for configuring the disk?

nohajc:
Sounds good. Perhaps the function could also take the filesystem type (with the `auto` type accepted too) and other mount options?

mtjhrc:
> Sounds good. Perhaps the function could also take the filesystem type (with the `auto` type accepted too) and other mount options?

Sure, sounds good.

nohajc:
I'm done with the implementation. 🙂

I used `krun_set_root_disk_remount` for the function name even though it's kind of similar to the deprecated `krun_set_root_disk`. The other thing is `remount` basically reveals an implementation detail in the name which I'm not sure is appropriate.

If anybody has a different idea how to call it, I'll be happy to change it.

slp:
LGTM, thanks! Could you please rebase it on top of latest `main`?

nohajc:
> LGTM, thanks! Could you please rebase it on top of latest `main`?

Done.

PR #386, opened 2025-08-12
rustc: Fix mismatched_lifetime_syntaxes

RuoqingHe:
Rust v1.89.0 is reporting rustc::mismatched_lifetime_syntaxes:

```console
warning: hiding a lifetime that's elided elsewhere is confusing
   --> src/devices/src/virtio/descriptor_utils.rs:499:13
    |
499 |     memory: &GuestMemoryMmap,
    |             ^^^^^^^^^^^^^^^^ the lifetime is elided here
...
504 | ) -> Result<DescriptorChain> {
    |             --------------- the same lifetime is hidden here
    |
    = help: the same lifetime is referred to in inconsistent ways, making the signature confusing
    = note: `#[warn(mismatched_lifetime_syntaxes)]` on by default
help: use `'_` for type paths
    |
504 | ) -> Result<DescriptorChain<'_>> {
    |                            ++++
```

fix as suggested by rustc to kill those warnings.

### DIFF
--- a/src/devices/src/virtio/rng/device.rs
+++ b/src/devices/src/virtio/rng/device.rs
@@ -14,10 +14,6 @@ pub(crate) const REQ_INDEX: usize = 0;
 // Supported features.
 pub(crate) const AVAIL_FEATURES: u64 = 1 << uapi::VIRTIO_F_VERSION_1 as u64;
 
-#[derive(Copy, Clone, Debug, Default)]
-#[repr(C, packed)]
-pub struct VirtioRng {}
-
 pub struct Rng {
     pub(crate) queues: Vec<VirtQueue>,
     pub(crate) queue_events: Vec<EventFd>,
--- a/src/rutabaga_gfx/src/rutabaga_os/descriptor.rs
+++ b/src/rutabaga_gfx/src/rutabaga_os/descriptor.rs
@@ -138,6 +138,7 @@ impl From<File> for SafeDescriptor {
 ///
 /// Note that with the exception of the last use-case (which requires proper error checking against
 /// the descriptor being closed), the `Descriptor` instance would be very short-lived.
+#[allow(unused)]
 #[derive(Copy, Clone, Hash, PartialEq, Eq, PartialOrd, Ord)]
 #[repr(transparent)]
 pub struct Descriptor(pub RawDescriptor);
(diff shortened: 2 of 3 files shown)


tylerfanelli (changes requested):
Sorry, didn't wait to see that there's some failing tests.

RuoqingHe:
> Sorry, failed to realize that there's some failing tests.

I saw those unused in 1.89.0 but not in 1.87.0, should we consider pinning our toolchain version 🙂

Fixing them now

src/devices/src/virtio/rng/device.rs
```diff
@@ -14,6 +14,7 @@ pub(crate) const REQ_INDEX: usize = 0;
 // Supported features.
 pub(crate) const AVAIL_FEATURES: u64 = 1 << uapi::VIRTIO_F_VERSION_1 as u64;
 
+#[allow(unused)]
```
tylerfanelli: Rather than keeping unused code around, perhaps its worth considering removing this altogether? @slp WDYT?

mtjhrc: I would definitely remove this `VirtioRng` struct here, it seems useless.

src/rutabaga_gfx/src/rutabaga_os/descriptor.rs
```diff
@@ -138,6 +138,7 @@ impl From<File> for SafeDescriptor {
 ///
 /// Note that with the exception of the last use-case (which requires proper error checking against
 /// the descriptor being closed), the `Descriptor` instance would be very short-lived.
+#[allow(unused)]
 #[derive(Copy, Clone, Hash, PartialEq, Eq, PartialOrd, Ord)]
 #[repr(transparent)]
 pub struct Descriptor(pub RawDescriptor);
```
mtjhrc, line 144: This other unused warning is in Rutabaga and there I'm not sure what to do here. But either suppressing it or removing the `impl` is fine. 
I think you can leave it as is. 

Maybe one day our downstream changes to Rutabaga could be upstreamed. 

Btw. looks like Rutabaga might move under [freedesktop](https://gitlab.freedesktop.org/freedesktop/freedesktop/-/issues/2459):
> The first project that will be housed there would be [rutabaga_gfx](https://crosvm.dev/book/appendix/rutabaga_gfx.html).  It’s already used in crosvm, [QEMU](https://www.qemu.org/docs/master/system/devices/virtio-gpu.html#virtio-gpu-rutabaga), [rust-vmm](https://github.com/rust-vmm/vhost-device/pull/668).  We are taking steps to delete in crosvm and make it upstream in Freedesktop, as per [user-demand](https://github.com/rust-vmm/vhost-device/issues/841#issuecomment-2949051329).  There is also a protocol generator that we can collaborate on.  So the initial project list would look like:

mtjhrc (changes requested):
LGTM, thanks! Just remove the unused structs too (except in Rutabaga)

mtjhrc (approved):

Issue #392, opened 2025-08-15
krun_set_gvproxy_path broken


nirs:
I tested current main with krunkit main and the unixSocketPath option does not work now with vment-helper.
- vmnet-helper example does not get the ip address
- in vment-helper log we see no traffic from the vm to the host, only from the host to the vm

### Build libkrun from main

```
git clone https://github.com/containers/libkrun.git
make EFI=1 NET=1 BLK=1 GPU=1
sudo make EFI=1 NET=1 BLK=1 GPU=1 install
```

### Build krunkit main linked with local libkrun

```
git clone https://github.com/containers/krunkit.git
cd krunkit
```

Patch krunkit to use /usr/local/lib:

```diff
diff --git a/build.rs b/build.rs
index 5ec55cd..5f9eb31 100644
--- a/build.rs
+++ b/build.rs
@@ -2,5 +2,5 @@
 
 fn main() {
     #[cfg(target_os = "macos")]
-    println!("cargo:rustc-link-search=/opt/homebrew/lib");
+    println!("cargo:rustc-link-search=/usr/local/lib");
 }
```

Build and install

```
make
sudo make install
```

Check that krunkit is linked correctly:

```
% otool -L /usr/local/bin/krunkit 
/usr/local/bin/krunkit:
	/usr/local/lib/libkrun-efi.1.dylib (compatibility version 1.0.0, current version 1.14.0)
	/System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation (compatibility version 150.0.0, current version 3502.1.255)
	/usr/lib/libobjc.A.dylib (compatibility version 1.0.0, current version 228.0.0)
	/System/Library/Frameworks/IOKit.framework/Versions/A/IOKit (compatibility version 1.0.0, current version 275.0.0)
	/usr/lib/libiconv.2.dylib (compatibility version 7.0.0, current version 7.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1351.0.0)
```

### Check out vment-helper

```console
git clone https://github.com/nirs/vmnet-helper.git
cd vmnet-helper
```

### Install vment-helper

```console
meson setup build
meson compile -C build
meson install -C build
sudo install -m 0640 sudoers.d/vmnet-helper /etc/sudoers.d/
```

### Create virtual environment with pyyaml

```console
python3 -m venv .venv                                                                                                                                                                        
source .venv/bin/activate
pip install pyyaml
```

[...]

PR #395, opened 2025-08-26
libkrun: on legacy interface delay net creation

slp:
The legacy network configuration interface allowed to configure the MAC address before or after the configuration of the network interface itself. This got accidentally broken with the introduction of the new network configuration API.

Let's fix the API for legacy users by delaying network interface creation until "krun_start_enter" for the legacy API.

Fixes: #392

### DIFF
--- a/src/libkrun/src/lib.rs
+++ b/src/libkrun/src/lib.rs
@@ -121,6 +121,13 @@ impl KrunfwBindings {
     }
 }
 
+#[derive(Clone)]
+#[cfg(feature = "net")]
+enum LegacyNetworkConfig {
+    VirtioNetPasst(RawFd),
+    VirtioNetGvproxy(PathBuf),
+}
+
 #[derive(Default)]
 struct ContextConfig {
     #[cfg(not(feature = "efi"))]
@@ -131,9 +138,12 @@ struct ContextConfig {
     env: Option<String>,
     args: Option<String>,
     rlimits: Option<String>,
+    #[cfg(feature = "net")]
+    legacy_net_cfg: Option<LegacyNetworkConfig>,
+    #[cfg(feature = "net")]
+    legacy_mac: Option<[u8; 6]>,
     net_index: u8,
     tsi_port_map: Option<HashMap<u16, u16>>,
-    mac: Option<[u8; 6]>,
     #[cfg(feature = "blk")]
     block_cfgs: Vec<BlockDeviceConfig>,
     #[cfg(feature = "blk")]
@@ -274,8 +284,9 @@ impl ContextConfig {
         }
     }
 
+    #[cfg(feature = "net")]
     fn set_net_mac(&mut self, mac: [u8; 6]) {
-        self.mac = Some(mac);
+        self.legacy_mac = Some(mac);
     }
 
     fn set_port_map(&mut self, new_port_map: HashMap<u16, u16>) -> Result<(), ()> {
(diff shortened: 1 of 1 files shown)


slp:
@nirs PTAL

src/libkrun/src/lib.rs
```diff
@@ -129,6 +136,8 @@ struct ContextConfig {
     env: Option<String>,
     args: Option<String>,
     rlimits: Option<String>,
+    #[cfg(feature = "net")]
+    legacy_net_cfg: Option<LegacyNetworkConfig>,
     net_index: u8,
     tsi_port_map: Option<HashMap<u16, u16>>,
     mac: Option<[u8; 6]>,
```
nirs: Isn't mac also legacy? we can rename it to legacy_mac and put group legacy variable together. This will make the code more clear, and easier to remove in the future.

src/libkrun/src/lib.rs
```diff
@@ -1911,6 +1908,35 @@ pub extern "C" fn krun_start_enter(ctx_id: u32) -> i32 {
     };
 
     let mut ctx_cfg = match CTX_MAP.lock().unwrap().remove(&ctx_id) {
+        #[cfg(feature = "net")]
+        Some(mut ctx_cfg) => {
+            {
+                let legacy_net_cfg = ctx_cfg.legacy_net_cfg.clone();
+                match legacy_net_cfg {
+                    Some(LegacyNetworkConfig::VirtioNetGvproxy(path)) => {
+                        let mac = ctx_cfg.mac.unwrap_or([0x5a, 0x94, 0xef, 0xe4, 0x0c, 0xee]);
```
nirs: This can be done once for both cases before the match.

src/libkrun/src/lib.rs
```diff
@@ -1911,6 +1908,35 @@ pub extern "C" fn krun_start_enter(ctx_id: u32) -> i32 {
...
+                        create_virtio_net(
+                            &mut ctx_cfg,
+                            VirtioNetBackend::UnixgramPath(path, true),
+                            mac,
+                            NET_COMPAT_FEATURES,
+                        );
```
nirs: Since this call is the same for both case, we can simplify to:

```rust
let backend = match legacy_cfg {
    Some(LegacyNetworkConfig::VirtioNetGvproxy(path)) =>
        VirtioNetBackend::UnixgramPath(path, true)
    Some(LegacyNetworkConfig::VirtioNetPasst(fd)) =>
        VirtioNetBackend::UnixstreamFd(fd)
}

let mac = ctx_cfg.mac.unwrap_or([0x5a, 0x94, 0xef, 0xe4, 0x0c, 0xee]);
create_virtio_net(&mut ctx_cfg, backend, mac, NET_COMPAT_FEATURES,);
```

(I ignored the None case for illustration)

src/libkrun/src/lib.rs
slp: Done.

src/libkrun/src/lib.rs
slp: I did that and refactored the code a bit too, thanks!

src/libkrun/src/lib.rs
```diff
@@ -2136,6 +2132,22 @@ pub extern "C" fn krun_start_enter(ctx_id: u32) -> i32 {
         return -libc::EINVAL;
     }
 
+    #[cfg(feature = "net")]
+    {
+        if let Some(legacy_net_cfg) = ctx_cfg.legacy_net_cfg.clone() {
```
nirs, line 2137: Maybe

```rust
if ctx_cfg.legacy_net_cfg.is_some() {
```

nirs (approved):

src/libkrun/src/lib.rs
mtjhrc: > Maybe
> 
> ```rust
> if ctx_cfg.legacy_net_cfg.is_some() {
> ```

That wouldn't really work, you would still have to `.unwrap()` or match on `legacy_net_cfg`.

PR #396, opened 2025-08-27
Debug integration tests

slp:


jakecorrenti (dismissed):
LGTM, thanks!

tests/test_cases/src/test_tsi_tcp_guest_connect.rs
```diff
@@ -30,6 +30,7 @@ mod host {
             let listener = self.tcp_tester.create_server_socket();
             thread::spawn(move || self.tcp_tester.run_server(listener));
             unsafe {
+                krun_call_u32!(krun_set_log_level(9))?;
```
mtjhrc: This should technically use  `krun_call!` macro (but that is a minor detail and doesn't matter much).

My plan was to save the log files in the tmp directory of each test upon failure (that's also why I added the API to redirect the log output to a file, and the tmp directories are kept on failure). I'm not sure how difficult it is to access files on github from the runner though... 

I don't really mind writing the log to stderr though, it should just not be the default IMO at such a high verbosity. Here I would specify the log level to just maybe `warn` or something. And then specify the RUST_LOG=trace specifically in the github ci configuration.

mtjhrc (changes requested):

tests/test_cases/src/test_tsi_tcp_guest_connect.rs
slp: Makes sense, I've just updated the PR. Thanks!

mtjhrc (commented):
LGTM, thanks!

slp:
> LGTM, thanks!

@mtjhrc Please approve the PR since it's still in "changes requested" state.

Issue #382, opened 2025-07-31
"init" doesn't translate unicode characters


slp:
See https://github.com/containers/crun/issues/1836

jakecorrenti:
I can tackle this

slp:
@jakecorrenti I'd like to have this one fixed on the next release. Have you started working on this? Would you mind if I give it a try?

jakecorrenti:
> [@jakecorrenti](https://github.com/jakecorrenti) I'd like to have this one fixed on the next release. Have you started working on this? Would you mind if I give it a try?

I started working on it, but I was hitting an issue where your initrd-krunfw script was failing [here](https://github.com/slp/initrd-krunfw/blob/a407d9ebeb6e353533d653f3dbcd0e385fc9090f/build.sh#L16) because the `gen_init_cpio` isn't getting created during kernel compilation.

I couldn't figure it out and put it aside until you got back so I could ask you. You're welcome to take it over if you'd like.

jakecorrenti:
I think this is also related: https://github.com/containers/libkrun/issues/358

slp:
> > [@jakecorrenti](https://github.com/jakecorrenti) I'd like to have this one fixed on the next release. Have you started working on this? Would you mind if I give it a try?
> 
> I started working on it, but I was hitting an issue where your initrd-krunfw script was failing [here](https://github.com/slp/initrd-krunfw/blob/a407d9ebeb6e353533d653f3dbcd0e385fc9090f/build.sh#L16) because the `gen_init_cpio` isn't getting created during kernel compilation.

Are you building the `SEV` or `TDX` variants of libkrunfw? `gen_init_cpio` is not generated when building the generic version of `libkrunfw`.

jakecorrenti:
> > > [@jakecorrenti](https://github.com/jakecorrenti) I'd like to have this one fixed on the next release. Have you started working on this? Would you mind if I give it a try?
> > 
> > 
> > I started working on it, but I was hitting an issue where your initrd-krunfw script was failing [here](https://github.com/slp/initrd-krunfw/blob/a407d9ebeb6e353533d653f3dbcd0e385fc9090f/build.sh#L16) because the `gen_init_cpio` isn't getting created during kernel compilation.
> 
> Are you building the `SEV` or `TDX` variants of libkrunfw? `gen_init_cpio` is not generated when building the generic version of `libkrunfw`.

No I was building the generic version, which explains the error. Should I have been doing something different to update the initrd bundle instead of using that script?

PR #397, opened 2025-08-28
init: translate unicode strings into ascii

slp:
So far, we were tolerating but ignoring unicode strings in the JSON files we read. That's a problem since OCI tools will encode some ascii chars in escaped unicode.

Since ascii characters can be directly translated from unicode, let's do that for valid ascii characters (0 to 127).

Fixes: #382


slp:
@jakecorrenti PTAL

Issue #373, opened 2025-07-07
Question about VMM process lifecycle


nohajc:
It is documented that calling `krun_start_enter` effectively transfers control over the process to libkrun which calls exit when the virtual machine is stopped (and therefore `krun_start_enter` does not return).

My question is twofold.

1. Why do you call `_exit` instead of the more "friendly" `exit` which among other things runs callbacks registered by `atexit`?
2. Would it be possible to modify libkrun so that it can be used in a context where process termination cannot be controlled by the hypervisor? I'm thinking macOS (user-space) system extension for example.

slp:
> 1. Why do you call `_exit` instead of the more "friendly" `exit` which among other things runs callbacks registered by `atexit`?

TBH, I don't know. That code is part of the little remains from Firecracker that are still in libkrun codebase. Can't say why they choose to use `_exit` there, instead of `std::process::exit` or any other option. I would be in favor of changing it if there is a benefit without side effects.

> 2. Would it be possible to modify libkrun so that it can be used in a context where process termination cannot be controlled by the hypervisor? I'm thinking macOS (user-space) system extension for example.

The possibility of having a different entry point other than `krun_start_enter` that would not exit the process and, instead return to the caller, has been discussed before. So far the consensus was to wait for the API revamp in v2, but if there's a use case and a PR implement such entry point as a new function alternative to `krun_start_enter`, I'd be in favor of merging it.

nohajc:
@slp So there aren't any fundamental reasons why this would be problematic? I mean in a similar sense such as when you have POSIX `exec` which basically replaces your current process memory so there's nothing to return to?

I know executing a virtual machine doesn't exactly do this but I'm no expert either. Basically, the question is whether you can easily restore CPU/thread/memory (or whatever) state when returning from the VM entry point after the hypervisor terminated.

mtjhrc:
> Why do you call _exit instead of the more "friendly" exit

I'm not sure either, maybe we could, but there probably could be some issues. `_exit` seems safer to me for now.

> Would it be possible to modify libkrun so that it can be used in a context where process termination cannot be controlled by the hypervisor

I think it should be possible, and I think it would be nice feature. The main problem would actually be the Rust code, where we probably don't have proper Drop implementations in places, don't handle a lot of objects lifetimes (e.g. using RawFd) and we don't have mechanisms for stopping worker threads cleanly... Basically the clean shutdown paths are untested or are not implemented at all, hence we just `_exit` the process currently.

[...]

PR #401, opened 2025-08-30
include: libkrun: add `krun_set_pidfile` API

jakecorrenti:
Add a new `krun_set_pidfile` API to allow the user to specify a path in which to write the current process's ID.

This was inspired by a new option I'm working on for krunkit: https://github.com/containers/krunkit/pull/71. I figured it could be useful for users here as well.

### DIFF
--- a/examples/chroot_vm.c
+++ b/examples/chroot_vm.c
@@ -325,6 +325,12 @@ int main(int argc, char *const argv[])
         return -1;
     }
 
+    if (err = krun_set_pidfile(ctx_id, "/tmp/testing.pid")) {
+        errno = -err;
+        perror("Error setting pidfile");
+        return -1;
+    }
+
     // Start and enter the microVM. Unless there is some error while creating the microVM
     // this function never returns.
     if (err = krun_start_enter(ctx_id)) {
--- a/include/libkrun.h
+++ b/include/libkrun.h
@@ -901,6 +901,18 @@ int32_t krun_nitro_set_start_flags(uint32_t ctx_id, uint64_t start_flags);
 int32_t krun_set_root_disk_remount(uint32_t ctx_id, const char *device, const char *fstype, const char *options);
 
 /**
+ * Configure the pidfile path.
+ *
+ * Arguments:
+ *  "ctx_id" - the configuration context ID.
+ *  "path" - path to the pidfile.
+ *
+ * Returns:
+ *  Zero on success or a negative error number on failure.
+ */
+int32_t krun_set_pidfile(uint32_t ctx_id, const char *path);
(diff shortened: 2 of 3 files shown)


mtjhrc:
I am not really a fan of this. Ideally this seems like something the user of this library should be responsible for. But I see your problem with that doing that - we don't have a clear point in the API where we finish building the VM and when when we are starting it. This could be solved more generally by having a function like  `krun_finalize()` before `krun_start_enter`, but maybe for API v2. Another generic option could be a callback (not that nice IMO)

Also another perspective: `krun_start_enter` failing is still an arbitrary point - the guest kernel could also fail to start (due to misconfigured kernel arguments perhaphs?, or just some bug in our devices), our init may fail, and also the application in the guest might fail, causing the VM to exit...

nirs:
I agree that this is better done in the application using libkrun, which is also a better place to remove the pidfile on exit.

The reason to create the pidfile as the last thing before we start the guest is to help the application detect successful start by waiting until the pidfile is created. But once the application detected that the vm was started, it needs to wait until the vm is accessible (e.g. via ssh). During this wait it can detect if libkrun failed after the pidfile was created. So I don't think we gain much by moving the functionality to libkrun.

Cleaning up in libkrun is harder since it should not change process state like ataxia handler. The application using libkrun is a better place to manage process lifecycle.

mtjhrc:
For cleanup: https://github.com/containers/libkrun/issues/373 is related. Ideally we would have a version of `krun_start_enter` that instead of exiting the process returns whenever the VM is done running...

nirs:
I did not that krun_start_enter does not return and call _exit. In this case the user cannot cleanup using atexit handler and this is a good reason to move the funcionallity into libkrun. But fixing clean to call exit() and let the user setup atexit handlers seems like a better approach.

jakecorrenti:
Closing based on feedback. Thanks all

jakecorrenti:
> I am not really a fan of this. Ideally this seems like something the user of this library should be responsible for. But I see your problem with that doing that - we don't have a clear point in the API where we finish building the VM and when when we are starting it. 

Yes, this one one of the reasons I thought it _could_ be helpful here.

>This could be solved more generally by having a function like `krun_finalize()` before `krun_start_enter`, but maybe for >API v2. Another generic option could be a callback (not that nice IMO)
> 
> Also another perspective: `krun_start_enter` failing is still an arbitrary point - the guest kernel could also fail to start (due to misconfigured kernel arguments perhaphs?, or just some bug in our devices), our init may fail, and also the application in the guest might fail, causing the VM to exit...

Also a fair point

PR #406, opened 2025-09-04
Round of dependency updates and consolidations

slp:
Let's update and consolidate some critical dependencies across the workspace.


src/utils/src/macos/eventfd.rs
```diff
@@ -6,115 +6,72 @@
...
-                // We may get EAGAIN if the eventfd is overstimulated, but we can safely
-                // ignore it as we can be sure the subscriber will get notified.
-                io::ErrorKind::WouldBlock => Ok(()),
-                _ => Err(error),
-            }
-        } else {
-            Ok(())
+        match write(&self.write_fd, &v.to_le_bytes()) {
+            Ok(_) => Ok(()),
+            Err(err) => Err(err.into()),
```
mtjhrc: Hmm, this changes the semantics a bit. But the new behavior should be better. 
`write` to a Linux `eventfd` can indeed cause `EAGAIN` under similar conditions (the counter is at its maximum value). Just pointing it out, because it's not mentioned in the commit message.

src/utils/src/macos/eventfd.rs
```diff
@@ -6,115 +6,72 @@
...
-        return Err(io::Error::last_os_error());
-    }
-
-    let ret = unsafe { fcntl(fd, F_SETFL, flags | O_NONBLOCK) };
-    if ret < 0 {
-        return Err(io::Error::last_os_error());
-    }
+fn set_nonblock(fd: &OwnedFd) -> result::Result<(), io::Error> {
+    let flags = fcntl(fd, FcntlArg::F_GETFL)?;
+    let flags = OFlag::from_bits(flags).ok_or(io::ErrorKind::InvalidData)?;
```
mtjhrc, line 24: You can just do `OFlag::from_bits_retain` to keep any unknown flags which are unknown to nix on the on the file descriptor.... which seems like the more correct semantics of this function.

src/utils/src/macos/eventfd.rs
slp: Good catch, thanks!

This was completely unintended. I've restored the previous semantics.

src/utils/src/macos/eventfd.rs
```diff
@@ -6,115 +6,80 @@
...
+        match write(&self.write_fd, &v.to_le_bytes()) {
+            Ok(_) => Ok(()),
+            Err(err) => {
+                match err {
+                    // We may get EAGAIN if the eventfd is overstimulated, but we can safely
+                    // ignore it as we can be sure the subscriber will get notified.
+                    Errno::EAGAIN => Ok(()),
+                    _ => Err(err.into()),
+                }
             }
```
mtjhrc: Bit of a nitpick, but you can just write it without nesting the match:
```suggestion
        match write(&self.write_fd, &v.to_le_bytes()) {
            // We may get EAGAIN if the eventfd is overstimulated, but we can safely
            // ignore it as we can be sure the subscriber will get notified.
            Ok(_) | Err(Errno::EAGAIN) => Ok(()),
            Err(err) => Err(err.into()),
```

mtjhrc: or

```suggestion
        match write(&self.write_fd, &v.to_le_bytes()) {
            Ok(_) => Ok(()),
            // We may get EAGAIN if the eventfd is overstimulated, but we can safely
            // ignore it as we can be sure the subscriber will get notified.
            Err(Errno::EAGAIN) => Ok(()),
            Err(err) => Err(err.into()),
```

slp: I like the first one, applied, thanks!

PR #407, opened 2025-09-06
rutabaga_gfx: fixes for wayland/sommelier

valpackett:
I've been trying to get the Wayland forwarding to work as well as on crosvm.. With these fixes both sommelier and wayland-proxy-virtwl do actually work in muvm.

See the commit messages for details.

(is there any custom krun-specific client relying on `CROSS_DOMAIN_MAX_IDENTIFIERS: usize = 4;` ? :thinking:)

### DIFF
--- a/src/rutabaga_gfx/src/cross_domain/mod.rs
+++ b/src/rutabaga_gfx/src/cross_domain/mod.rs
@@ -1125,6 +1125,13 @@ impl RutabagaContext for CrossDomainContext {
         // Items that are removed from the table after one usage.
         match item {
             CrossDomainItem::ShmBlob(descriptor) => {
+                #[allow(unused_mut)]
+                let mut access = RUTABAGA_MAP_ACCESS_RW;
+                #[cfg(target_os = "linux")]
+                if fcntl(&descriptor, FcntlArg::F_GET_SEALS)? & libc::F_SEAL_WRITE != 0 {
+                    access &= !RUTABAGA_MAP_ACCESS_WRITE;
+                };
+
                 let hnd = RutabagaHandle {
                     os_handle: descriptor,
                     handle_type: RUTABAGA_MEM_HANDLE_TYPE_SHM,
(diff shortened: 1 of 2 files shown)


src/rutabaga_gfx/src/cross_domain/mod.rs
```diff
@@ -1122,6 +1122,11 @@ impl RutabagaContext for CrossDomainContext {
         // Items that are removed from the table after one usage.
         match item {
             CrossDomainItem::ShmBlob(descriptor) => {
+                let mut access = RUTABAGA_MAP_ACCESS_RW;
+                if fcntl(descriptor.as_raw_fd(), FcntlArg::F_GET_SEALS)? & libc::F_SEAL_WRITE != 0 {
```
slp: This fails on macOS because it lacks everything related to SEALS. Please put it behind a build-time target_os conditional.

TBH, the right solution would probably putting the whole cross-domain functionality behind a build-time target_os conditional, but that'll require a significant amount of changes, and we want to eventually (at some point in 2026?) converge with upstream rutabaga_gfx, now that it's an independent project separated from crosvm.

slp:
Code looks good, apart from the minor issue with the build-time conditional on macOS. Tested on Asahi with your fork of [wayland-proxy-virtwl](https://github.com/valpackett/wayland-proxy-virtwl) and the wayland applications I've tested work perfectly fine. Thanks @valpackett !

src/rutabaga_gfx/src/cross_domain/mod.rs
valpackett: Sure.

> we want to eventually (at some point in 2026?) converge with upstream rutabaga_gfx

heh, just wanted to ask about that! nice.

valpackett: Added `cfg()`. And rebased.

slp (approved):
LGTM, thanks!

PR #408, opened 2025-09-08
Implement krun_get_max_vcpus function for both macOS and Linux

ericcurtin:
So we can get the max amount of vcpus available.


src/libkrun/src/lib.rs
```diff
@@ -1638,6 +1638,37 @@ pub unsafe extern "C" fn krun_check_nested_virt() -> i32 {
     -libc::EOPNOTSUPP
 }
 
+/// Gets the maximum number of vCPUs supported by the hypervisor.
+///
+/// Returns the maximum number of vCPUs that can be created by this hypervisor,
+/// or a negative error code on failure.
+#[no_mangle]
+pub extern "C" fn krun_get_max_vcpus() -> i32 {
```
jakecorrenti, line 1647: I'm not sure how necessary this is on lines 1668-1669 considering we only support those two OSes:
```rust
    #[cfg(not(any(target_os = "macos", target_os = "linux")))]
    -libc::EOPNOTSUPP
```

I think you could probably get away with the following instead if we did, in fact, want to keep the sanity check. It should prevent someone's code from even compiling if they use the function:

```suggestion
#[cfg(any(target_os = "macos", target_os = "linux"))]
#[no_mangle]
pub extern "C" fn krun_get_max_vcpus() -> i32 {
```

src/libkrun/src/lib.rs
```diff
@@ -1638,6 +1638,37 @@ pub unsafe extern "C" fn krun_check_nested_virt() -> i32 {
...
+    #[cfg(target_os = "macos")]
+    {
+        use hvf::bindings::{hv_vm_get_max_vcpu_count, HV_SUCCESS};
+        let mut max_vcpu_count: u32 = 0;
+        let ret = unsafe { hv_vm_get_max_vcpu_count(&mut max_vcpu_count as *mut u32) };
+        if ret == HV_SUCCESS {
+            max_vcpu_count as i32
+        } else {
+            -libc::EINVAL
+        }
```
jakecorrenti: I think returning EINVAL here is fine, but it's not very descriptive of what happened. What if we did something like this as well right before:
```rust
 error!("Error retrieving max vcpu count: {ret:?}");
```

src/libkrun/src/lib.rs
```diff
@@ -1638,6 +1638,37 @@ pub unsafe extern "C" fn krun_check_nested_virt() -> i32 {
...
+    }
+
+    #[cfg(target_os = "linux")]
+    {
+        use kvm_ioctls::Kvm;
+        match Kvm::new() {
+            Ok(kvm) => kvm.get_max_vcpus() as i32,
+            Err(_) => -libc::EINVAL,
```
jakecorrenti: Same comment as above. I think the user should at least be made aware of the return code returned by KVM if we aren't going to return it to them directly.

src/libkrun/src/lib.rs
ericcurtin: Done

src/libkrun/src/lib.rs
ericcurtin: Done

src/libkrun/src/lib.rs
ericcurtin: Done

src/libkrun/src/lib.rs
jakecorrenti: Did you re-push? I'm not seeing anything on my end

ericcurtin:
Sorry @jakecorrenti now it's pushed

ericcurtin:
@jakecorrenti @slp ready for merge?

mtjhrc (approved):
LGTM

PR #410, opened 2025-09-12
Implement support for TSIv3

slp:
To support sockets beyond AF_INET, TSIv3 generalizes socketaddresses behind an abstract sockaddr type, just as the kernel does internally.

We need to follow suit here, and luckily we can rely on nix's SockaddrStorage to do most of the abstraction work for us.

Then, building on top of that, we implement support for AF_INET6 and AF_UNIX.

This is an ABI breaking change, so we also require the libkrunfw counterpart https://github.com/containers/libkrunfw/pull/101

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,6 +83,24 @@ A conventional virtual interface that allows the guest to communicate with the o
 
 Use `krun_add_net_unixstream` and/or `krun_add_net_unixdgram` to add a virtio-net interface connected to the userspace network proxy.
 
+## Security model
+
+The libkrun security model is primarily defined by the consideration that both the guest and the VMM pertain to the same security context. For many operations, the VMM acts as a proxy for the guest within the host. Host resources that are accessible to the VMM can potentially be accessed by the guest through it.
+
+While defining the security implementation of your environment, you should think about the guest and the VMM as a single entity. To prevent the guest from accessing host's resources, you need to use the host's OS security features to run the VMM inside an isolated context. On Linux, the primary mechanism to be used for this purpose is namespaces. Single-user systems may have a more relaxed security policy and just ensure the VMM runs with a particular UID/GID.
+
+While most virtio devices allow the guest to access resources from the host, two of them require special consideration when used: virtio-fs and virtio-vsock+TSI.
+
+### virtio-fs
+
+When exposing a directory in a filesystem from the host to the guest through virtio-fs devices configured with `krun_set_root` and/or `krun_add_virtiofs`, libkrun **does not** provide any protection against the guest attempting to access other directories in the same filesystem, or even other filesystems in the host.
+
+A mount point isolation mechanism from the host should be used in combination with virtio-fs.
+
+### virtio-vsock + TSI
+
+When TSI is enabled, the VMM acts as a proxy for AF_INET, AF_INET6 and AF_UNIX sockets, for both incoming and outgoing connections. For all that matters, the VMM and the guest should be considered to be running in the network context. As such, you should apply on the VMM whatever restrictions you want to apply on the guest.
+
 ## Building and installing
 
 ### Linux (generic variant)


src/devices/src/virtio/vsock/tsi_stream.rs
```diff
@@ -71,6 +76,8 @@ impl TcpProxy {
     ) -> Result<Self, ProxyError> {
         let family = match family as i32 {
             AF_INET => AddressFamily::Inet,
+            AF_INET6 => AddressFamily::Inet6,
+            AF_UNIX => AddressFamily::Unix,
```
mtjhrc: Now that this supports `AF_UNIX`, I think this whole file (`tcp.rs`) and the `TcpProxy` struct really need to be renamed. It is confusing that a `TcpProxy` also handles unix stream sockets. I would suggest naming it `TsiStreamProxy`.
(the name should contain `TSI` because we also have a `UnixProxy` that doesn't use TSI).

src/devices/src/virtio/vsock/tsi_dgram.rs
```diff
@@ -48,18 +49,21 @@ impl UdpProxy {
...
-        )
-        .map_err(ProxyError::CreatingSocket)?;
+        let family = match family as i32 {
+            AF_INET => AddressFamily::Inet,
+            AF_INET6 => AddressFamily::Inet6,
+            AF_UNIX => AddressFamily::Unix,
+            _ => return Err(ProxyError::InvalidFamily),
+        };
+
+        let fd = socket(family, SockType::Datagram, SockFlag::empty(), None)
```
mtjhrc: Same here, file/struct should to be renamed to something like `TsiDgramProxy`

src/devices/src/virtio/vsock/tsi_stream.rs
slp: Makes sense, I've just done that.

mtjhrc: Actually, some of the log messages also contain "tcp: ", "vsock::tcp: "  at the beginning. Which is useless, because env_logger already prefixes the name of the file. It would be nice to remove that, now that it's not really true.

slp: Right, I've cleaned up the messages.

slp:
@mtjhrc do you need more changes in this PR?

mtjhrc:
Code LGTM! 
I wanted to test it / play around with it a bit (like you asked), before approving it. I should get to it later today. 

I am also wondering what happens to anonymous namespace unix domain sockets. Does the kernel side attempt to hijack them into TSI too?

mtjhrc (changes requested):
> I am also wondering what happens to anonymous namespace unix domain sockets. Does the kernel side attempt to hijack them into TSI too?

I have managed to test this, yes they are hijacked too, forwarding them works correctly (works with `podman --network host`).

Regarding the PR, there is an issue when resolving paths of unix domain sockets (even within the guest) see https://github.com/containers/libkrunfw/issues/103

src/devices/src/virtio/vsock/tcp.rs
```diff
@@ -1,15 +1,20 @@
...
 use std::num::Wrapping;
 use std::os::fd::{FromRawFd, OwnedFd};
 use std::os::unix::io::{AsRawFd, RawFd};
 use std::sync::{Arc, Mutex};
 
+use libc::AF_INET;
+#[cfg(target_os = "linux")]
+use libc::EINVAL;
+#[cfg(target_os = "macos")]
+use libc::EINVAL;
```
jakecorrenti: nit (nonblocking): do we need the differentiation here? We only support these two OSes, but would it be a poor assumption for us to just import `libc::EINVAL`?

src/devices/src/virtio/vsock/tcp.rs
```diff
@@ -83,7 +89,12 @@ impl TcpProxy {
             Err(e) => error!("couldn't obtain fd flags id={id}, err={e}"),
         };
 
-        setsockopt(&fd, sockopt::ReusePort, &true).map_err(ProxyError::SettingReusePort)?;
+        if family == AddressFamily::Unix {
+            setsockopt(&fd, sockopt::ReuseAddr, &true).map_err(ProxyError::SettingReusePort)?;
```
jakecorrenti: Should we return the `ProxyError::SettingReuseAddr` error instead?

jakecorrenti (commented):
Just one error handling nitpick, but otherwise code lgtm

src/devices/src/virtio/vsock/tcp.rs
slp: Nice catch, fixed, thanks!

slp:
> Regarding the PR, there is an issue when resolving paths of unix domain sockets (even within the guest) see [containers/libkrunfw#103](https://github.com/containers/libkrunfw/issues/103)

I've extended the `Networking` section in the README.md clarifying this as a know limitation.

@mtjhrc Could you please take another look to see if we can get this one merged? Thanks!

README.md
```diff
@@ -58,11 +58,48 @@ Each variant generates a dynamic library with a different name (and ```soname```
...
+
+While defining the security implementation of your environment, you should think about the guest and the VMM as a single entity. To prevent the guest from accessing host's resources, you need to use the host's OS security features to run the VMM inside an isolated context. On Linux, the primary mechanism to be used for this purpose is namespaces. Single-user systems may have a more relaxed security policy and just ensure the VMM runs with a particular UID/GID.
+
+While most virtio devices allow the guest to access resources from the host, two of them require special consideration when used: virtio-fs and virtio-vsock+TSI.
+
+### virtio-fs
+
+When exposing a directory in a filesystem from the host to the guest through virtio-fs devices configured with `krun_set_root` and/or `krun_add_virtiofs`, libkrun **does not** provide any protection against the guest attempting to access other directories in the same filesystem, or even other filesystems in the host.
+
+A mount point isolation mechanism from the host should be used in combination with virtio-fs.
```
ggoodman, line 98:
```suggestion

A mount point isolation mechanism from the host should be used in combination with virtio-fs.

In addition, when using virtio-fs, a guest may exhaust filesystem resources such as inode limits and disk capacity. Controls should be implemented on the host to mitigate this.
```

slp: Oops... sorry, this came in literally one minute after I merged the PR. Could you please create another PR with the additional text?

ggoodman: I created #453.

PR #412, opened 2025-09-19
Generalize external firmware support

slp:
So far, only the EFI flavor was capable of booting from a firmware. Let's generalize firmware loading into the generic flavor, introducing a new API `krun_set_firmware` to allow users configuring the path to the firmware to be loaded into the VM.

### DIFF
--- a/src/arch/src/lib.rs
+++ b/src/arch/src/lib.rs
@@ -44,8 +44,8 @@ pub mod x86_64;
 #[cfg(target_arch = "x86_64")]
 pub use crate::x86_64::{
     arch_memory_regions, configure_system, get_kernel_start, initrd_load_addr,
-    layout::CMDLINE_MAX_SIZE, layout::IRQ_BASE, layout::IRQ_MAX, Error, layout::FIRMWARE_SIZE, layout::FIRMWARE_START,
-    layout::MMIO_MEM_START, layout::RESET_VECTOR,
+    layout::CMDLINE_MAX_SIZE, layout::FIRMWARE_SIZE, layout::FIRMWARE_START, layout::IRQ_BASE,
+    layout::IRQ_MAX, layout::MMIO_MEM_START, layout::RESET_VECTOR, Error,
 };
 
 /// Type for returning public functions outcome.
--- a/src/arch/src/x86_64/mod.rs
+++ b/src/arch/src/x86_64/mod.rs
@@ -17,10 +17,10 @@ pub mod msr;
 /// Logic for configuring x86_64 registers.
 pub mod regs;
 
-use crate::{ArchMemoryInfo, InitrdConfig};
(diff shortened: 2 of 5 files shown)


.github/workflows/code_quality-aarch64-darwin.yml
```diff
@@ -22,6 +22,9 @@ jobs:
       - name: Install dependencies
         run: brew tap slp/krunkit && brew install virglrenderer clang-format llvm
 
+      - name: Create a fake init
+        run: touch init/init
```
tylerfanelli, line 26: Why is this needed?

include/libkrun.h
```diff
@@ -47,23 +46,27 @@ int32_t krun_set_log_level(uint32_t level);
...
  *  "options"   - Bitmask of logging options, use 0 for default options.
- *                KRUN_LOG_OPTION_NO_ENV to disallow environment variables to override these settings.
+ *                KRUN_LOG_OPTION_NO_ENV to disallow environment variables to
+ * override these settings.
  *
  * Returns:
  *  Zero on success or a negative error number on failure.
  */
-int32_t krun_init_log(int target_fd, uint32_t level, uint32_t style, uint32_t options);
+int32_t krun_init_log(int target_fd, uint32_t level, uint32_t style,
```
tylerfanelli: Seems like there' some unrelated changes other than adding `krun_set_firmware` in this commit. Not a problem, just making sure you're aware (perhaps this was from some C formatting tool?)

tylerfanelli (dismissed):
LGTM. Two non-blocking comments.

include/libkrun.h
mtjhrc: Yeah, please remove this, the file doesn't currently `clang-format` nicely, because of our syntax for the doc comments (we should consider maybe using doxygen, which it might understand).

.github/workflows/code_quality-aarch64-darwin.yml
slp: To satisfy this in virtio-fs:

```
static INIT_BINARY: &[u8] = include_bytes!("../../../../../../init/init");
```

This was gated out in the EFI variant, but with the unification we need it for building both variants.

include/libkrun.h
slp: Thanks, vscode did this silently for me. I've restored sanity...

jakecorrenti (approved):
Changes lgtm.

I'm still trying to understand the bigger picture... Am I understanding correctly that this opens the door to allowing the user to specify, for example, they want to use edk2 instead of libkrunfw?

slp:
> Changes lgtm.
> 
> I'm still trying to understand the bigger picture... Am I understanding correctly that this opens the door to allowing the user to specify, for example, they want to use edk2 instead of libkrunfw?

The door was already open with `krun_set_kernel`, but now we're moving even further by allowing loading a FW instead of a kernel. This allows for boot-from-disk use cases.

PR #416, opened 2025-09-24
Fix gui_vm compilation + warning, missing examples/Cargo.lock

mtjhrc:


jakecorrenti (approved):
Code LGTM.

I think it would be nice for the first commit to just have a quick message saying which packages were updated/added, but that's just me.

mtjhrc:
> Code LGTM.
> 
> I think it would be nice for the first commit to just have a quick message saying which packages were updated/added, but that's just me.

Sure done. To clarify though, the the packages were already "updated" (cargo build would use the newer versions) - we were just missing the Cargo.lock to lock the minor versions and hashes.

jakecorrenti (approved):
lgtm.

Seems there's an issue with the CI, though.

mtjhrc:
Yeah, I'm looking into the CI, but this PR doesn't touch the integration test compilation which broke at all. (so I guess it broke when the rust tool chain updates itself?), weird...

PR #417, opened 2025-09-24
CI improvements

mtjhrc:
This refactors the CI code, removes the usage of the no longer maintained `actions-rs` GitHub action, and adds more checks:
- check if we have up-to-date `Cargo.lock` files committed
- check (`cargo fmt` and `clippy`)  the examples workspace
- check (`cargo-fmt` and `clippy`) the tests workspace

This also resolves the small issues (rustfmt + clippy) uncovered by the new checks


.github/workflows/code-quality.yml
```diff
@@ -0,0 +1,165 @@
...
+          - name: "Examples (Linux aarch64)"
+            runner: ubuntu-24.04-arm
+    runs-on: ${{ matrix.runner }}
+    steps:
+      - uses: actions/checkout@v4
+      
+      - name: Setup build environment
+        uses: ./.github/actions/setup-build-env
+      
+      - name: Cache GLib 2.82 + GTK 4.16 build
```
bilelmoussaoui, line 79: What is the motivation behind building GTK manually?

bilelmoussaoui: Sorry, I just saw that in the commit message but wouldn't using a fedora container be enough?

bilelmoussaoui: In any case, gtk would build glib automatically if the required version is not installed. So you can drop the glib build part

mtjhrc: Yeah I realized maybe I should have done that.  But then we would have to maintain both list of build libraries for Fedora and for Ubuntu in the CI (not sure if we can get KVM in a container).  Anyways I'm hoping we can just drop this once Ubuntu 26.04 LTS is released.

bilelmoussaoui: > In any case, gtk would build glib automatically if the required version is not installed. So you can drop the glib build part

This would only work if you pull from git though but i guess what you have is good enough. When it comes to build deps, installing libgtk4-dev should pull all the needed deps i guess

slp:
This one looks great, nice work here @mtjhrc

Issue #420, opened 2025-09-30
Support for Raspberry Pi


sgsaenger:
Hi, i'm having trouble getting libkrun to work on Pis, even though KVM works standalone (QEMU/libvirt).

Output from CM4:
```
$ podman run --rm -it --runtime=/usr/local/bin/krun alpine

thread '<unnamed>' panicked at src/vmm/src/linux/vstate.rs:445:30:
Error creating the Kvm object: Error(13)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
fatal runtime error: failed to initiate panic, error 2389678208, aborting
```

Output from CM5:
```
podman run --rm -it --runtime=/usr/local/bin/krun alpine

thread '<unnamed>' panicked at src/devices/src/legacy/kvmgicv3.rs:42:59:
called `Result::unwrap()` on an `Err` value: Error(19)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
fatal runtime error: failed to initiate panic, error 3281714368, aborting
```

The same happens when running the included the example:
```
./chroot_vm rootfs_fedora/ /bin/sh

thread '<unnamed>' panicked at src/devices/src/legacy/kvmgicv3.rs:42:59:
called `Result::unwrap()` on an `Err` value: Error(19)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
fatal runtime error: failed to initiate panic, error 372322368, aborting
fish: Job 1, './chroot_vm rootfs_fedora/ /bin…' terminated by signal SIGABRT (Abort)
```

Using the latest versions: `libkrun` v1.15.1, `libkrunfw` v4.10.0, `crun` v1.24, `podman` v5.6.1 on current bookworm-based PiOS.
Are there any known limitations prohibiting this use case?

slp:
> Output from CM4:

```
$ podman run --rm -it --runtime=/usr/local/bin/krun alpine

thread '<unnamed>' panicked at src/vmm/src/linux/vstate.rs:445:30:
Error creating the Kvm object: Error(13)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
fatal runtime error: failed to initiate panic, error 2389678208, aborting
```

This one happened because libkrun couldn't open `/dev/kvm`, most likely because the user running podman didn't have permission to access it (seems like Debian default permissions requires users to be part of the `kvm` group).


> Output from CM5:

```
podman run --rm -it --runtime=/usr/local/bin/krun alpine

[...]

PR #421, opened 2025-10-01
Fall back to GICv2 if GICv3 is not supported

slp:
The SoC in some popular boards (namely, the RPi family) doesn't support an architected vGIC, which is required for requesting KVM the instantiation of a GICv3. To relieve the users from having to configure the gic version manually, try first to instantiate a GICv3, and fall back to a GICv2 if it fails.

Fixes: #420

### DIFF
--- a/src/vmm/src/builder.rs
+++ b/src/vmm/src/builder.rs
@@ -30,8 +30,6 @@ use crate::vmm_config::external_kernel::{ExternalKernel, KernelFormat};
 use crate::vmm_config::net::NetBuilder;
 #[cfg(all(target_os = "linux", target_arch = "riscv64"))]
 use devices::legacy::KvmAia;
-#[cfg(all(target_os = "linux", target_arch = "aarch64"))]
-use devices::legacy::KvmGicV3;
 #[cfg(target_arch = "x86_64")]
 use devices::legacy::KvmIoapic;
 use devices::legacy::Serial;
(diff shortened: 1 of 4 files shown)


src/vmm/src/builder.rs
```diff
@@ -837,10 +837,18 @@ pub fn build_microvm(
...
-        )))));
+        intc = {
+            // The SoC in some popular boards (namely, the RPi family) doesn't support an
+            // architected vGIC, which is required for requesting KVM the instantiation of a
+            // GICv3. To relieve the users from having to configure the gic version manually,
+            // try first to instantiate a GICv3, and fall back to a GICv2 if it fails.
+            let vcpu_count = vm_resources.vm_config().vcpu_count.unwrap() as u64;
+            let gic = match KvmGicV3::new(vm.fd(), vcpu_count) {
+                Ok(gicv3) => IrqChipDevice::new(Box::new(gicv3)),
+                Err(_) => IrqChipDevice::new(Box::new(KvmGicV2::new(vm.fd(), vcpu_count))),
```
mtjhrc: I feel like we should log the error here, and say we falling back to GicV2, no? (Maybe on like `info!` level, not error)

slp: Done. I've put that with a `warn!`

mtjhrc (approved):
I'm not able to test this, but code LGTM

jakecorrenti (approved):
Code LGTM. None of my Pis are available either (running various servers), so I can't test it.

Food for thought: I wonder if down the line we should consider making some of this code more generic? It seems to be fairly similar to the GicV3 implementation; maybe we could abstract some of it away.

PR #423, opened 2025-10-10
virtio/console: Remove use of unstable library feature

tylerfanelli:
`is_multiple_of` is an unstable library feature.


tylerfanelli:
My local environment warned about the use of `is_multiple_of`, yet the CI tests suggest to use it instead. I'll err towards the CI and close this PR.

PR #424, opened 2025-10-10
init: allow running the user-provided init as pid 1 (fix #223)

### LINKED ISSUE #223: Running kind/systemd in krun container fails
I tried running [kind](https://kind.sigs.k8s.io/) with the podman runtime set to `krun` and I'm getting the following error:

```
$ KIND_EXPERIMENTAL_PROVIDER=podman kind create cluster --retain
using podman due to KIND_EXPERIMENTAL_PROVIDER
enabling experimental podman provider
Creating cluster "kind" ...
 ✓ Ensuring node image (kindest/node:v1.31.0) 🖼
 ✗ Preparing nodes 📦
ERROR: failed to create cluster: could not find a log line that matches "Reached target .*Multi-User System.*|detected cgroup v1"
```

Looking at the logs of the node container I see the following:

```
$ podman logs kind-control-plane
INFO: changing snapshotter from "overlayfs" to "fuse-overlayfs"
INFO: enabling containerd-fuse-overlayfs service
Created symlink /etc/systemd/system/multi-user.target.wants/containerd-fuse-overlayfs.service → /etc/systemd/system/containerd-fuse-overlayfs.service.
INFO: ensuring we can execute mount/umount even with userns-remap
INFO: remounting /sys read-only
INFO: making mounts shared
INFO: detected cgroup v2
INFO: clearing and regenerating /etc/machine-id
Initializing machine ID from random generator.
INFO: setting iptables to detected mode: legacy
INFO: detected IPv4 address:
INFO: detected IPv6 address:
INFO: starting init
Couldn't find an alternative telinit implementation to spawn.
```

So kind fails at starting systemd inside the container. I have verified this by using a very minimal systemd container as described here: https://developers.redhat.com/blog/2019/04/24/how-to-run-systemd-in-a-container#enter_podman

Running such a container results in the same error. I think the issue boils down to krun using PID 1 for `/init.krun` and not the container entrypoint as other runtimes (e.g. crun) do. So I'm not entirely sure this is fixable at all.

valpackett:
Some init systems like systemd absolutely need to run as PID 1. Add an environment variable to allow doing so at the cost of the exit code feature.

---

Yes, this is kinda in the direction of "not-so-micro" VMs, but we wanted to try a tiny minimal systemd setup for our own project where we need to run user-configurable background services (like VPN clients) for desktop apps. #223 mentions the systemd-in-a-container use case for k8s stuff. So hopefully this would be acceptable? (after polishing it a bit more: I guess it should be configurable from the JSON file as well..)

### DIFF
--- a/init/init.c
+++ b/init/init.c
@@ -2,6 +2,7 @@
 #include <errno.h>
 #include <fcntl.h>
 #include <limits.h>
+#include <stdbool.h>
 #include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
@@ -1044,6 +1045,7 @@ int main(int argc, char **argv)
     int sockfd;
     int status;
     int saved_errno;
+    bool init_pid1 = false;
     char localhost[] = "localhost\0";
     char *hostname;
     char *krun_home;
@@ -1052,6 +1054,7 @@ int main(int argc, char **argv)
     char *krun_root;
     char *krun_root_fstype;
     char *krun_root_options;
+    char *env_init_pid1;
     char *config_workdir, *env_workdir;
     char *rlimits;
     char **config_argv, **exec_argv;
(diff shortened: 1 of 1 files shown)


mtjhrc:
> So hopefully this would be acceptable? 

Yes definitely. In fact I have also ran into this limitation and I was already thinking we should implement a mechanism like this eventually. Thanks for working on it!

PR #426, opened 2025-10-10
Allow virtio_gpu to auto-fallback to a safe rutabaga config

slp:
Instead of generating a panic when rutabaga initialization fails, let's have virtio_gpu retry with a safe configuration.

### DIFF
--- a/src/devices/src/virtio/gpu/virtio_gpu.rs
+++ b/src/devices/src/virtio/gpu/virtio_gpu.rs
@@ -210,17 +210,14 @@ impl VirtioGpu {
         })
     }
 
-    #[allow(clippy::too_many_arguments)]
-    pub fn new(
+    pub fn create_rutabaga(
         mem: GuestMemoryMmap,
         queue_ctl: Arc<Mutex<VirtQueue>>,
         interrupt: InterruptTransport,
+        fence_state: Arc<Mutex<FenceState>>,
         virgl_flags: u32,
-        #[cfg(target_os = "macos")] map_sender: Sender<WorkerMessage>,
         export_table: Option<ExportTable>,
-        displays: Box<[DisplayInfo]>,
-        display_backend: DisplayBackend,
-    ) -> Self {
+    ) -> Option<Rutabaga> {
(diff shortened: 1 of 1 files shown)


slp:
@valpackett PTAL

src/rutabaga_gfx/src/virgl_renderer.rs
```diff
@@ -335,6 +324,19 @@ impl VirglRenderer {
...
+            // Initialize it only once and use the non-send/non-sync Renderer struct to keep things tied
+            // to whichever thread called this function first.
+            static INIT_ONCE: AtomicBool = AtomicBool::new(false);
+            if INIT_ONCE
+                .compare_exchange(false, true, Ordering::Acquire, Ordering::Acquire)
+                .is_err()
+            {
+                return Err(RutabagaError::AlreadyInUse);
+            }
+        }
```
mtjhrc: I think the original intention of this was to prevent calling `virgl_renderer_init` when it was already initialized, no? But this change allows initializing virglrenderer a second time. I'm not sure if  that is even a problem though... 

If we want to keep the check working you can is still keep the check before `virgl_renderer_init`:
```Rust
if INIT_ONCE.compare_exchange(false, true, Ordering::SeqCst, Ordering::SeqCst).is_err() { 
    return `RutabagaError::AlreadyInUse` 
}
```
But after calling `virgl_renderer_init`, set it to the actual result:
INIT_ONCE =  (virgl_renderer_init() == 0)

(Not sure about the memory ordering...)

slp: I've fixed by keeping the check at the beginning of the method and storing `false` if the initialization fails. TBH, the whole use of atomics there is pretty overkill, knowing the crate isn't thread-safe and we're driving it from a single thread, but since it's already there let's keep it.

mtjhrc: Hmm, I think the atomic are ok (short of an stdlib utility to abstract this bool). It's single threaded, but that's precisely why you need to enforce that at initialization time for the crate to be sound.  Methods can then use `&self`, being `!Send`, `!Sync` to enforce being single threaded.

PR #427, opened 2025-10-14
libkrun: only create shutdown_efd on macOS

slp:
We're only using shutdown_efd on macOS, so avoid creating it on other platforms.

Combined with the fact that crun closes every file descriptor between krun_create_ctx() and krun_start_enter(), this was triggering a double close() that ended closing a random file descriptor.

Reported-by: Matej Hrica <mhrica@redhat.com>


jakecorrenti (approved):
LGTM

PR #428, opened 2025-10-15
Enable relaxed sync by default on macOS

jakecorrenti:
On macOS, `std::fs::File::flush()` uses `F_FULLFSYNC`, which syncs the
buffers to the drive and asks the drive to wait for flushing all the
data to the permanent storage. In contrast, Apple
Virtualization.Framework only issues an `fsync()`, which doesn't ask the
drive to flush its buffered data.

When on macOS, tell `imago` to use relaxed sync to match the behavior of
Virtualization.Framework by default.

Related to https://github.com/containers/podman/issues/27216

### DIFF
--- a/src/devices/src/virtio/block/device.rs
+++ b/src/devices/src/virtio/block/device.rs
@@ -18,9 +18,10 @@ use std::result;
 use std::sync::Arc;
 use std::thread::JoinHandle;
 
-use imago::file::File as ImagoFile;
-use imago::qcow2::Qcow2;
-use imago::SyncFormatAccess;
+use imago::{
+    file::File as ImagoFile, qcow2::Qcow2, raw::Raw, DynStorage, Storage, StorageOpenOptions,
+    SyncFormatAccess,
+};
 use log::{error, warn};
 use utils::eventfd::{EventFd, EFD_NONBLOCK};
 use virtio_bindings::{
@@ -64,14 +65,14 @@ impl CacheType {
 /// Helper object for setting up all `Block` fields derived from its backing file.
 pub(crate) struct DiskProperties {
     cache_type: CacheType,
-    pub(crate) file: Arc<SyncFormatAccess<ImagoFile>>,
+    pub(crate) file: Arc<SyncFormatAccess<Box<dyn DynStorage>>>,
     nsectors: u64,
     image_id: Vec<u8>,
 }
 
 impl DiskProperties {
     pub fn new(
-        disk_image: Arc<SyncFormatAccess<ImagoFile>>,
+        disk_image: Arc<SyncFormatAccess<Box<dyn DynStorage>>>,
         disk_image_id: Vec<u8>,
         cache_type: CacheType,
     ) -> io::Result<Self> {
@@ -94,7 +95,7 @@ impl DiskProperties {
         })
     }
 
-    pub fn file(&self) -> &SyncFormatAccess<ImagoFile> {
+    pub fn file(&self) -> &SyncFormatAccess<Box<dyn DynStorage>> {
         self.file.as_ref()
     }
 
@@ -176,7 +177,7 @@ pub struct Block {
     // Host file and properties.
     disk: Option<DiskProperties>,
     cache_type: CacheType,
-    disk_image: Arc<SyncFormatAccess<ImagoFile>>,
+    disk_image: Arc<SyncFormatAccess<Box<dyn DynStorage>>>,
     disk_image_id: Vec<u8>,
     worker_thread: Option<JoinHandle<()>>,
     worker_stopfd: EventFd,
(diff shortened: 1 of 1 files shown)


jakecorrenti:
Before:
```
mysql> CREATE TABLE IF NOT EXISTS customers (  id INT AUTO_INCREMENT PRIMARY KEY,  customer_name TEXT,  created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP ) ENGINE=INNODB;
Query OK, 0 rows affected (0.13 sec)

mysql> INSERT INTO customers (id, customer_name) SELECT n, CONCAT('Customer', n)  FROM ( select a.N + b.N * 10 + c.N *
    -> 100 + d.N * 1000 + e.N * 10000 + f.N * 100000 + g.N * 1000000 + 1 N from (select 0 as N union all select 1 union all select 2 union all select 3 union all select 4 union all select 5 union all select 6 union all select 7 union all select
    -> 8 union all select 9) a  , (select 0 as N union all select 1 union all select 2 union all select 3 union all select 4 union all select 5 union all select 6 union all select 7 union all select 8 union all select 9) b  , (select 0 as N union all select 1 union all se
lect 2 union all select 3 union all select 4 union all select 5 union all select 6 union all select 7 union all select 8 union all select 9) c  , (select 0 as N union all select 1 union all select 2 union all select 3 union all select 4 union all select 5 union all select
 6 union all select 7 union all select 8 union all select
    -> 9) d  , (select 0 as N union all select 1 union all select 2 union all select 3 union all select 4 union all select 5 union all select 6 union all select 7 union all select 8 union all select 9) e  , (select 0 as N union all select 1 union all select 2 union all se
lect 3 union all select 4 union all select 5 union all select 6 union all select 7 union all select 8 union all select 9) f  , (select 0 as N union all select 1 union all select 2 union all select 3 union all select 4 union all select 5 union all select 6 union all select
 7 union all select 8 union all select 9) g ) t;
Query OK, 10000000 rows affected (6 min 6.51 sec)
Records: 10000000  Duplicates: 0  Warnings: 0
```

After:
```
mysql> CREATE TABLE IF NOT EXISTS customers (  id INT AUTO_INCREMENT PRIMARY KEY,  customer_name TEXT,  created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP ) ENGINE=INNODB;
Query OK, 0 rows affected (0.03 sec)

mysql> INSERT INTO customers (id, customer_name) SELECT n, CONCAT('Customer', n)  FROM ( select a.N + b.N * 10 + c.N *
    -> 100 + d.N * 1000 + e.N * 10000 + f.N * 100000 + g.N * 1000000 + 1 N from (select 0 as N union all select 1 union all select 2 union all select 3 union all select 4 union all select 5 union all select 6 union all select 7 union all select
    -> 8 union all select 9) a  , (select 0 as N union all select 1 union all select 2 union all select 3 union all select 4 union all select 5 union all select 6 union all select 7 union all select 8 union all select 9) b  , (select 0 as N union all select 1 union all se
lect 2 union all select 3 union all select 4 union all select 5 union all select 6 union all select 7 union all select 8 union all select 9) c  , (select 0 as N union all select 1 union all select 2 union all select 3 union all select 4 union all select 5 union all select
 6 union all select 7 union all select 8 union all select
    -> 9) d  , (select 0 as N union all select 1 union all select 2 union all select 3 union all select 4 union all select 5 union all select 6 union all select 7 union all select 8 union all select 9) e  , (select 0 as N union all select 1 union all select 2 union all se
lect 3 union all select 4 union all select 5 union all select 6 union all select 7 union all select 8 union all select 9) f  , (select 0 as N union all select 1 union all select 2 union all select 3 union all select 4 union all select 5 union all select 6 union all select
 7 union all select 8 union all select 9) g ) t;
Query OK, 10000000 rows affected (46.92 sec)
Records: 10000000  Duplicates: 0  Warnings: 0
```

jakecorrenti:
CC for awareness @Honny1 @baude

nirs:
@jakecorrenti fsync() on macOS is not safe. It may be a good default since other solutions are using the same, but user needs a way to enable real fsync() (like we have on Linux).

We should have a way to configure this via libkrun API, and expose the configuration to users in krunkit, similar to this vfkit proposal:
https://github.com/crc-org/vfkit/issues/383

jakecorrenti:
> @jakecorrenti fsync() on macOS is not safe. It may be a good default since other solutions are using the same, but user needs a way to enable real fsync() (like we have on Linux).
> 
> 
> 
> We should have a way to configure this via libkrun API, and expose the configuration to users in krunkit, similar to this vfkit proposal:
> 
> https://github.com/crc-org/vfkit/issues/383

Yes, you're correct. We are going to come back in with another PR to make the necessary changes to the libkrun API, then subsequently krunkit.

This specific PR is so we can get the performance regression resolved in Podman 5.7 whose release process is starting next week.

jakecorrenti:
@slp @tylerfanelli @mtjhrc PTAL, somewhat time sensitive for podman 5.7

src/devices/src/virtio/block/device.rs
```diff
@@ -20,7 +20,8 @@ use std::thread::JoinHandle;
 
 use imago::file::File as ImagoFile;
 use imago::qcow2::Qcow2;
-use imago::SyncFormatAccess;
+use imago::raw::Raw;
+use imago::{DynStorage, Storage, StorageOpenOptions, SyncFormatAccess};
```
tylerfanelli:
```suggestion
use imago::{raw::Raw, DynStorage, Storage, StorageOpenOptions, SyncFormatAccess};
```

jakecorrenti: Done.

tylerfanelli (approved):

Issue #431, opened 2025-10-18
devices: cannot find function `statx` in crate `libc`


peppergrayxyz:
Build fails on llvm/musl/libcxx (https://github.com/containers/libkrun/issues/429)

musl 1.2.5 added support for the statx system call & rust ships unknown-linux-musl with musl 1.2.3 therefore statx doesn't exist

```Rust
   Compiling devices v0.1.0 (/home/pepper/dev/ghrepos/libkrun/src/devices)
error: couldn't read `src/devices/src/virtio/fs/linux/../../../../../../init/init`: No such file or directory (os error 2)
  --> src/devices/src/virtio/fs/linux/passthrough.rs:36:29
   |
36 | static INIT_BINARY: &[u8] = include_bytes!("../../../../../../init/init");
   |                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error[E0412]: cannot find type `statx` in crate `libc`
   --> src/devices/src/virtio/fs/linux/passthrough.rs:150:39
    |
150 |       let mut stx = MaybeUninit::<libc::statx>::zeroed();
    |                                         ^^^^^ help: a struct with a similar name exists: `stat`
    |
   ::: /home/pepper/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/libc-0.2.172/src/unix/linux_like/linux/musl/b64/aarch64/mod.rs:10:1
    |
10  | / s! {
11  | |     pub struct stat {
12  | |         pub st_dev: crate::dev_t,
13  | |         pub st_ino: crate::ino_t,
...   |
112 | | }
    | |_- similarly named struct `stat` defined here

error[E0425]: cannot find function `statx` in crate `libc`
   --> src/devices/src/virtio/fs/linux/passthrough.rs:158:15
    |
158 |         libc::statx(                                                                                                      |               ^^^^^ help: a function with a similar name exists: `stat`
    |
   ::: /home/pepper/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/libc-0.2.172/src/unix/mod.rs:876:5
    |
876 |     pub fn stat(path: *const c_char, buf: *mut stat) -> c_int;
    |     ---------------------------------------------------------- similarly named function `stat` defined here
                                                                                                                            error[E0425]: cannot find value `STATX_BASIC_STATS` in crate `libc`
   --> src/devices/src/virtio/fs/linux/passthrough.rs:162:19
    |
[...]

PR #432, opened 2025-10-18
passthrough.rs fails with cannot find type `statx` in crate `libc`

peppergrayxyz:
musl 1.2.5 added support for the statx system call rust ships unknown-linux-musl with musl 1.2.3, milage on other libs may varay

Enable portability by repling libc::statx with rustix::statx

fixes https://github.com/containers/libkrun/issues/431

### DIFF
--- a/src/devices/Cargo.toml
+++ b/src/devices/Cargo.toml
@@ -37,6 +37,7 @@ utils = { path = "../utils" }
 polly = { path = "../polly" }
 rutabaga_gfx = { path = "../rutabaga_gfx", features = ["virgl_renderer", "virgl_renderer_next"], optional = true }
 imago = { version = "0.1.4", features = ["sync-wrappers", "vm-memory"] }
+rustix = { version = "1.1.2", features = ["fs"] }
 
 [target.'cfg(target_os = "macos")'.dependencies]
 hvf = { path = "../hvf" }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -441,6 +441,7 @@ dependencies = [
  "pipewire",
  "polly",
  "rand",
+ "rustix",
  "rutabaga_gfx",
  "thiserror 2.0.12",
  "utils",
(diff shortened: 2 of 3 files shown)


src/devices/Cargo.toml
```diff
@@ -37,6 +37,7 @@ utils = { path = "../utils" }
 polly = { path = "../polly" }
 rutabaga_gfx = { path = "../rutabaga_gfx", features = ["virgl_renderer", "virgl_renderer_next"], optional = true }
 imago = { version = "0.1.4", features = ["sync-wrappers", "vm-memory"] }
+rustix = { version = "1.1.2", features = ["fs"] }
```
slp, line 40: I think adding a new dependency just for this is a bit overkill. Let's simply wait for the rust toolchain to ship a newer version of musl, shouldn't take long.

mtjhrc: Yeah, adding the dependency for such a simple thing seems like an overkill to me too.  

An alternative could be to use `libc::syscall` directly.

musl 1.2.5 was released on February 29, 2024, so it's taking a while :smile:. Nevertheless seems like we will get it soon-ish:
https://github.com/rust-lang/rust/pull/142682

Btw. I was actually considering whether we may want to start using `rustix` over `nix` - it seemed to be nicer in some areas than nix, but it's hard to say...

peppergrayxyz: [statx-sys](https://github.com/oxalica/statx-sys) does that, but has been deprecated in favor of `libc::statx` and `rustix`. It doesn't support `stx_mnt_id` but can be patched easily. But we could use it, thereby avoiding pulling in `rustix` and also writing the wrapper from scratch.

I created a draft here:
- libkrun patch: https://github.com/containers/libkrun/commit/66a4c5fdb6be5205f2edd6ce658a079aca68ef7e
- statx-sys patch: https://github.com/oxalica/statx-sys/commit/c0cd9f6b654279be5b4e37fd06d17004687b2ed7

I'd move the statx-sys fork to libkrun or just include the files in the project if we wanna go forward with this.

slp:
@peppergrayxyz Please consider using `libc::syscall` as @mtjhrc suggested. You can borrow the code from virtiofsd: https://gitlab.com/virtio-fs/virtiofsd/-/blob/main/src/passthrough/stat.rs?ref_type=heads#L102

polarathene:
FWIW, the [upgrade to musl 1.2.5 will land in the upcoming Rust 1.93](https://github.com/rust-lang/rust/issues/150760) release.

slp:
We're already in 1.94, so let's close this one.

Issue #435, opened 2025-10-19
devices: this function takes 3 arguments but 2 arguments were supplied


peppergrayxyz:
Build fails on aarch64/riscv64 (#429):

arch_memory_regions  takes 3 arguments but 2 arguments were supplied

```Rust
error[E0061]: this function takes 3 arguments but 2 arguments were supplied
--> src/arch/src/aarch64/linux/regs.rs:156:36
 |
156 |         let (_mem_info, regions) = arch_memory_regions(layout::FDT_MAX_SIZE + 0x1000, 0);
 |                                    ^^^^^^^^^^^^^^^^^^^---------------------------------- argument #3 of type `Option<usize>` is missing
 |
note: function defined here
--> src/arch/src/aarch64/mod.rs:41:8
 |
41  | pub fn arch_memory_regions(
 |        ^^^^^^^^^^^^^^^^^^^
...
44  |     firmware_size: Option<usize>,
 |     ----------------------------
help: provide the argument
 |
156 |         let (_mem_info, regions) = arch_memory_regions(layout::FDT_MAX_SIZE + 0x1000, 0, /* Option<usize> */);
 |                                                                                        +++++++++++++++++++++
```

slp:
Fixed by #487

PR #436, opened 2025-10-19
test fails with arch_memory_regions missing argument #3

peppergrayxyz:
arch_memory_regions needs 3 arguments, while it previously used to need 2. This was updated in arch/src/x86, but not in aarch64 and riscv64

Added a 3rd parameter 'None' where it was missing.

Fixes: #435


slp:
Fixed by #487

Issue #439, opened 2025-10-21
devices: the trait bound `DummyIrqChip: GICDevice/AIADevice` is not satisfied


peppergrayxyz:
Build fails on arch64 (https://github.com/containers/libkrun/issues/429):

```Rust
Compiling devices v0.1.0
error[E0277]: the trait bound `DummyIrqChip: GICDevice` is not satisfied
   --> src/devices/src/legacy/irqchip.rs:171:23
    |
171 |     impl IrqChipT for DummyIrqChip {
    |                       ^^^^^^^^^^^^ the trait `GICDevice` is not implemented for `DummyIrqChip`
    |
    = help: the following other types implement trait `GICDevice`:
              KvmGicV2
              KvmGicV3
              irqchip::IrqChipDevice
note: required by a bound in `irqchip::IrqChipT`
   --> src/devices/src/legacy/irqchip.rs:127:33
    |
127 | pub trait IrqChipT: BusDevice + GICDevice {
    |                                 ^^^^^^^^^ required by this bound in `IrqChipT`
```

and RISC-V:

```Rust
Compiling devices v0.1.0
error[E0277]: the trait bound `DummyIrqChip: AIADevice` is not satisfied
   --> src/devices/src/legacy/irqchip.rs:191:23
    |
191 |     impl IrqChipT for DummyIrqChip {
    |                       ^^^^^^^^^^^^ the trait `AIADevice` is not implemented for `DummyIrqChip`
    |
    = help: the following other types implement trait `AIADevice`:
              KvmAia
              irqchip::IrqChipDevice
note: required by a bound in `irqchip::IrqChipT`
   --> src/devices/src/legacy/irqchip.rs:138:33
    |
138 | pub trait IrqChipT: BusDevice + AIADevice {
    |                                 ^^^^^^^^^ required by this bound in `IrqChipT`
```

PR #443, opened 2025-10-21
Creat DummyIrqChips GICDevice and AIADevice

peppergrayxyz:
Create missing devices for aarch64 and riscv64

Fixes: #439

### DIFF
--- a/src/devices/src/legacy/irqchip.rs
+++ b/src/devices/src/legacy/irqchip.rs
@@ -166,6 +166,47 @@ pub mod test_utils {
         }
     }
 
+    #[cfg(target_arch = "aarch64")]
+    impl GICDevice for DummyIrqChip {
+        fn device_properties(&self) -> Vec<u64> {
+            vec![]
+        }
+        fn vcpu_count(&self) -> u64 {
+            0
+        }
+        fn fdt_compatibility(&self) -> String {
+            "vendor,dummy-gic".into()
+        }
+        fn fdt_maint_irq(&self) -> u32 {
+            0
+        }
+        fn version(&self) -> u32 {
+            0
+        }
+    }
+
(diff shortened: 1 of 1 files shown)


src/devices/src/legacy/irqchip.rs
```diff
@@ -166,6 +166,47 @@ pub mod test_utils {
         }
     }
 
+    #[cfg(all(any(test, feature = "test_utils"), target_arch = "aarch64"))]
```
mtjhrc: This impl is wrapped in a mod that is `#[cfg(any(test, feature = "test_utils"))]`, repeating it here is not really necessary. You should just be able to do `#[cfg(target_arch = "aarch64")]`

src/devices/src/legacy/irqchip.rs
```diff
@@ -166,6 +166,47 @@ pub mod test_utils {
...
+        }
+        fn fdt_maint_irq(&self) -> u32 {
+            0
+        }
+        fn version(&self) -> u32 {
+            0
+        }
+    }
+
+    #[cfg(all(any(test, feature = "test_utils"), target_arch = "riscv64"))]
```
mtjhrc: Same here

src/devices/src/legacy/irqchip.rs
peppergrayxyz: fixed!

jakecorrenti:
@mtjhrc what is the GH Actions limitation preventing us from running aarch64 integration tests in CI?

slp:
> @mtjhrc what is the GH Actions limitation preventing us from running aarch64 integration tests in CI?

The GitHub runners for aarch64 doesn't have KVM enabled.

slp (approved):
LGTM, thanks @peppergrayxyz !

Issue #440, opened 2025-10-21
mmio/vstate: unresolved import `devices::legacy::KvmIoapic`


peppergrayxyz:
Build fails on arch64 (https://github.com/containers/libkrun/issues/429):

Definitions of x86_64-only `KvmIoapic` are used unconditionally.

```Rust
error[E0432]: unresolved import `devices::legacy::KvmIoapic`
   --> src/vmm/src/device_manager/kvm/mmio.rs:320:9
    |
320 |         legacy::KvmIoapic,
    |         ^^^^^^^^^^^^^^^^^ no `KvmIoapic` in `legacy`
    |
note: found an item that was configured out
   --> /home/pepper/dev/ghrepos/libkrun2/libkrun/src/devices/src/legacy/mod.rs:65:26
    |
65  | pub use self::kvmioapic::KvmIoapic;
    |                          ^^^^^^^^^
note: the item is gated here
   --> /home/pepper/dev/ghrepos/libkrun2/libkrun/src/devices/src/legacy/mod.rs:64:1
    |
64  | #[cfg(all(target_os = "linux", target_arch = "x86_64"))]
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error[E0432]: unresolved import `devices::legacy::KvmIoapic`
    --> src/vmm/src/linux/vstate.rs:1776:9
     |
1776 |     use devices::legacy::KvmIoapic;
     |         ^^^^^^^^^^^^^^^^^^^^^^^^^^ no `KvmIoapic` in `legacy`
     |
note: found an item that was configured out
    --> /home/pepper/dev/ghrepos/libkrun2/libkrun/src/devices/src/legacy/mod.rs:65:26
     |
65   | pub use self::kvmioapic::KvmIoapic;
     |                          ^^^^^^^^^
note: the item is gated here
    --> /home/pepper/dev/ghrepos/libkrun2/libkrun/src/devices/src/legacy/mod.rs:64:1
     |
64   | #[cfg(all(target_os = "linux", target_arch = "x86_64"))]
     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

slp:
Fixed by #487

PR #444, opened 2025-10-21
Define portable KVM API

peppergrayxyz:
Encapsulate platform specific creation of KVM interrupt controller

Fixes: #440


slp:
Fixed by #487

Issue #441, opened 2025-10-21
vmm: variant or associated item `KernelMmap` not found for this enum


peppergrayxyz:
Build fails on arch64 (https://github.com/containers/libkrun/issues/429):

- `default_guest_memory` is x86_64 only
- `create_guest_memory` signature has changed and the call is broken on aarch64

```Rust
error[E0599]: no variant or associated item named `KernelMmap` found for enum `builder::Payload` in the current scope
    --> src/vmm/src/builder.rs:2186:68
     |
500  | enum Payload {
     | ------------ variant or associated item `KernelMmap` not found for this enum
...
2186 |         create_guest_memory(mem_size_mib, &vm_resources, &Payload::KernelMmap)
     |                                                                    ^^^^^^^^^^ variant or associated item not found in `builder::Payload`

error[E0308]: arguments to this function are incorrect
    --> src/vmm/src/builder.rs:2225:13
     |
2225 |             create_guest_memory(128, None, Payload::Empty).unwrap();
     |             ^^^^^^^^^^^^^^^^^^^            -------------- expected `&Payload`, found `Payload`
     |
note: expected `&VmResources`, found `Option<_>`
    --> src/vmm/src/builder.rs:2225:38
     |
2225 |             create_guest_memory(128, None, Payload::Empty).unwrap();
     |                                      ^^^^
     = note: expected reference `&resources::VmResources`
                     found enum `Option<_>`
note: function defined here
    --> src/vmm/src/builder.rs:1379:4
     |
1379 | fn create_guest_memory(
     |    ^^^^^^^^^^^^^^^^^^^
1380 |     mem_size: usize,
1381 |     vm_resources: &VmResources,
     |     --------------------------
1382 |     payload: &Payload,
     |     -----------------
help: consider borrowing here
     |
2225 |             create_guest_memory(128, None, &Payload::Empty).unwrap();
     |                                            +

error[E0308]: mismatched types
    --> src/vmm/src/builder.rs:2224:13
     |
2224 |         let (guest_memory, _arch_memory_info) =
     |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected a tuple with 4 elements, found one with 2 elements
2225 |             create_guest_memory(128, None, Payload::Empty).unwrap();
     |             ------------------------------------------------------- this expression has type `(vm_memory::GuestMemoryMmap, arch::ArchMemoryInfo, shm::ShmManager, builder::PayloadConfig)`
     |
[...]

PR #445, opened 2025-10-21
fix create_guest_memory on aarch64

peppergrayxyz:
- make default_guest_memory conditonal for x86_64
- update create_guest_memory to current signature
- reorder aarch64 test case to mimic x86_64 test
- fix dead code warning for Payload::None

Fixes: #441

### DIFF
--- a/src/vmm/src/builder.rs
+++ b/src/vmm/src/builder.rs
@@ -504,6 +504,7 @@ enum Payload {
     KernelCopy,
     ExternalKernel(ExternalKernel),
     #[cfg(test)]
+    #[allow(dead_code)]
     Empty,
     Firmware,
     #[cfg(feature = "tee")]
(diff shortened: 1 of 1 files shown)


src/vmm/src/builder.rs
```diff
@@ -2232,6 +2232,12 @@ pub mod tests {
             cpu_template: None,
         };
 
+        let vm_resources = VmResources::default();
+
+        let (guest_memory, _arch_memory_info, _shm_manager, _payload_config) =
+            create_guest_memory(128, &vm_resources, &Payload::Empty).unwrap();
```
slp, line 2238: Thanks for taking some time to fix this!

I think it'd be better changing `default_guest_memory` above to create an empty VmResources (`VmResources::default()`) and creating the memory with an empty payload (`Payload::Empty`) so it can be used both in x86_64 and aarch64 tests.

slp:
Closing this due to lack of activity.

Issue #442, opened 2025-10-21
unittests src/lib.rs: test_regions_gt_1024gb and test_get_fdt_addr fail


peppergrayxyz:
Build fails on arch64 (https://github.com/containers/libkrun/issues/429):

The memory address handling does not match what is expected by the test cases.

```Rust 
     Running unittests src/lib.rs (target/debug/deps/arch-cf6282063f513990)

running 5 tests
test aarch64::tests::test_regions_lt_1024gb ... ok
test aarch64::tests::test_regions_gt_1024gb ... FAILED
test aarch64::linux::regs::tests::test_read_mpidr ... ok
test aarch64::tests::test_get_fdt_addr ... FAILED
test aarch64::linux::regs::tests::test_setup_regs ... ok

failures:

---- aarch64::tests::test_regions_gt_1024gb stdout ----

thread 'aarch64::tests::test_regions_gt_1024gb' panicked at src/arch/src/aarch64/mod.rs:137:9:
assertion `left == right` failed
  left: 1096290402304
 right: 2199023255552
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- aarch64::tests::test_get_fdt_addr stdout ----

thread 'aarch64::tests::test_get_fdt_addr' panicked at src/arch/src/aarch64/mod.rs:152:9:
assertion `left == right` failed
  left: 1073741824
 right: 1073745920


failures:
    aarch64::tests::test_get_fdt_addr
    aarch64::tests::test_regions_gt_1024gb
```

slp:
Fixed by #487

PR #446, opened 2025-10-21
return memory_regions and fdt_addr as expected

peppergrayxyz:
- memory_regions: align up, then cap to the maximum allowed DRAM size
- fdt_addr: put at beginning while <= FDT_MAX_SIZE then move up one page

Fixes: #442


slp:
Fixed by #487

PR #452, opened 2025-10-29
include, block: add `krun_add_disk3` API

jakecorrenti:
Add the `krun_add_disk3` API that allows users to explicitly configure the sync and cache mode for a virtio-blk device. 

This is a followup to https://github.com/containers/libkrun/pull/428

### DIFF
--- a/include/libkrun.h
+++ b/include/libkrun.h
@@ -220,6 +220,60 @@ int32_t krun_add_disk2(uint32_t ctx_id,
                        uint32_t disk_format,
                        bool read_only);
 
+
+/* Supported sync modes */
+
+/**
+ * Ignore VIRTIO_BLK_F_FLUSH.
+ * WARNING: may lead to loss of data 
+ */ 
+#define KRUN_SYNC_NONE 0
+/**
+ * Honor VIRTIO_BLK_F_FLUSH requests, but relax strict hardware syncing on macOS.
+ * This is the recommended mode.
+ *
+ * On macOS this flushes the OS buffers, but does not ask the drive to flush
+ * its buffered data, which significantly improves performance. 
+ * On Linux this is the same as full sync.
+ */
+#define KRUN_SYNC_RELAXED 1
+/** 
+ * Honor VIRTIO_BLK_F_FLUSH, strictly flushing buffers to physical disk.
+ */
+#define KRUN_SYNC_FULL 2
+
+/**
+ * Adds a disk image to be used as a general partition for the microVM.
+ *
+ * This API is mutually exclusive with the deprecated krun_set_root_disk and
+ * krun_set_data_disk methods and must not be used together.
+ *
+ * SECURITY NOTE:
+ * See the security note for `krun_add_disk2`.
+ *
+ * Arguments:
+ *  "ctx_id"      - the configuration context ID.
+ *  "block_id"    - a null-terminated string representing the partition.
+ *  "disk_path"   - a null-terminated string representing the path leading to the disk image.
+ *  "disk_format" - the disk image format (i.e. KRUN_DISK_FORMAT_{RAW, QCOW2})
+ *  "read_only"   - whether the mount should be read-only. Required if the caller does not have
+ *                  write permissions (for disk images in /usr/share).
+ *  "direct_io"   - whether to bypass the host caches.
+ *  "sync_mode"   - whether to enable VIRTIO_BLK_F_FLUSH. On macOS, an additional relaxed sync
+ *                  mode is available, which is enabled by default, and will not ask the drive
+ *                  to flush its buffered data.
+ *
+ * Returns:
+ *  Zero on success or a negative error number on failure.
+ */
+ int32_t krun_add_disk3(uint32_t ctx_id,
+                       const char *block_id,
+                       const char *disk_path,
+                       uint32_t disk_format,
+                       bool read_only,
+                       bool direct_io,
+                       uint32_t sync_mode);
+
 /**
  * NO LONGER SUPPORTED. DO NOT USE.
  *
--- a/src/devices/src/virtio/block/mod.rs
+++ b/src/devices/src/virtio/block/mod.rs
@@ -40,3 +40,43 @@ pub enum ImageType {
     Qcow2,
     Vmdk,
 }
+
+impl TryFrom<u32> for ImageType {
+    type Error = ();
+
+    fn try_from(disk_format: u32) -> Result<Self, Self::Error> {
+        match disk_format {
+            0 => Ok(ImageType::Raw),
+            1 => Ok(ImageType::Qcow2),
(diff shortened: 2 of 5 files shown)


include/libkrun.h
```diff
@@ -1034,6 +1034,27 @@ int32_t krun_add_serial_console_default(uint32_t ctx_id,
  */
 int32_t krun_set_root_disk_remount(uint32_t ctx_id, const char *device, const char *fstype, const char *options);
 
+/**
+ * On macOS, `std::fs::File::flush()` uses `F_FULLFSYNC`, which syncs the buffers to the drive an asks
+ * the drive to wait for flushing all the data to the permanent storage. In contrast, Apple
+ * Virtualization.Framework only issues an `fsync()`, which doesn't ask the drive to flush it's buffered
+ * data.
+ *
+ * Configure whether to match the Virtualization.Framework behavior.
```
mtjhrc:
```suggestion
 * Configure whether to match the relaxed Virtualization.Framework behavior.
```
This could make it a bit more clear

mtjhrc: Also perhaps also mention that the relaxed mode is enabled by default.

jakecorrenti: Good point. I also just realized I accidentally disabled it by default, so I'll fix that :)

jakecorrenti: Now that I'm re-reading your message, do you mean that I should point out how the relaxed behavior is default for Apple or that we have it currently enabled by default

mtjhrc: Both.

"relaxed mode" isn't some established term, so by adding the word relaxed to that sentence made it serve as our definition (and associate it with the previous paragraph explaining the problem).

Separately from that it might be nice to mention what is the default in libkrun.

jakecorrenti: should be all set now

tylerfanelli (commented):
On non macOS targets, should we define the API to just return `EINVAL`?

jakecorrenti:
> On non macOS targets, should we define the API to just return `EINVAL`?

On a non-macOS target it will be a compile-time error. IMO this better than a runtime error so it's clear it's not even an option, but if you think it should be a runtime error I'd love to hear your reasoning

slp:
Hm... I think I would favor a more generic approach, implementing a `krun_add_disk3` instead. This new API should have two additional arguments:

- cache_mode: to choose between "CACHED" and "UNCACHED".
  - CACHED: default behavior, how libkrun works right now.
  - UNCACHED: imago configured with direct_io.
- sync_mode: to choose between "FULL", "RELAXED" and "NONE".
  - FULL: default behavior, how libkrun worked until the relaxed mode on macOS was introduced.
  - RELAXED: on macOS, enable imago relaxed mode. On Linux, we may consider doing a partial fsync.
  - NONE: ignore VIRTIO_BLK_F_FLUSH.

jakecorrenti:
> Hm... I think I would favor a more generic approach, implementing a `krun_add_disk3` instead. This new API should have two additional arguments:
> 
>     * cache_mode: to choose between "CACHED" and "UNCACHED".
>       
>       * CACHED: default behavior, how libkrun works right now.
>       * UNCACHED: imago configured with direct_io.
> 
>     * sync_mode: to choose between "FULL", "RELAXED" and "NONE".
>       
>       * FULL: default behavior, how libkrun worked until the relaxed mode on macOS was introduced.
>       * RELAXED: on macOS, enable imago relaxed mode. On Linux, we may consider doing a partial fsync.
>       * NONE: ignore VIRTIO_BLK_F_FLUSH.

Ah, ok I see now. I'll do that instead. Thanks

src/libkrun/src/lib.rs
```diff
@@ -719,6 +724,72 @@ pub unsafe extern "C" fn krun_add_disk2(
...
+    };
+
+    let format = match disk_format {
+        0 => ImageType::Raw,
+        1 => ImageType::Qcow2,
+        _ => {
+            // Do not continue if the user cannot specify a valid disk format
+            return -libc::EINVAL;
+        }
+    };
```
mtjhrc: Newly added `ImageType::Vmdk` is missing here.

I think this needs to be refactored, copy pasting this around between `krun_add_disk3`,  and `krun_add_disk2` will surely lead to more errors like that. 

I'd suggest doing `impl TryFrom<u32> for ImageType` and use that.

include/libkrun.h
```diff
@@ -220,6 +220,47 @@ int32_t krun_add_disk2(uint32_t ctx_id,
...
                        bool read_only);
 
+
+/* Supported sync modes */
+
+/* Ignore VIRTIO_BLK_F_FLUSH */
+ #define KRUN_SYNC_NONE 0;
+/* MacOS ONLY: do not ask the drive to flush its buffered datta */
+ #define KRUN_SYNC_RELAXED 1;
+ #define KRUN_SYNC_FULL 2;
```
mtjhrc:
```suggestion
/**
 * Ignore VIRTIO_BLK_F_FLUSH. 
 * WARNING: may lead to loss of data 
 **/ 
#define KRUN_SYNC_NONE 0

/**
 * Honor VIRTIO_BLK_F_FLUSH requests, but relax strict hardware syncing on macOS.
 * This is the recommended mode.
 *
 * On macOS this flushes the OS buffers, but does not ask the drive to flush 
 * its buffered data, which significantly improves performance. 
 * On Linux this is the same as full sync.
 **/
#define KRUN_SYNC_RELAXED 1

/** 
 * Honor VIRTIO_BLK_F_FLUSH, strictly flushing buffers to physical disk.
 **/
#define KRUN_SYNC_FULL 2
```

src/libkrun/src/lib.rs
jakecorrenti: Yeah, good catch. I must've missed it when doing a rebase.

jakecorrenti: Should be fixed now.

include/libkrun.h
jakecorrenti: Fixed, thanks!

src/devices/src/virtio/block/mod.rs
```diff
@@ -40,3 +40,43 @@ pub enum ImageType {
     Qcow2,
     Vmdk,
 }
+
+impl TryFrom<u32> for ImageType {
+    type Error = i32;
```
mtjhrc: Maybe this is bit of a nitpick, but I'd rather not spread the usage of plain `i32` for error types outside of the main libkrun `lib.rs` to the "nice" rust code.  

Here you could just put `type Error = ();` and handle returning the correct error at the callsite. 
(I mean you could also define a custom empty Error struct for the error, but that is unnecessary because the failure is obvious. )

jakecorrenti: Fair point. Fixed.

mtjhrc (approved):
Looks great now!

PR #457, opened 2025-11-05
Update nitro enclaves support

tylerfanelli:
The initial phase of AWS nitro enclave support added a few nitro-specific APIs, mainly one to set EIF files. Although simple, this was rather limiting, as containerized applications needed to pre-build and store EIF files for their specific applications within container images. This also likely meant that they could not run these images outside of an enclave.

In this v2 phase of nitro enclave support, we instead provide a stock EIF file as part of libkrun. This EIF file contains the guest kernel+initramfs that the enclave will use. The guest rootfs and execution variables are written from libkrun over vsock to the enclave, which reads these components and uses them to execute a "containerized" process **within the nitro enclave**. This essentially decouples the applications running within an enclave from the enclave itself, and allows containers to run as nitro enclaves with **no modification** to the underlying container images themselves.

On a nitro instance, you can test a simple `cat /etc/os-release` example (`examples/nitro.c`) by providing a rootfs as well as the number of vCPUs and amount of RAM allocated to the enclave.

### DIFF
--- a/init/nitro/archive.c
+++ b/init/nitro/archive.c
@@ -8,8 +8,7 @@
 
 #include "include/archive.h"
 
-static struct archive *
-reader_init(void *buf, size_t size)
+static struct archive *reader_init(void *buf, size_t size)
 {
     struct archive *r;
     int ret;
@@ -35,8 +34,7 @@ reader_init(void *buf, size_t size)
     return r;
 }
 
-static struct archive *
-writer_init()
+static struct archive *writer_init()
 {
     struct archive *w;
     int ret;
@@ -56,8 +54,7 @@ writer_init()
     return w;
 }
 
-static int
-extract(struct archive *r, struct archive *w)
+static int extract(struct archive *r, struct archive *w)
 {
     struct archive_entry *entry;
     const void *buf;
@@ -67,32 +64,38 @@ extract(struct archive *r, struct archive *w)
 
     while ((ret = archive_read_next_header(r, &entry)) != ARCHIVE_EOF) {
         if (ret != ARCHIVE_OK) {
-            printf("error reading archive entry\ncause: %s\n", archive_error_string(r));
+            printf("error reading archive entry\ncause: %s\n",
+                   archive_error_string(r));
             return -1;
         }
 
         ret = archive_write_header(w, entry);
         if (ret != ARCHIVE_OK) {
-            printf("error writing %s entry\ncause: %s\n", archive_entry_pathname(entry), archive_error_string(w));
+            printf("error writing %s entry\ncause: %s\n",
+                   archive_entry_pathname(entry), archive_error_string(w));
             return -1;
         }
 
-        while ((ret = archive_read_data_block(r, &buf, &size, &offset)) != ARCHIVE_EOF) {
+        while ((ret = archive_read_data_block(r, &buf, &size, &offset)) !=
+               ARCHIVE_EOF) {
             if (ret != ARCHIVE_OK) {
-                printf("error reading archive data block\ncause: %s\n", archive_error_string(r));
+                printf("error reading archive data block\ncause: %s\n",
+                       archive_error_string(r));
                 return -1;
             }
 
             ret = archive_write_data_block(w, buf, size, offset);
             if (ret != ARCHIVE_OK) {
-                printf("error writing archive data block\ncause: %s\n", archive_error_string(w));
+                printf("error writing archive data block\ncause: %s\n",
+                       archive_error_string(w));
                 return -1;
             }
         }
 
         ret = archive_write_finish_entry(w);
         if (ret != ARCHIVE_OK) {
-            printf("error finishing %s entry\ncause: %s\n", archive_entry_pathname(entry), archive_error_string(w));
+            printf("error finishing %s entry\ncause: %s\n",
+                   archive_entry_pathname(entry), archive_error_string(w));
             return -1;
         }
     }
@@ -100,8 +103,7 @@ extract(struct archive *r, struct archive *w)
     return 0;
 }
 
-static void
-archive_cleanup(struct archive *r, struct archive *w)
+static void archive_cleanup(struct archive *r, struct archive *w)
 {
     if (r != NULL) {
         archive_read_close(r);
@@ -109,13 +111,12 @@ archive_cleanup(struct archive *r, struct archive *w)
     }
 
     if (w != NULL) {
-	    archive_write_close(w);
+        archive_write_close(w);
         archive_write_free(w);
     }
 }
 
-int
-archive_extract(void *buf, size_t size)
+int archive_extract(void *buf, size_t size)
 {
     struct archive *reader, *writer;
     int ret;
--- a/init/nitro/cgroups_init.c
+++ b/init/nitro/cgroups_init.c
@@ -8,8 +8,7 @@
 
 #include "include/cgroups_init.h"
 
-int
-cgroups_init()
+int cgroups_init()
 {
     const char *fpath = "/proc/cgroups";
     int ret, heir, groups, enabled;
@@ -36,7 +35,7 @@ cgroups_init()
 
         if (ret != 4) {
             fclose(f);
-            errno = errno ? : EINVAL;
+            errno = errno ?: EINVAL;
             perror("fscan /sys/fs/cgroup");
             return -errno;
         }
--- a/init/nitro/fs_init.c
+++ b/init/nitro/fs_init.c
@@ -2,20 +2,19 @@
 
 #include <errno.h>
 #include <stdio.h>
-#include <unistd.h>
 #include <sys/mount.h>
 #include <sys/stat.h>
 #include <sys/types.h>
+#include <unistd.h>
 
 #include "include/fs_init.h"
 
-int
-filesystem_init()
+int filesystem_init()
 {
     int ret;
 
-    ret = mount("proc", "/proc", "proc", MS_NODEV | MS_NOSUID | MS_NOEXEC,
-        NULL);
+    ret =
+        mount("proc", "/proc", "proc", MS_NODEV | MS_NOSUID | MS_NOEXEC, NULL);
     if (ret < 0) {
         perror("mount /proc");
         return -errno;
@@ -46,14 +45,14 @@ filesystem_init()
     }
 
     ret = mount("tmpfs", "/run", "tmpfs", MS_NODEV | MS_NOSUID | MS_NOEXEC,
-        "mode=0755");
+                "mode=0755");
     if (ret < 0) {
         perror("mount /run");
         return -errno;
     }
 
-    ret = mount("tmpfs", "/tmp", "tmpfs", MS_NODEV | MS_NOSUID | MS_NOEXEC,
-        NULL);
+    ret =
+        mount("tmpfs", "/tmp", "tmpfs", MS_NODEV | MS_NOSUID | MS_NOEXEC, NULL);
     if (ret < 0) {
         perror("mount /tmp");
         return -errno;
@@ -66,7 +65,7 @@ filesystem_init()
     }
 
     ret = mount("shm", "/dev/shm", "tmpfs", MS_NODEV | MS_NOSUID | MS_NOEXEC,
-        NULL);
+                NULL);
     if (ret < 0) {
         perror("mount /dev/shm");
         return -errno;
@@ -78,22 +77,21 @@ filesystem_init()
         return -errno;
     }
 
-    ret = mount("devpts", "/dev/pts", "devpts", MS_NOSUID | MS_NOEXEC,
-        NULL);
+    ret = mount("devpts", "/dev/pts", "devpts", MS_NOSUID | MS_NOEXEC, NULL);
     if (ret < 0) {
         perror("mount /dev/pts");
         return -errno;
     }
 
-    ret = mount("sysfs", "/sys", "sysfs", MS_NODEV | MS_NOSUID | MS_NOEXEC,
-        NULL);
+    ret =
+        mount("sysfs", "/sys", "sysfs", MS_NODEV | MS_NOSUID | MS_NOEXEC, NULL);
     if (ret < 0) {
         perror("mount /sys");
         return -errno;
     }
 
     ret = mount("cgroup_root", "/sys/fs/cgroup", "tmpfs",
-                    MS_NODEV | MS_NOSUID | MS_NOEXEC, "mode=0755");
+                MS_NODEV | MS_NOSUID | MS_NOEXEC, "mode=0755");
     if (ret < 0) {
         perror("mount /sys/fs/cgroup");
         return -errno;
--- a/init/nitro/main.c
+++ b/init/nitro/main.c
@@ -3,39 +3,40 @@
 #define _GNU_SOURCE
 #include <errno.h>
 #include <fcntl.h>
+#include <linux/vm_sockets.h>
 #include <signal.h>
 #include <stddef.h>
 #include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
-#include <unistd.h>
 #include <sys/mount.h>
 #include <sys/reboot.h>
 #include <sys/socket.h>
 #include <sys/stat.h>
-#include <sys/types.h>
 #include <sys/syscall.h>
+#include <sys/types.h>
 #include <sys/wait.h>
-#include <linux/vm_sockets.h>
+#include <unistd.h>
 
 #include <nsm.h>
 
 #include "include/archive.h"
-#include "include/fs_init.h"
 #include "include/cgroups_init.h"
+#include "include/fs_init.h"
 #include "include/vsock.h"
 
-#define finit_module(fd, param_values, flags) (int)syscall(__NR_finit_module, fd, param_values, flags)
+#define finit_module(fd, param_values, flags)                                  \
+    (int)syscall(__NR_finit_module, fd, param_values, flags)
 
 #define HEART_BEAT 0xb7
 #define VSOCK_CID 3
 #define VSOCK_PORT 9000
 
-#define NSM_PCR_ROOTFS      16
-#define NSM_PCR_EXEC_DATA   17
+#define NSM_PCR_ROOTFS 16
+#define NSM_PCR_EXEC_DATA 17
 
-#define NSM_PCR_CHUNK_SIZE   0x800  // 2 KiB.
+#define NSM_PCR_CHUNK_SIZE 0x800 // 2 KiB.
 
 /*
  * Block or unblock signals.
@@ -44,8 +45,7 @@
  *       Therefore, perror output may not be displayed if a failure occurs
  *       during this setup.
  */
-int
-sig_mask(int mask)
+int sig_mask(int mask)
 {
     sigset_t set;
     int ret;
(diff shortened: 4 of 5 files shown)


init/nitro/main.c
```diff
@@ -0,0 +1,231 @@
...
+ * Initialize a devtmpfs at /dev.
+ */
+int
+dev_init()
+{
+    int ret;
+
+    ret = mount("dev", "/dev", "devtmpfs", MS_NOSUID | MS_NOEXEC, NULL);
+    if (ret < 0) {
+        perror("mount");
```
jakecorrenti: IMO it would be beneficial to have more descriptive error messages; maybe something along the lines of "failed to mount /dev as devtmpfs" for this specific instance.

init/nitro/main.c
```diff
@@ -0,0 +1,231 @@
...
+console_init()
+{
+    const char *path = "/dev/console";
+    FILE *file;
+    int ret;
+
+    // Initialize /dev.
+    ret = dev_init();
+    if (ret < 0)
+        return ret;
```
jakecorrenti: I don't think mounting `/dev` needs to be its own function since it's essentially just wrapping `mount` and adding a `perror()`
```suggestion
    // Initialize a devtmpfs at /dev
    ret = mount("dev", "/dev", "devtmpfs", MS_NOSUID | MS_NOEXEC, NULL);
    if (ret < 0) {
        perror("failed to mount /dev as devtmpfs");
        return ret;
    }
```

init/nitro/main.c
```diff
@@ -0,0 +1,231 @@
...
+    if (file == NULL)
+        return -1;
+
+    file = freopen(path, "w", stdout);
+    if (file == NULL)
+        return -1;
+
+    file = freopen(path, "w", stderr);
+    if (file == NULL)
+        return -1;
```
jakecorrenti: We should return `-errno` and add `perror("failed to redirect <stream> to /dev/console")` so the user can get some notion of what went wrong. (like you did below in `nsm_init()`)

init/nitro/main.c
```diff
@@ -0,0 +1,231 @@
...
+    }
+
+    ret = read(sock_fd, buf, 1);
+    if (ret != 1) {
+        perror("vsock read");
+        return -errno;
+    }
+
+    if (buf[0] != HEART_BEAT)
+        return -1;
```
jakecorrenti:
```suggestion
    if (buf[0] != HEART_BEAT) {
        puts("failed to establish connection");
        return -1;
    }
```

src/libkrun/src/lib.rs
```diff
@@ -373,10 +373,29 @@ impl TryFrom<ContextConfig> for NitroEnclave {
...
+
+            let path = PathBuf::from(&fs.shared_dir);
+
+            if !path.as_path().is_dir() {
+                error!("rootfs path {:?} is not a valid directory", path.display());
+                return Err(-libc::EINVAL);
+            }
+
+            path
+        };
```
jakecorrenti: You could also write this such that you're not directly indexing the vec; up to you.
```suggestion
    let rootfs = if let Some(fs) = &ctx.vmr.fs.first() {
            let path = PathBuf::from(&fs.shared_dir);
            if !path.as_path().is_dir() {
                error!("rootfs path {:?} is not a valid directory", path.display());
                return Err(-libc::EINVAL);
            }
            path
    } else {
        error!("rootfs path required");
        return Err(-libc::EINVAL);
    };
```

src/nitro/src/enclaves.rs
```diff
@@ -52,61 +69,59 @@ impl NitroEnclave {
...
 
-        builder
-            .append_dir_all("rootfs", self.rootfs.clone())
-            .unwrap();
+        if result != Ok(1) {
+            return Err(NitroError::PollMoreThanOneSelectedEvent);
+        }
 
-        Ok(builder.into_inner().unwrap())
+        Ok(())
```
jakecorrenti: Instead of doing the individual checks on the result values you can do a match with a guard
```rust
        match result {
            Ok(0) => Err(NitroError::PollNoSelectedEvents),
            Ok(x) if x > 1 => Err(NitroError::PollMoreThanOneSelectedEvent),
            _ => Ok(())
        }
```

init/nitro/main.c
```diff
@@ -177,32 +181,118 @@ krun_signal()
...
+    read_len = read(sock_fd, len_b, 8);
+    if (read_len < 8) {
+        close(sock_fd);
+        perror("vsock rootfs archive length read");
         return -1;
+    }
+
+    memcpy(&len, len_b, sizeof(uint64_t));
+
+    *size = len;
```
jakecorrenti: need to check if size is NULL first?

src/nitro/src/enclaves.rs
```diff
@@ -51,10 +53,10 @@ impl NitroEnclave {
         let mut builder = tar::Builder::new(Vec::new());
 
         builder
-            .append_dir_all("rootfs", self.rootfs.clone())
-            .unwrap();
+            .append_dir("rootfs", self.rootfs.clone())
```
jakecorrenti: why do we not want a recursive append anymore?

init/nitro/cgroups_init.c
```diff
@@ -0,0 +1,66 @@
...
+        perror("fopen /proc/cgroups");
+        return -errno;
+    }
+
+    // Skip the first line.
+    for (;;) {
+        ret = fgetc(f);
+        if (ret == EOF || ret == '\n')
+            break;
+    }
```
jakecorrenti:
```suggestion
    while (ret != EOF && ret != '\n')
        ret = fgetc(f);
```

init/nitro/cgroups_init.c
```diff
@@ -0,0 +1,66 @@
...
+        static const char base_path[] = "/sys/fs/cgroup/";
+        char path[sizeof(base_path) - 1 + 64];
+        char *name = path + sizeof(base_path) - 1;
+        ret = fscanf(f, "%64s %d %d %d\n", name, &heir, &groups, &enabled);
+        if (ret == EOF)
+            break;
+
+        if (ret != 4) {
+            errno = errno ? : EINVAL;
+            perror("fscan /sys/fs/cgroup");
```
jakecorrenti, line 39: close file pointer

init/nitro/cgroups_init.c
```diff
@@ -0,0 +1,66 @@
...
+            return -errno;
+        }
+
+        if (enabled) {
+            memcpy(path, base_path, sizeof(base_path) - 1);
+
+            ret = mkdir(path, 0755);
+            if (ret < 0) {
+                perror("mkdir cgroup path");
```
jakecorrenti, line 49: close file pointer here as well

init/nitro/cgroups_init.c
```diff
@@ -0,0 +1,66 @@
...
+                return -errno;
+            }
+
+            ret = mount(name, path, "cgroup", MS_NODEV | MS_NOSUID | MS_NOEXEC,
+                        name);
+            if (ret < 0) {
+                perror("mount cgroup");
+                return -errno;
```
jakecorrenti, line 58: close file pointer here too

init/nitro/fs_init.c
```diff
@@ -0,0 +1,140 @@
...
+
+    return 0;
+}
+
+int
+filesystem_init()
+{
+    int ret;
+
+    ret = dir_mount("proc", "/proc", "proc", MS_NODEV | MS_NOSUID | MS_NOEXEC,
```
jakecorrenti: just call `mount()` instead of wrapping it. There's functionally no difference / readability difference

init/nitro/fs_init.c
```diff
@@ -0,0 +1,140 @@
...
+                    NULL);
+    if (ret < 0) {
+        perror("mount /proc");
+        return -errno;
+    }
+
+    ret = symlink_add("/dev/fd", "/proc/self/fd");
```
jakecorrenti: we don't need to wrap `symlink` either

init/nitro/main.c
```diff
@@ -340,6 +349,10 @@ main(int argc, char *argv[])
     if (ret < 0)
         exit(ret);
 
+    ret = launch(exec_argv, exec_envp);
+    if (ret < 0)
+        exit(ret);
+
     exit(0);
```
jakecorrenti, line 428: `exit(ret)` should be sufficient

init/nitro/main.c
```diff
@@ -285,6 +292,73 @@ launch(char **argv, char **envp)
...
+    uint32_t pcr_data_size;
+    uint8_t pcr_data[NSM_PCR_DATA_SIZE];
+    bool locked;
+    int ret;
+
+    pcr_data_size = NSM_PCR_DATA_SIZE;
+
+    ret = nsm_describe_pcr(nsm_fd, pcr, &locked, pcr_data, &pcr_data_size);
+    if (ret != ERROR_CODE_SUCCESS)
+        return -ret;
```
jakecorrenti: Why are you describing the PCR state before extending it? You already have all the information needed to just call `nsm_extend_pcr()`; check the function's docs in `/usr/include/nsm.h`.

init/nitro/main.c
tylerfanelli: I don't think so. `size` is allocated on the stack from the caller. If this was a library meant to be used by other consumers, sure.

init/nitro/cgroups_init.c
tylerfanelli: I don't think that would be correct because `ret` isn't initialized to anything. Perhaps
```
do {
    ret = fgetc(f);
} while (ret != EOF && ret != '\n');
```
Would be a valid alternative.

init/nitro/main.c
tylerfanelli: `exit(0)` is fine. It indicates success. `ret` could be some other positive value which could be interpreted as an error code.

jakecorrenti: after `launch()`, `ret` will always be a non-positive number. I'm fine with keeping it as-is, though.

init/nitro/main.c
```diff
@@ -0,0 +1,241 @@
...
+    }
+
+    return 0;
+}
+
+/*
+ * Initialize a devtmpfs at /dev.
+ */
+int
+dev_init()
```
jakecorrenti: can be removed in this commit.

init/nitro/main.c
```diff
@@ -0,0 +1,241 @@
...
+    sock_fd = socket(AF_VSOCK, SOCK_STREAM, 0);
+    if (sock_fd < 0) {
+        perror("vsock initialization");
+        return -errno;
+    }
+
+    // Connect to the host.
+    ret = connect(sock_fd, (struct sockaddr *) &saddr, sizeof(saddr));
+    if (ret < 0) {
+        perror("vsock connect");
```
jakecorrenti, line 176: close socket fd

init/nitro/main.c
```diff
@@ -0,0 +1,241 @@
...
+        return -errno;
+    }
+
+    // Write the heartbeat to the host and read it back to ensure that the
+    // communication is established.
+    ret = write(sock_fd, buf, 1);
+    if (ret != 1) {
+        perror("vsock write");
```
jakecorrenti: close socket fd

init/nitro/main.c
```diff
@@ -0,0 +1,241 @@
...
+        return -errno;
+    }
+
+    ret = read(sock_fd, buf, 1);
+    if (ret != 1) {
+        perror("vsock read");
```
jakecorrenti: close socket fd

init/nitro/main.c
```diff
@@ -184,12 +188,14 @@ krun_signal()
     // communication is established.
     ret = write(sock_fd, buf, 1);
     if (ret != 1) {
+        close(sock_fd);
```
jakecorrenti, line 184: this and the addition on line 198 should be in the first commit

init/nitro/archive.c
```diff
@@ -0,0 +1,134 @@
...
+        return -1;
+
+    ret = extract(reader, writer);
+    if (ret < 0)
+        return -1;
+
+    archive_cleanup(reader, writer);
+
+    return 0;
+}
```
jakecorrenti, line 138: We should also cleanup on a failed extraction, I believe
```suggestion
    ret = extract(reader, writer);
    archive_cleanup(reader, writer);
    return ret;
}
```

init/nitro/vsock.c
```diff
@@ -0,0 +1,103 @@
...
+    buf = (char **) malloc(sizeof(char *) * (size + 1));
+    if (buf == NULL) {
+        close(sock_fd);
+        perror("vsock char list buffer malloc");
+        return -errno;
+    }
+
+    for (i = 0; i < size; ++i) {
+        ret = vsock_rcv(sock_fd, (void **) &buf[i], NULL);
+        if (ret < 0)
```
jakecorrenti: need to free memory allocated for `buf`

init/nitro/vsock.c
```diff
@@ -0,0 +1,103 @@
...
+
+    size = 0;
+    ret = vsock_len_read(sock_fd, &size);
+    if (ret < 0)
+        return ret;
+
+    // Allocate extra space for NULL-terminator.
+    buf = (char **) malloc(sizeof(char *) * (size + 1));
+    if (buf == NULL) {
+        close(sock_fd);
```
jakecorrenti: IMO we should delegate freeing socket resources back to the caller here since we didn't allocate that memory within this function. At the very least the behavior should be documented to avoid possible double-frees in the future.

init/nitro/vsock.c
```diff
@@ -0,0 +1,103 @@
...
+    ret = vsock_len_read(sock_fd, &len);
+    if (ret < 0)
+        return ret;
+
+    if (size != NULL)
+        *size = len;
+
+    buf = (uint8_t *) malloc(sizeof(uint8_t) * len);
+    if (buf == NULL) {
+        close(sock_fd);
```
jakecorrenti: same comment here w/r/t socket resources being freed

init/nitro/vsock.c
```diff
@@ -0,0 +1,103 @@
...
+        perror("vsock byte buffer malloc");
+        return -errno;
+    }
+
+    while (len) {
+        read_len = read(sock_fd, &buf[idx], len);
+        if (read_len <= 0) {
+            close(sock_fd);
```
jakecorrenti: same comment here w/r/t socket resources being freed

src/nitro/src/enclaves.rs
```diff
@@ -1,97 +1,279 @@
...
+
+        let rootfs_dirname: String = {
+            let split: Vec<&str> = self.rootfs.split("/").collect();
+
+            split.last().unwrap_or(&"/").to_string()
+        };
+
+        for entry in
+            fs::read_dir(PathBuf::from(self.rootfs.clone())).map_err(NitroError::RootFsArchive)?
+        {
```
jakecorrenti: I think you can do something like
```suggestion
        let pathbuf = PathBuf::from(self.rootfs.clone());
        let rootfs_dirname = pathbuf.file_name().unwrap_or(&OsStr::new("/")).to_str().unwrap();
        
        for entry in
            fs::read_dir(pathbuf).map_err(NitroError::RootFsArchive)?
        {
```

src/nitro/src/enclaves.rs
```diff
@@ -32,29 +41,69 @@ pub struct NitroEnclave {
...
+                    "unable to convert file name to String object",
+                )));
+            };
+            let split: Vec<&str> = string.split("/").collect();
+            let Some(last) = split.last() else {
+                return Err(NitroError::RootFsArchive(io::Error::new(
+                    ErrorKind::Other,
+                    "unable to parse file path leaf name",
+                )));
+            };
```
jakecorrenti: Aren't `filename` and `last` always going to be the same value? `file_name()` returns the leaf not a full path

src/nitro/src/enclaves.rs
```diff
@@ -32,29 +41,69 @@ pub struct NitroEnclave {
...
+
+            if !ROOTFS_DIR_DENYLIST.contains(last) && last != &rootfs_dirname {
```
jakecorrenti: I don't think it's possible for an entry to be the parent itself

src/nitro/src/enclaves.rs
```diff
@@ -32,29 +41,69 @@ pub struct NitroEnclave {
...
+                } else if filetype.is_file() {
+                    builder
+                        .append_path_with_name(entry.path(), format!("rootfs/{}", last))
+                        .map_err(NitroError::RootFsArchive)?;
+                }
+            }
+        }
 
-        Ok(builder.into_inner().unwrap())
+        builder.into_inner().map_err(NitroError::RootFsArchive)
```
jakecorrenti, line 113: Should probably be in a previous commit

jakecorrenti:
was there a problem with the PCR code where you removed it for this revision?

src/nitro/src/enclaves.rs
```diff
@@ -1,97 +1,279 @@
...
+        {
+            let entry = entry.map_err(NitroError::RootFsArchive)?;
+            let filetype = entry.file_type().map_err(NitroError::RootFsArchive)?;
+            let filename = entry.file_name();
+            let Ok(string) = filename.clone().into_string() else {
+                return Err(NitroError::RootFsArchive(io::Error::new(
+                    ErrorKind::Other,
+                    "unable to convert file name to String object",
+                )));
+            };
```
jakecorrenti:
```suggestion
            let filename = entry.file_name().into_string().map_err(|_| {
                NitroError::RootFsArchive(io::Error::new(
                    ErrorKind::Other,
                    "unable to convert file name to String object",
                ))
            })?;
```

src/nitro/src/enclaves.rs
tylerfanelli: The top-level directory is included in the archiving. We are essentially renaming the top-level directory to `rootfs`, so we ignore that listing.

src/nitro/src/enclaves.rs
tylerfanelli: Yes, thanks for pointing that out.

init/nitro/vsock.c
tylerfanelli: Thanks. See https://github.com/containers/libkrun/pull/457/commits/667dacaf987d9fdb9a48af9928a68e34b4ecdb50.

init/nitro/archive.c
```diff
@@ -0,0 +1,131 @@
...
+    struct archive *reader, *writer;
+    int ret;
+
+    reader = reader_init(buf, size);
+    if (reader == NULL)
+        return -1;
+
+    writer = writer_init();
+    if (writer == NULL)
+        return -1;
```
jakecorrenti, line 131: You're leaking memory for the readers and writers if `reader_init` or `writer_init` fails right after it successfully allocates the memory on the heap.

init/nitro/main.c
```diff
@@ -0,0 +1,438 @@
...
+    if (fd < 0 && errno == ENOENT)
+        return 0;
+    else if (fd < 0) {
+        perror("nsm.ko open");
+        return -errno;
+    }
+
+    // Load the NSM module.
+    ret = finit_module(fd, "", 0);
+    if (ret < 0) {
```
jakecorrenti, line 127: missing an `fclose` here

jakecorrenti:
Just need to see that formatting workflow go green as well

init/nitro/archive.c
tylerfanelli: Not the writer, but only the reader if `writer_init` fails. I've made the necessary change.

PR #461, opened 2025-11-11
Add cross-compilation support to Makefile

nohajc:
At the moment, I'm porting `init.krun` to FreeBSD (hoping to open that PR soon) and I thought it would be neat to have the option to compile on macOS without spinning up a VM if one so chooses.

This relies on `clang` and `lld` (the LLVM linker). Plus, you need to provide path to a suitable sysroot. I tested the build by copying `/usr/include`, `/usr/lib/aarch64-linux-gnu` and `/usr/lib/gcc/aarch64-linux-gnu` from a Debian Bookworm system.

Just wasn't sure if I should strip the binary or not. Right now it's configured to only strip `debug_info`.


nohajc:
I suppose I should also document this in README if the change is accepted.

Makefile
```diff
@@ -100,10 +100,18 @@ all: $(LIBRARY_RELEASE_$(OS)) libkrun.pc
 
 debug: $(LIBRARY_DEBUG_$(OS)) libkrun.pc
 
+ifeq ($(SYSROOT_LINUX),)
+# Build on Linux host
+CC_LINUX=$(CC)
+else
+# Cross-compile on macOS with the LLVM linker (brew install lld)
+CC_LINUX=clang -target aarch64-linux-gnu -fuse-ld=lld -Wl,-strip-debug --sysroot $(SYSROOT_LINUX) -Wno-c23-extensions
```
nohajc: Oh, just realized the target triple cannot be hardcoded.

Makefile
```diff
@@ -73,6 +73,7 @@ ifeq ($(TIMESYNC),1)
 endif
 
 OS = $(shell uname -s)
+ARCH = $(shell uname -m)
```
nohajc, line 85: This can return `arm64` on macOS or `aarch64` on Linux but clang seems happy with either in the target triple.

slp:
Having to gather all binaries for putting together a sysroot is a bit cumbersome. I was wondering if we could, perhaps, provide a Containerfile that would build `init/init` on macOS Podman/Docker. I think that would simplify a lot the process for the users.

nohajc:
My idea was to minimize build dependencies (i.e. it should be doable without any container runtime / virtualization solution). I first added this for FreeBSD on my branch where making a sysroot was as simple as downloading and unpacking base.txz.

Maybe I could write a similar script for Linux. For example, it's possible to use skopeo and umoci to download and extract files from any OCI image. But I'm sure one could find Debian tarballs or something like that as well.

nohajc:
Well, turns out it's not easy to find a ready-made Linux rootfs with header files and the necessary libs. I know Chromium generates archives like that but the one I tried before didn't work.

nohajc:
Yep, you can just download these four deb packages, unpack with `ar` and `tar` and you have a working 40MB sysroot:

```
https://deb.debian.org/debian/pool/main/g/glibc/libc6_2.42-2_arm64.deb
https://deb.debian.org/debian/pool/main/g/glibc/libc6-dev_2.42-2_arm64.deb
https://deb.debian.org/debian/pool/main/g/gcc-12/libgcc-12-dev_12.5.0-6_arm64.deb
https://deb.debian.org/debian/pool/main/l/linux/linux-libc-dev_6.12.57-1_all.deb
```

nohajc:
Ok, I updated the Makefile to generate a Debian sysroot automatically on macOS.
It downloads `https://deb.debian.org/debian/dists/$(DEBIAN_DIST)/main/binary-$(ARCH)/Packages.xz` and parses it to find the following packages:

```
libc6, libc6-dev, libgcc-12-dev, linux-libc-dev
```

These are downloaded and unpacked to `linux-sysroot/`. The only homebrew dependencies are `lld` and `xz`.

This should ultimately simplify homebrew packaging of libkrun as well.

Makefile
```diff
@@ -109,10 +110,18 @@ all: $(LIBRARY_RELEASE_$(OS)) libkrun.pc
 
 debug: $(LIBRARY_DEBUG_$(OS)) libkrun.pc
 
+ifeq ($(SYSROOT_LINUX),)
+# Build on Linux host
+CC_LINUX=$(CC)
+else
+# Cross-compile on macOS with the LLVM linker (brew install lld)
+CC_LINUX=clang -target $(ARCH)-linux-gnu -fuse-ld=lld -Wl,-strip-debug --sysroot $(SYSROOT_LINUX) -Wno-c23-extensions
```
slp: Please use `/usr/bin/clang` instead of `clang` here. Homebrew puts a wrapper in PATH that messes up with the command line, so we need to avoid by using an absolute path.

slp: Also, please use the same indentation we have in other conditionals above.

Makefile
slp, line 1: Please squash this patch inside the first one.

Makefile
nohajc: > Homebrew puts a wrapper in PATH that messes up with the command line

Oh, you're right, I forgot I've already seen this issue before.

nohajc:
@slp Thanks for the review! I applied your suggestions.

nohajc:
Tbh, I don't like this kind of indentation where you're mixing spaces and tabs (and tabs have special meaning).
I just had to debug why my variables were not set properly and it was because they cannot be indented by a tab.

If you open the Makefile in VS Code, syntax highlighting actually reveals this. Also, it's already broken in couple of other places where a tab has been used by accident.

slp:
> Tbh, I don't like this kind of indentation where you're mixing spaces and tabs (and tabs have special meaning). I just had to debug why my variables were not set properly and it was because they cannot be indented by a tab.
> 
> If you open the Makefile in VS Code, syntax highlighting actually reveals this. Also, it's already broken in couple of other places where a tab has been used by accident.

Agreed, we should clean it up eventually.

PR #467, opened 2025-11-14
init: set MS_SHARED for the root mount

valpackett:
systemd expects container runtimes to do this[1], otherwise various things (such as credential passing) [break](https://github.com/systemd/systemd/issues/39501). It's generally the reasonable default and other container software might expect it to already be set as well.

[1]: https://systemd.io/CONTAINER_INTERFACE/

### DIFF
--- a/init/init.c
+++ b/init/init.c
@@ -1133,6 +1133,11 @@ int main(int argc, char **argv)
         }
     }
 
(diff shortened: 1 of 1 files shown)


jakecorrenti (approved):
Thank you for the contribution, @valpackett! 

Code LGTM, but could you rebase and re-push?

valpackett:
Sure, done!

PR #470, opened 2025-11-21
init/nitro: Avoid measuring `/etc/{hostname, hosts}` with NSM PCRs

tylerfanelli:
Avoid measuring the `/etc/{hostname, hosts}` files, as their values are influenced by the container ID when running containers. The container ID is an ephemeral value across reboots, and thus produces differing measurements for otherwise identical container root filesystems.

### DIFF
--- a/init/nitro/archive.c
+++ b/init/nitro/archive.c
@@ -168,7 +168,7 @@ static int extract(int nsm_fd, struct archive *r, struct archive *w)
     return 0;
 
 err:
-    return -ret;
+    return ret;
 }
 
 /*


init/nitro/archive.c
```diff
@@ -106,6 +107,19 @@ static int extract(int nsm_fd, struct archive *r, struct archive *w)
...
+         * Avoid measuring /etc/hostname, which is an ephemeral value in
+         * containers (the hostname being the container ID).
+         */
+        measure &= (strstr(path, "rootfs/etc/hostname") == NULL);
+
+        /*
+         * Also avoid measuring /etc/hosts, which also contains the ephemeral
+         * hostname.
+         */
+        measure &= (strstr(path, "rootfs/etc/hosts") == NULL);
```
jakecorrenti, line 122: just curious, why'd you decide to do a bitwise operation here? would `measure = (strstr(path, <other path>) == NULL)` not be the same without the extra operation?

jakecorrenti (approved):

init/nitro/archive.c
tylerfanelli: This would turn measure to false if the path is `rootfs/etc/hostname` OR `rootfs/etc/hosts`.

If path is `rootfs/etc/hosts`:

`measure = true`
`measure &= (strstr(path, "rootfs/etc/hostname") == NULL)` == `true & true` == `true`
`measure &= (strstr(path, "rootfs/etc/hostname") == NULL)` == `true & false` == `false`

PR #474, opened 2025-11-28
Add support for readonly FLAT/ZERO VMDK format

hsiangkao:
Also see: https://github.com/containerd/nerdbox/issues/30

Currently, it's mainly used for EROFS to merge hundreds of sub-blobs (container image layers) into one block device to avoid having too many block devices (and it would even be impossible for virtio-mmio since legacy IRQs is much limited in libkrun).

Why VMDK is useful?  Since it seems to be the only standard and simple way to support one-single block device that consists of a collection of multiple file parts among popular virtualization products such as QEMU and VirtualBox.

Update the `krun_add_disk2` API to specify the VMDK format:
 - KRUN_DISK_FORMAT_VMDK

### DIFF
--- a/include/libkrun.h
+++ b/include/libkrun.h
@@ -167,6 +167,9 @@ int32_t krun_add_disk(uint32_t ctx_id, const char *block_id, const char *disk_pa
 /* Supported disk image formats */
 #define KRUN_DISK_FORMAT_RAW 0
 #define KRUN_DISK_FORMAT_QCOW2 1
+/* Note: Only supports FLAT/ZERO formats without delta links */
+#define KRUN_DISK_FORMAT_VMDK 2
+
 /**
  * Adds a disk image to be used as a general partition for the microVM. The supported
  * image formats are: "raw" and "qcow2".
--- a/src/devices/Cargo.toml
+++ b/src/devices/Cargo.toml
@@ -38,7 +38,7 @@ arch = { path = "../arch" }
 utils = { path = "../utils" }
 polly = { path = "../polly" }
 rutabaga_gfx = { path = "../rutabaga_gfx", features = ["virgl_renderer", "virgl_renderer_next"], optional = true }
-imago = { version = "0.1.6", features = ["sync-wrappers", "vm-memory"] }
+imago = { version = "0.2.1", features = ["sync-wrappers", "vm-memory"] }
 
 [target.'cfg(target_os = "macos")'.dependencies]
 hvf = { path = "../hvf" }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -165,6 +165,25 @@ dependencies = [
 ]
 
 [[package]]
+name = "bincode"
+version = "2.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "36eaf5d7b090263e8150820482d5d93cd964a81e4019913c972f4edcc6edb740"
+dependencies = [
+ "bincode_derive",
+ "unty",
+]
+
+[[package]]
+name = "bincode_derive"
+version = "2.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bf95709a440f45e986983918d0e8a1f30a9b1df04918fc828670606804ac3c09"
+dependencies = [
+ "virtue",
+]
+
+[[package]]
 name = "bindgen"
 version = "0.69.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -739,23 +758,22 @@ dependencies = [
 
 [[package]]
 name = "imago"
-version = "0.1.6"
+version = "0.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a8e45dce78f8a4700ea570488b35af0d192ee0dd355a1d625f07f6ea49ac25bd"
+checksum = "5a46885ecbabe024b9e6e3ee9ee3a32ad05adf2b6269f45aa912793d9c54c805"
 dependencies = [
  "async-trait",
- "bincode",
+ "bincode 2.0.1",
  "cfg-if",
  "libc",
  "miniz_oxide",
- "nix 0.29.0",
+ "nix 0.30.1",
  "page_size",
  "rustc_version",
- "serde",
  "tokio",
  "tracing",
  "vm-memory",
- "windows-sys 0.59.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -1588,7 +1606,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1750ba11a6a6bba3c220da714caa0226aa34e417dce3975d2953062240717dea"
 dependencies = [
  "base64",
- "bincode",
+ "bincode 1.3.3",
  "bitfield",
  "bitflags 2.9.1",
  "byteorder",
@@ -1884,6 +1902,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7dd6e30e90baa6f72411720665d41d89b9a3d039dc45b8faea1ddd07f617f6af"
 
 [[package]]
+name = "unty"
+version = "0.0.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6d49784317cd0d1ee7ec5c716dd598ec5b4483ea832a2dced265471cc0f690ae"
+
+[[package]]
 name = "utf8parse"
 version = "0.2.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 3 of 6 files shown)


include/libkrun.h
```diff
@@ -167,6 +167,9 @@ int32_t krun_add_disk(uint32_t ctx_id, const char *block_id, const char *disk_pa
 /* Supported disk image formats */
 #define KRUN_DISK_FORMAT_RAW 0
 #define KRUN_DISK_FORMAT_QCOW2 1
+/* It's used only when you know what you're doing */
+#define KRUN_DISK_FORMAT_VMDK 2
```
mtjhrc: Could you revise this comment? The current wording comes across as dismissive and doesn’t describe the actual constraints. If there are specific risks or limitations that require 'knowing what you're doing', please document them briefly —e.g., `/* Note: Only supports ____ */ `

hsiangkao: Yeah, make sense. I revise it as `/* Note: Only supports FLAT/ZERO formats without delta links */`

src/devices/Cargo.toml
```diff
@@ -38,7 +38,7 @@ arch = { path = "../arch" }
 utils = { path = "../utils" }
 polly = { path = "../polly" }
 rutabaga_gfx = { path = "../rutabaga_gfx", features = ["virgl_renderer", "virgl_renderer_next"], optional = true }
-imago = { version = "0.1.5", features = ["sync-wrappers", "vm-memory"] }
+imago = { features = ["sync-wrappers", "vm-memory"], git = "https://gitlab.com/hreitz/imago.git", branch = "main" }
```
slp: We can only consume dependencies published on crates.io. The stable branch of imago doesn't have vmdk support, so we need to wait a bit for `0.2.0` to be released.

hsiangkao: Currently nerdbox will directly rely on this but libkrun doesn't support virtio-pci so the total number of virtio-blk device is much limited (even support it still ineffencient compared to this way.

BTW, vmdk support is much seperate from the main raw and qcow2 support, I wonder if it might be fine for 0.1.x? or is there some timeline to release 0.2.0? Cc @xanClic

XanClic: I’m working on 0.2.0, it should be doable by next week. If I can’t get it done, I’ll make at least an 0.1.7 with it.

hsiangkao: That is awesome, Thank you all!

XanClic: I’ve just released 0.2.0. As far as I’m aware the only breaking change is that some enums are now marked `#[non_exhaustive]`, but I don’t think libkrun uses any of them.

hsiangkao: > I’ve just released 0.2.0. As far as I’m aware the only breaking change is that some enums are now marked `#[non_exhaustive]`, but I don’t think libkrun uses any of them.

Hey @XanClic, many thanks, I will use update this commit soon, but I checked the previous discussion: it might need to wait for fedora packaging to be landed first?

XanClic: Only @slp can say, but I’ll need to provide a Fedora update anyway.

I’ll definitely push it to Rawhide. I’m not so sure about the stable releases; it is technically a breaking update, so shouldn’t be done, but given that libkrun is the only user I know of and that this MR effectively confirms there is no breaking change for libkrun, it should be fine, I suppose…

hsiangkao: > Only @slp can say, but I’ll need to provide a Fedora update anyway.
> 
> I’ll definitely push it to Rawhide. I’m not so sure about the stable releases; it is technically a breaking update, so shouldn’t be done, but given that libkrun is the only user I know of and that this MR effectively confirms there is no breaking change for libkrun, it should be fine, I suppose…

I noticed imago just upgraded vm-memory to 0.18, which is incompatible with the current libkrun 0.16 one:

```
error[E0277]: the trait bound `VolatileSlice<'_>: imago::misc_helpers::ImagoAsRef<'_, vm_memory::volatile_memory::VolatileSlice<'_, _>>` is not satisfied
    --> src/devices/src/virtio/file_traits.rs:430:64
     |
 430 |         let (iovec, _guard) = IoVectorMut::from_volatile_slice(bufs);
     |                               -------------------------------- ^^^^ the trait `imago::misc_helpers::ImagoAsRef<'_, vm_memory::volatile_memory::VolatileSlice<'_
, _>>` is not implemented for `VolatileSlice<'_>`
     |                               |
     |                               required by a bound introduced by this call
     |
     = help: the following other types implement trait `imago::misc_helpers::ImagoAsRef<'a, T>`:
               `&'a U` implements `imago::misc_helpers::ImagoAsRef<'a, T>`
               `&'a vm_memory::volatile_memory::VolatileSlice<'a, B>` implements `imago::misc_helpers::ImagoAsRef<'a, vm_memory::volatile_memory::VolatileSlice<'a, B>
>`
     = note: required for `&VolatileSlice<'_>` to implement `imago::misc_helpers::ImagoAsRef<'_, vm_memory::volatile_memory::VolatileSlice<'_, _>>`
note: required by a bound in `IoVectorMut::<'a>::from_volatile_slice`
    --> /home/xize.gx/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/imago-0.2.0/src/io_buffers.rs:1002:19
     |
 999 |     pub fn from_volatile_slice<
     |            ------------------- required by a bound in this associated function
...
1002 |             Item: ImagoAsRef<'a, vm_memory::VolatileSlice<'a, B>>,
     |                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ required by this bound in `IoVectorMut::<'a>::from_volatile_slice`
```

I'm not quite familiar with Rust, how to resolve it, just upgrade libkrun vm-memory to 0.18, but then it will need more unrelated vm-memory code adaption, what's the common practice on this..

XanClic: Ah, would it help if I change imago’s version requirement to `vm-memory = { version = ">=0.16, <0.19", optional = true }`?

hsiangkao: > Ah, would it help if I change imago’s version requirement to `vm-memory = { version = ">=0.16, <0.19", optional = true }`?

maybe... that would cause less churn..

XanClic: I’m more than happy to do that, the imago code works across all those versions (whereas I believe the update to 0.18 will require changes to libkrun), I just don’t know whether it can fully resolve the problem :) From local testing, it looks like it works, so I’ll do it.

XanClic: I’ve pushed it to the `main` branch, so if you use the `git = "https://gitlab.com/hreitz/imago.git", branch = "main"` reference again, will it use the 0.16 version? (You may need to play around with Cargo.lock, e.g. reset it to what it was before, with just a single vm-memory dependency)

hsiangkao: > I’ve pushed it to the `main` branch, so if you use the `git = "https://gitlab.com/hreitz/imago.git", branch = "main"` reference again, will it use the 0.16 version? (You may need to play around with Cargo.lock, e.g. reset it to what it was before, with just a single vm-memory dependency)

Yes, that works now.

```sh
$make BLK=1 NET=1 -j12
cargo build --release --features blk --features net
rm -f libkrun.pc libkrun.pc-t
sed -e 's|@prefix@|/usr/local|' \
    -e 's|@libdir@|/usr/local/lib64|' \
    -e 's|@includedir@|/usr/local/include|' \
    -e 's|@PACKAGE_NAME@|libkrun|' \
    -e 's|@PACKAGE_VERSION@|1.16.0|' \
    libkrun.pc.in > libkrun.pc-t
mv libkrun.pc-t libkrun.pc
    Updating git repository `https://gitlab.com/hreitz/imago.git`
     Locking 1 package to latest compatible version
      Adding imago v0.2.0 (https://gitlab.com/hreitz/imago.git?branch=main#04dd6991)
   Compiling imago v0.2.0 (https://gitlab.com/hreitz/imago.git?branch=main#04dd6991)
   Compiling devices v0.1.0 (/home/xize.gx/libkrun/src/devices)
   Compiling vmm v0.1.0 (/home/xize.gx/libkrun/src/vmm)
   Compiling libkrun v1.16.0 (/home/xize.gx/libkrun/src/libkrun)
    Finished `release` profile [optimized] target(s) in 54.91s
cp target/release/libkrun.so target/release/libkrun.so.1.16.0
```

XanClic: Great! I’ve published 0.2.1 with that change.

hsiangkao: I've updated the PR, @slp could you leave more hints?

slp:
@hsiangkao LGTM, but please rebase on current upstream.

hsiangkao:
> @hsiangkao LGTM, but please rebase on current upstream.

Done.

slp:
@hsiangkao please commit the changes to `Cargo.lock` (resulting from the update of `imago`) too. That's the reason why CI is failing.

hsiangkao:
> @hsiangkao please commit the changes to `Cargo.lock` (resulting from the update of `imago`) too. That's the reason why CI is failing.

Ah sorry, done.

slp (approved):
@hsiangkao Thanks, let's get it merged.

Thanks for your help, too @XanClic !

hsiangkao:
> @hsiangkao Thanks, let's get it merged.
> 
> Thanks for your help, too @XanClic !

Many thanks for considering this feature too!

Issue #460, opened 2025-11-10
Inability to boot with memory below 1074mb


ggoodman:
Note: I'm using a bit of a weird setup, where I'm statically linking `libkrun` and embedding the kernel using `#include_bytes!` in my host CLI instead of using `libkrunfw`. I'll add the relevant patches below as an appendix in case this is a case of me breaking things.

# Test scenarios

## Scenario 1: `--memory=1073mb`

Here is the full log. No evidence of my init booting. Exit code: `0`.

```
[2025-11-10T14:22:31Z DEBUG vmm::macos::vstate] Guest memory host_addr=0x300000000 guest_addr=40000000 len=3ff00000
[2025-11-10T14:22:31Z DEBUG vmm::macos::vstate] Guest memory host_addr=0x130000000 guest_addr=80000000 len=20000000
[2025-11-10T14:22:31Z DEBUG utils::macos::epoll] EpollEvent new: 3
[2025-11-10T14:22:31Z DEBUG utils::macos::epoll] EpollEvent data: 3
[2025-11-10T14:22:31Z DEBUG utils::macos::epoll] add fd in: 3
[2025-11-10T14:22:31Z DEBUG devices::legacy::aarch64::gpio] SET_IRQ_LINE (GPIO)=33
[2025-11-10T14:22:31Z DEBUG utils::macos::epoll] EpollEvent new: 33
[2025-11-10T14:22:31Z DEBUG utils::macos::epoll] EpollEvent data: 33
[2025-11-10T14:22:31Z DEBUG utils::macos::epoll] add fd in: 33
[2025-11-10T14:22:31Z DEBUG devices::virtio::mmio[balloon]] set_irq_line: 34
[2025-11-10T14:22:31Z DEBUG utils::macos::epoll] EpollEvent new: 49
[2025-11-10T14:22:31Z DEBUG utils::macos::epoll] EpollEvent data: 49
[2025-11-10T14:22:31Z DEBUG utils::macos::epoll] add fd in: 49
[2025-11-10T14:22:31Z DEBUG devices::virtio::mmio[rng]] set_irq_line: 35
[2025-11-10T14:22:31Z DEBUG utils::macos::epoll] EpollEvent new: 68
[2025-11-10T14:22:31Z DEBUG utils::macos::epoll] EpollEvent new: 70
[2025-11-10T14:22:31Z DEBUG utils::macos::epoll] EpollEvent new: 66
[...]

PR #476, opened 2025-11-28
arch/aarch64: use different RAM start addresses

slp:
EDK2 expects RAM to start at 0x4000_0000, and a while ago we decided to
    make this default for both the EFI and non-EFI flavors. The problem with
    this approach is that, on aarch64, the Linux kernel expects to be loaded
    at 0x8000_0000, meaning that region needs to be covered by RAM.

    For large VMs this isn't a problem but for VMs smaller than 1GB, that
    region gets uncovered. In theory, we could add another region dedicated
    to the kernel at 0x8000_0000 and let the rest of RAM start at
    0x4000_0000, but this has two problems:

    - We would be using more RAM than what we were configured to.
    - It's very hard to find out exactly the size of the kernel region.

    So, instead, let's go back to use two different start addresses,
    depending on wether we're booting a firmware or a kernel, but let's do
    it in a cleaner way, so we don't depend on having an independent flavor.

    Fixes: #460

### DIFF
--- a/src/arch/src/aarch64/mod.rs
+++ b/src/arch/src/aarch64/mod.rs
@@ -84,23 +84,19 @@ pub fn arch_memory_regions(
 }
 
 /// Configures the system and should be called once per vm before starting vcpu threads.
-/// For aarch64, we only setup the FDT.
-///
-/// # Arguments
-///
-/// * `guest_mem` - The memory to be used by the guest.
-/// * `cmdline_cstring` - The kernel commandline.
-/// * `vcpu_mpidr` - Array of MPIDR register values per vcpu.
-/// * `device_info` - A hashmap containing the attached devices for building FDT device nodes.
-/// * `gic_device` - The GIC device.
-/// * `initrd` - Information about an optional initrd.
+/// For aarch64, we only setup SMBIOS.
 #[allow(clippy::too_many_arguments)]
 pub fn configure_system(
-    _guest_mem: &GuestMemoryMmap,
-    _smbios_oem_strings: &Option<Vec<String>>,
+    guest_mem: &GuestMemoryMmap,
+    mem_info: &ArchMemoryInfo,
+    smbios_oem_strings: &Option<Vec<String>>,
 ) -> super::Result<()> {
-    smbios::setup_smbios(_guest_mem, layout::SMBIOS_START, _smbios_oem_strings)
-        .map_err(Error::Smbios)?;
+    // When booting EFI, RAM starts at 0x4000_0000, while when doing a direct kernel
+    // boot RAM starts at 0x8000_0000. Only write SMBIOS in the former case.
+    if mem_info.ram_start_addr < layout::SMBIOS_START {
+        smbios::setup_smbios(guest_mem, layout::SMBIOS_START, smbios_oem_strings)
+            .map_err(Error::Smbios)?;
+    }
 
     Ok(())
 }
--- a/src/vmm/src/lib.rs
+++ b/src/vmm/src/lib.rs
@@ -309,8 +309,12 @@ impl Vmm {
 
         #[cfg(target_arch = "aarch64")]
         {
-            arch::aarch64::configure_system(&self.guest_memory, _smbios_oem_strings)
-                .map_err(Error::ConfigureSystem)?;
+            arch::aarch64::configure_system(
+                &self.guest_memory,
+                &self.arch_memory_info,
+                _smbios_oem_strings,
+            )
+            .map_err(Error::ConfigureSystem)?;
         }
 
         #[cfg(target_arch = "riscv64")]


src/arch/src/aarch64/linux/regs.rs
```diff
@@ -109,7 +108,12 @@ arm64_sys_reg!(MPIDR_EL1, 3, 0, 0, 0, 5);
 /// * `cpu_id` - Index of current vcpu.
 /// * `boot_ip` - Starting instruction pointer.
 /// * `mem` - Reserved DRAM for current VM.
```
tylerfanelli: `mem` was removed in favor of `mem_info`.

src/arch/src/aarch64/linux/regs.rs
```diff
@@ -124,7 +128,7 @@ pub fn setup_regs(vcpu: &VcpuFd, cpu_id: u8, boot_ip: u64, mem: &GuestMemoryMmap
         // "The device tree blob (dtb) must be placed on an 8-byte boundary and must
         // not exceed 2 megabytes in size." -> https://www.kernel.org/doc/Documentation/arm64/booting.txt.
         // We are choosing to place it the end of DRAM. See `get_fdt_addr`.
```
tylerfanelli: `get_fdt_addr` was removed.

src/arch/src/aarch64/mod.rs
```diff
@@ -43,13 +46,23 @@ pub fn arch_memory_regions(
...
+        DRAM_MEM_START_KERNEL
+    };
     let page_size: usize = unsafe { libc::sysconf(libc::_SC_PAGESIZE).try_into().unwrap() };
     let dram_size = align_upwards!(size, page_size);
-    let ram_last_addr = layout::DRAM_MEM_START + (dram_size as u64);
+    let ram_last_addr = ram_start_addr + (dram_size as u64);
     let shm_start_addr = ((ram_last_addr / 0x4000_0000) + 1) * 0x4000_0000;
-    let fdt_addr = ram_last_addr - layout::FDT_MAX_SIZE as u64;
+    let fdt_addr = if firmware_size.is_some() {
+        DRAM_MEM_START_EFI
```
tylerfanelli, line 59: So if EFI firmware exists, the FDT is placed at the _beginning_ of DRAM.
```
let fdt_addr = if firmware_size.is_some() {
        DRAM_MEM_START_EFI
```

Yet, if direct kernel boot, the FDT is placed at the _end_ of DRAM?
```
} else {
        ram_last_addr - layout::FDT_MAX_SIZE as u64
};
```

src/arch/src/aarch64/mod.rs
```diff
@@ -84,23 +84,19 @@ pub fn arch_memory_regions(
...
-    _smbios_oem_strings: &Option<Vec<String>>,
+    guest_mem: &GuestMemoryMmap,
+    mem_info: &ArchMemoryInfo,
+    smbios_oem_strings: &Option<Vec<String>>,
 ) -> super::Result<()> {
-    smbios::setup_smbios(_guest_mem, layout::SMBIOS_START, _smbios_oem_strings)
-        .map_err(Error::Smbios)?;
+    // When booting EFI, RAM starts at 0x4000_0000, while when doing a direct kernel
+    // boot RAM starts at 0x8000_0000. Only write SMBIOS in the former case.
+    if mem_info.ram_start_addr < layout::SMBIOS_START {
```
tylerfanelli, line 96: Instead of doing these calculations, would it make sense to include a `bool` indicating if there is EFI firmware present? It may make no functional difference, but it would be a bit simpler in setting up the memory regions and such.

src/arch/src/aarch64/linux/regs.rs
slp: Fixed, thanks!

src/arch/src/aarch64/linux/regs.rs
slp: Ditto.

src/arch/src/aarch64/mod.rs
slp: Yes. For direct kernel boot on aarch64, the kernel *must* be placed at 0x8000_0000. Ideally, DRAM should also start there, because otherwise it needs to be large enough to cover both the start address and 0x8000_0000 (this is what this PR is all about). This means we can't put FDT at the beginning of DRAM, so we do the next best thing, which is putting it at the end.

On the other hand, EDK2 is firmware so it's loaded outside DRAM (at 0x0, while DRAM starts at 0x4000_0000), and expects to find the FDT at 0x4000_0000. So we're pretty much stuck with having to do this different memory layout dance.

src/arch/src/aarch64/mod.rs
slp: I honestly don't think it's worth further extending `ArchMemoryInfo` when we can just check `ram_start_addr`. We won't be supporting arbitrary DRAM start addresses anyway.

PR #477, opened 2025-11-28
aarch64: small fixes for BSD systems

slp:
This PR contains some small fixes that enable booting BSD systems on aarch64. Enables booting FreeBSD with an EDK2 firmware and smolBSD with a direct kernel boot.

### DIFF
--- a/src/vmm/src/builder.rs
+++ b/src/vmm/src/builder.rs
@@ -11,7 +11,7 @@ use kernel::cmdline::Cmdline;
 use std::collections::HashMap;
 use std::fmt::{Display, Formatter};
 use std::fs::File;
-use std::io::{self, Read};
+use std::io::{self, IsTerminal, Read};
 use std::os::fd::AsRawFd;
 use std::os::fd::{BorrowedFd, FromRawFd};
 use std::path::PathBuf;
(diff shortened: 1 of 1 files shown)


nohajc:
It does boot fine, so that part's fixed. However, I still have the issue that ^C will terminate the whole VM. Also, ^D worked with your previous branch, now it's just literally echoed to the console and nothing happens. Maybe something to do with terminal raw mode? Arrows for navigation don't work either.

slp:
> It does boot fine, so that part's fixed. However, I still have the issue that ^C will terminate the whole VM. Also, ^D worked with your previous branch, now it's just literally echoed to the console and nothing happens. Maybe something to do with terminal raw mode? Arrows for navigation don't work either.

Yes, that's because we aren't setting the raw mode for the serial device. I was planning on fixing it in a different PR, but I guess we might fix it here and now as well.

nohajc:
Either way works for me. Thank you!

slp:
@nohajc I've added a commit setting up raw mode for legacy serial devices. Please give it a try if you have the chance.

nohajc:
@slp Works great now!

PR #479, opened 2025-11-30
vmm: linux: Do not use guest_memfd outside of TEE builds (fixes #478)

### LINKED ISSUE #478: guest memfd fails on aarch64 / experimental private memory is used..
Since commit [32e200bd6 ("KVM: arm64: Enable support for guest_memfd backed memory")](https://git.kernel.org/pub/scm/linux/kernel/git/next/linux-next.git/commit/?id=32e200bd6e4407c31aef0392be041274bfa3f20e), guest memfd is enabled on aarch64.. and muvm explodes with an EINVAL, because that commit has a check

valpackett:
aarch64 KVM has recently introduced guest_memfd support, but since there are no VM TEEs with upstream support on that arch, it returns EINVAL when we try to create a non-mappable memfd (which is what flags==0 implies).

Then.. even if we make it mappable and skip setting the private flag, we get a 100% CPU busy loop in the guest. Same happens in QEMU though so that's not a libkrun bug. Still, let's avoid *all* of that by not trying to use guest_memfd for protecting the VM's memory outside of platforms where that's explicitly supported (SEV/TDX).

### DIFF
--- a/src/vmm/src/linux/vstate.rs
+++ b/src/vmm/src/linux/vstate.rs
@@ -669,7 +669,17 @@ impl Vm {
         let start = region.start_addr().raw_value();
         let end = start + region.len();
 
-        if !self.fd.check_extension(GuestMemfd) {
+        // GuestMemfd is generally intended for either of two purposes:
+        // * sharing the memory with out-of-process components, and conversely,
+        // * hiding the memory completely from the VMM process (Confidential Computing).
+        //
+        // We only use it for the second use case currently, so don't even try to use it
+        // outside of TEE builds. Software-protected VMs are only available on x86_64 and
+        // are marked with strongly-worded warnings about them being for development only,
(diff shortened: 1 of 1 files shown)


src/vmm/src/linux/vstate.rs
```diff
@@ -686,6 +696,10 @@ impl Vm {
                     .map_err(Error::SetUserMemoryRegion)?;
             };
         } else {
+            if !self.fd.check_extension(GuestMemfd) {
```
tylerfanelli, line 699: Can this be made an `else if` to avoid unnecessary nesting?

slp: If this was made into an `else if`, we would need another conditional branch to return `Error::KvmCap(GuestMemfd)`. It's better as it is here.

slp:
@valpackett thanks a lot for investigating this issue and getting it fixed

PR #481, opened 2025-12-03
examples: Fix obsolete comment

FabienTregan:
The code had been updated but not the comment.
Maybe replacing `(2048 MiB)` with `(in MiB)` would avoid the problem from occurring again, but this version matched the code in `chroot_vm.c` (from which it is probably copied ?)


mtjhrc (approved):
Thanks

PR #482, opened 2025-12-05
vmm/amdsnp: Remove sev library dependency

tylerfanelli:
The sev library was useful when the SEV-SNP Linux API was unstable, but has become a source of issues when updating dependent packages. The KVM SEV-SNP API is now stable, so we can import a library just containing the launch API.

Works as intended, just need to release the initial version of sev-snp-launch to crates.io and this is ready to merge.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -157,15 +157,6 @@ checksum = "72b3254f16251a8381aa12e40e3c4d2f0199f8c6508fbecb9d91f575e0fbb8c6"
 
 [[package]]
 name = "bincode"
-version = "1.3.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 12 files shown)


tylerfanelli:
Marking as draft. It works fine, but I'm exploring either importing this library into libkrun directly or implementing the ioctls directly with kvm_ioctls. Either of these make one less dependency we need to worry about during packaging.

slp:
FWIW, this sounds reasonable to me.

slp:
@tylerfanelli do we want this to be part of the next release? If so, we'd need to it merge it ASAP.

tylerfanelli:
Tested on linux 6.17.0 SEV-SNP, works as intended.

tylerfanelli:
> @tylerfanelli do we want this to be part of the next release? If so, we'd need to it merge it ASAP.

@slp can you review?

PR #483, opened 2025-12-10
Add explicit vsock and TSI configuration API

mtjhrc:
Not sure if this is the API we want to go with , we should have some way to control TSI and vsock.

### DIFF
--- a/include/libkrun.h
+++ b/include/libkrun.h
@@ -327,6 +327,10 @@ int32_t krun_add_virtiofs2(uint32_t ctx_id,
    as required by gvproxy in vfkit mode. */
 #define NET_FLAG_VFKIT 1 << 0
 
+/* TSI (Transparent Socket Impersonation) feature flags for vsock */
+#define KRUN_TSI_HIJACK_INET  (1 << 0)
+#define KRUN_TSI_HIJACK_UNIX  (1 << 1)
+
 /* Taken from uapi/linux/virtio_net.h */
 #define NET_FEATURE_CSUM 1 << 0
 #define NET_FEATURE_GUEST_CSUM 1 << 1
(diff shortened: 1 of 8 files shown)


mtjhrc:
@slp After https://github.com/containers/libkrunfw/pull/106 we don't enable UNIX hijacking in libkrun implicitly by default. Should I restore that behavior and enable it?

slp:
@mtjhrc this is looking good, what's missing to be ready for review?

mtjhrc:
Note, this PR also removes the heuristic that enabled unix socket hijacking if "/"  is the rootfs. This got broken anyways with the new libkrunfw, which now requires a kernel argument for hijacking unix sockets.

include/libkrun.h
```diff
@@ -862,6 +866,27 @@ int32_t krun_add_vsock_port2(uint32_t ctx_id,
                              uint32_t port,
                              const char *c_filepath,
                              bool listen);
+
+/**
+ * Add a vsock device with specified TSI features.
+ *
+ * By default, libkrun creates a vsock device implicitly with TSI hijacking
+ * enabled based on heuristics. Calling this function overrides the implicit
+ * behavior and explicitly configures the vsock device.
```
slp: The comment above seems to point that calling this function will automatically disable the implicit vsock, but that's not what happens. Calling this one without calling `krun_disable_implicit_vsock()` first will lead to `[libcrun:krun]: could not add vsock configuration: File exists`.

I think we need to either have this function also disable the implicit vsock, or clarify that `krun_disable_implicit_vsock()` needs to be called first.

slp:
Tested with crun and found no regressions. Also, enabling TSI_UNIX from crun does make unix sockets to work, as expected.

I've only have a minor complain explained in a comment, otherwise LGTM.

slp (changes requested):
Please take a look at the comment about disabling the implicit vsock.

include/libkrun.h
mtjhrc: Yeah, I forgot to update the comment. The comment now matches the actuall behavior (user is expected to call `krun_disable_implicit_vsock` first).

PR #485, opened 2025-12-11
hvf: forward read/write ops on MDCCINT_EL1 and CNTHCTL_EL2

slp:
Starting with Tahoe, macOS HVF traps to the VMM when the guest accesses MDCCINT_EL1. Deal with this by using the HVF API to read or write the value of this system register.

This fixes nested virt on Tahoe.


jakecorrenti (approved):
Code LGTM

PR #487, opened 2025-12-13
Fix aarch64 unit tests and extend CI to run them on self-hosted runner

slp:


mtjhrc (approved):
This is great, and was much needed
Thanks!

PR #490, opened 2025-12-17
nitro: improve console support

jakecorrenti:
Rather than tightly coupling the application output with the kernel
output, which is only enabled when the debug start flag is passed to the
nitro enclave, separate the two.

The application output socket in the guest will always be created and
the host is expected to create a socket to connect to the guest and
listen for output.

The debug output socket can be optionally created by the host if it is
desired, no longer forcing the user to see the kernel messages.

### DIFF
--- a/examples/nitro.c
+++ b/examples/nitro.c
@@ -102,6 +102,70 @@ bool parse_cmdline(int argc, char *const argv[], struct cmdline *cmdline)
     return false;
 }
 
+void *listen_enclave_app_output(void *opague)
+{
+    int ret, sock_fd, bytes_read, client_fd;
+    char buffer[BUFSIZE];
+    struct sockaddr_vm addr;
+    struct timeval timeval;
+
+    sock_fd = socket(AF_VSOCK, SOCK_STREAM, 0);
+    if (sock_fd < 0) {
+        perror("unable to create host socket for application output");
+        return (void *)-1;
+    }
+
+    bzero((char *) &addr, sizeof(struct sockaddr_vm));
+    addr.svm_family = AF_VSOCK;
+    addr.svm_cid = VMADDR_CID_ANY;
+    addr.svm_port = 8081;
+
+    memset(&timeval, 0, sizeof(struct timeval));
+    timeval.tv_sec = 5;
+
+    ret = setsockopt(sock_fd, AF_VSOCK, SO_VM_SOCKETS_CONNECT_TIMEOUT, (void *) &timeval, sizeof(struct timeval));
+    if (ret < 0) {
+        close(sock_fd);
+        perror("unable to set socket options for application output socket");
+        return (void *)-1;
+    }
+
+    ret = bind(sock_fd, (struct sockaddr *) &addr, sizeof(addr));
+    if (ret < 0) {
+        close(sock_fd);
+        perror("unable to bind the host application output socket to the address");
+        return (void *)-1;
+    }
+
+    ret = listen(sock_fd, 1);
+    if (ret < 0) {
+        close(sock_fd);
+        perror("unable to listen for incoming connection to host application output socket.");
+        return (void *)-1;
+    }
+
+    client_fd = accept(sock_fd, NULL, NULL);
+    if (client_fd < 0) {
+        close(sock_fd);
+        perror("unable to connect host application output socket to guest socket");
+        return (void *)-1;
+    }
+
+    close(sock_fd);
+
+    while((bytes_read = read(client_fd, buffer, sizeof(buffer) - 1)) > 0) {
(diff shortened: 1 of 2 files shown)


jakecorrenti:
> We shouldn't concern ourselves with journald at this point, that's an implementation detail that crun can handle. Rather, let's just print the application output.

should be set

examples/nitro.c
```diff
@@ -95,6 +95,71 @@ bool parse_cmdline(int argc, char *const argv[], struct cmdline *cmdline)
...
+    }
+
+    ret = listen(sock_fd, 1);
+    if (ret < 0) {
+        close(sock_fd);
+        puts("unable to listen for incoming connection to host socket.");
+        return (void *)-1;
+    }
+
+    int client_fd;
```
tylerfanelli: Move to beginning of function.

examples/nitro.c
```diff
@@ -95,6 +95,71 @@ bool parse_cmdline(int argc, char *const argv[], struct cmdline *cmdline)
...
+        buffer[bytes_read] = '\0';
+        printf("%s", buffer);
+        fflush(stdout);
+    }
+
+    if (bytes_read != 0)
+        puts("\napp socket read error");
+
+    close(client_fd);
+    return (void *)0;
```
tylerfanelli, line 166: You return `(void *)-1` on error and `(void *)0` on success. What's the point of returning a `void *` in general if its just an error code? A `bool` would suffice, no?

examples/nitro.c
```diff
@@ -95,6 +95,71 @@ bool parse_cmdline(int argc, char *const argv[], struct cmdline *cmdline)
...
+        close(sock_fd);
+        puts("unable to set socket options");
+        return (void *)-1;
+    }
+
+    ret = bind(sock_fd, (struct sockaddr *) &addr, sizeof(addr));
+    if (ret < 0) {
+        close(sock_fd);
+        puts("unable to bind the host socket to the address");
+        exit(1);
```
tylerfanelli: This is the only `exit(1)` call, the rest `return (void *)-1`. Is this on purpose? Why exit the process?

examples/nitro.c
```diff
@@ -95,6 +95,71 @@ bool parse_cmdline(int argc, char *const argv[], struct cmdline *cmdline)
...
+void *listen_enclave_app_output(void *opague)
+{
+    int ret, sock_fd, bytes_read;
+    char buffer[BUFSIZE];
+    struct sockaddr_vm addr;
+    struct timeval timeval;
+
+    sock_fd = socket(AF_VSOCK, SOCK_STREAM, 0);
+    if (sock_fd < 0) {
+        puts("unable to create host socket");
```
tylerfanelli: s/puts/printf

examples/nitro.c
```diff
@@ -95,6 +95,71 @@ bool parse_cmdline(int argc, char *const argv[], struct cmdline *cmdline)
...
+
+    close(sock_fd);
+
+    while((bytes_read = read(client_fd, buffer, sizeof(buffer) - 1)) > 0) {
+        buffer[bytes_read] = '\0';
+        printf("%s", buffer);
+        fflush(stdout);
+    }
+
+    if (bytes_read != 0)
```
tylerfanelli: Seems you're trying to check an error code. Perhaps `bytes_read < 0` would be more clear.

examples/nitro.c
```diff
@@ -95,6 +95,71 @@ bool parse_cmdline(int argc, char *const argv[], struct cmdline *cmdline)
...
+        puts("\napp socket read error");
```
tylerfanelli: s/puts/printf

tylerfanelli (changes requested):

examples/nitro.c
jakecorrenti: `void *(*start_routine)(void *)` is the signature of the function pointer in the `pthread_create` function.

re-reading the manpages it says 
> The new thread terminates in one of the following ways:
[...] 
It returns from `start_routine()`, This is equivalent to calling `pthread_exit(3)` with the value supplied in the return statement. 

so maybe instead of `-1` I should be returning `1`.

examples/nitro.c
jakecorrenti: No, this wasn't on purpose. This was left out when I did a quick refactor

jakecorrenti:
@tylerfanelli PTAL, addressed your comments

examples/nitro.c
tylerfanelli: Ah, sorry, I didn't realize it was a thread handler. The -1 is probably preferred to indicate error in this case.

PR #491, opened 2025-12-19
nitro: Preliminary networking support with passt, remove cached EIF

tylerfanelli:


tylerfanelli:
@jakecorrenti This offers the preliminary support for you to rebase on top of. I'll revisit with a second patch series removing the unwraps and adding general cleanups after your PR is merged. The next changes will include both the application output and networking.

If you approve, we can merge this series, rebase yours on top, and merge that as well.

src/nitro/src/enclaves.rs
```diff
@@ -11,7 +11,7 @@ use nix::poll::{poll, PollFd, PollFlags, PollTimeout as NixPollTimeout};
 use std::{
     ffi::{CString, OsStr},
     fs,
-    io::{self, ErrorKind, Read, Write},
+    io::{self, Read, Write},
```
jakecorrenti, line 15: In a separate PR we should probably add a Clippy CI check for the nitro feature so that we can catch these things at review.

src/libkrun/src/lib.rs
```diff
@@ -384,6 +384,34 @@ impl TryFrom<ContextConfig> for NitroEnclave {
...
+                    let device = list.pop_front().unwrap();
+
+                    match device.lock().unwrap().backend() {
+                        VirtioNetBackend::UnixstreamFd(_) | VirtioNetBackend::UnixstreamPath(_) => {
+                        }
+                        _ => {
+                            error!("configured virtio-net backend must be unix stream");
+                            return Err(-libc::EINVAL);
+                        }
+                    };
```
jakecorrenti, line 402: Might look a little cleaner, but up to you :) 

```rust
if !matches!(
    device.lock().unwrap().backend(),
    VirtioNetBackend::UnixstreamFd(_) | VirtioNetBackend::UnixstreamPath(_)
) {
    error!("configured virtio-net backend must be unix stream");
    return Err(-libc::EINVAL);
}

```

src/nitro/src/enclaves.rs
```diff
@@ -72,7 +77,51 @@ impl NitroEnclave {
 
         self.write_exec(&mut stream)?;
 
-        Ok(cid)
+        match fork::fork().unwrap() {
```
jakecorrenti: Why do we need a new crate to do this? Can't we just use `libc::fork()`?

jakecorrenti (approved):

Issue #405, opened 2025-09-01
virtio-net: read/write multiple messages per syscall on macOS


nirs:
On macOS we can double read and write performance and reduce cpu usage with the private APIs:

```
ssize_t recvmsg_x(int s, const struct msghdr_x *msgp, unsigned int cnt, int flags);
ssize_t sendmsg_x(int s, const struct msghdr_x *msgp, unsigned int cnt, int flags);
```

These are like recvmmsg and sendmmsg in Linux/FreeBSD, transferring multiple messages in one syscall.

The best documentation is here:
https://github.com/nirs/vmnet-helper/blob/main/socket_x.h

Example usage in vmnet-helper:
- https://github.com/nirs/vmnet-helper/blob/a2809117dbc90e7a566954463af8142355d832f2/helper.c#L555 
- https://github.com/nirs/vmnet-helper/blob/a2809117dbc90e7a566954463af8142355d832f2/helper.c#L665

In both case we fallback to standard read/write as a backup.

I learned about this from bun:
https://github.com/oven-sh/bun/blob/main/packages/bun-usockets/src/internal/networking/bsd.h

Issue #385, opened 2025-08-06
virtio-net: read directly from the backend to the descriptor chain


slp:
Right now, we're using intermediate buffers when reading from the backends, and then writing to the descriptor chain. For backends capable of using `readv`, we should translate the descriptor chain into an `iovec` and use it to do the operation in a single copy.

slp:
@mtjhrc please let me know if you want to tackle this one yourself.

mtjhrc:
Yeah I can work on this later. 

I'm thinking maybe we should have it as some generic utility/abstraction for dealing with IO <-> descriptor chain reading/writing, It should also be useful for other devices. Anyways starting with virtio-net makes sense.

nirs:
See also #405

PR #493, opened 2025-12-22
net: add a new flag NET_FLAG_INCLUDE_VNET_HEADER

akerouanton:
This flag should be used to indicate to libkrun that downstream network backend wants to receive and transmit the virtio-net header along with Ethernet frames.

Network backends using this flag can then forward unmodified headers to another VM or build a sensible virtio_net_hdr (e.g. with GSO fields correctly set) such that receiving VM handles GSO'd frames properly.

### DIFF
--- a/src/devices/src/virtio/net/worker.rs
+++ b/src/devices/src/virtio/net/worker.rs
@@ -3,7 +3,7 @@ use crate::virtio::net::backend::ConnectError;
 use crate::virtio::net::tap::Tap;
 use crate::virtio::net::unixgram::Unixgram;
 use crate::virtio::net::unixstream::Unixstream;
-use crate::virtio::net::{MAX_BUFFER_SIZE, QUEUE_SIZE};
+use crate::virtio::net::{FRAME_HEADER_LEN, MAX_BUFFER_SIZE, QUEUE_SIZE};
 use crate::virtio::{DeviceQueue, InterruptTransport};
 
 use super::backend::{NetBackend, ReadError, WriteError};
@@ -29,7 +29,7 @@ pub struct NetWorker {
     rx_has_deferred_frame: bool,
 
     tx_iovec: Vec<(GuestAddress, usize)>,
-    tx_frame_buf: [u8; MAX_BUFFER_SIZE],
+    tx_frame_buf: [u8; MAX_BUFFER_SIZE + FRAME_HEADER_LEN],
     tx_frame_len: usize,
 }
 
@@ -40,6 +40,7 @@ impl NetWorker {
         interrupt: InterruptTransport,
         mem: GuestMemoryMmap,
         _vnet_features: u64,
+        include_vnet_header: bool,
         cfg_backend: VirtioNetBackend,
     ) -> Result<Self, ConnectError> {
         let backend = match cfg_backend {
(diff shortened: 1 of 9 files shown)


src/libkrun/src/lib.rs
```diff
@@ -967,14 +970,16 @@ pub unsafe extern "C" fn krun_add_net_unixstream(
         return -libc::EINVAL;
     }
 
-    if (features & !NET_ALL_FEATURES) != 0 {
+    if (features & !(NET_ALL_FEATURES | NET_FLAG_INCLUDE_VNET_HEADER)) != 0 {
         return -libc::EINVAL;
     }
```
mtjhrc: This check is incorrect on 2 levels:
 - as it stands this PR only adds support for this feature for unix datagrams, not streams (which this function is configuring) . 
- the NET_FLAG_INCLUDE_VNET_HEADER is a flag not a feature (as correctly in the krun_add_net_unixgram).

src/libkrun/src/lib.rs
```diff
@@ -967,14 +970,16 @@ pub unsafe extern "C" fn krun_add_net_unixstream(
...
 
+    let include_vnet_header: bool = flags & NET_FLAG_INCLUDE_VNET_HEADER != 0;
```
mtjhrc, line 980: As implied above, include_vnet_header, should always be false here, because it's not implemented for the stream socket.

src/libkrun/src/lib.rs
mtjhrc, line 1038: If you only implement the flag for the unixgram sockets, I think it would be better to put the bool inside the enum (like send_vfkit_magic is).

Note: I am not sure why send_vfkit_magic would not passed into `VirtioNetBackend::UnixgramFd`, that seems like a bug.

mtjhrc (changes requested):
Thanks for PR, it certainly makes sense to expose the header to interested backends.
I am curious though, could you share more detail, what backend you would use with this?

Please fix the mentioned confusion between stream/datagram sockets and features/flags. Also I guess it would be nice to also implement this for stream sockets too, but not strictly necessary.

akerouanton:
> I am curious though, could you share more detail, what backend you would use with this?

I tried for a couple hours to get a kernel stacktrace to pinpoint exactly where's the issue and why this PR fixes it, but to no avail. Let me describe that with plain words instead, but let me know if you want something more concrete.

I don't plan to use this flag with any existing backend right now. I'm working on a container runtime that leverages https://github.com/containerd/nerdbox and has its own virtual switch that connects multiple VMs. Every VM gets one or more virtio-net devices, and each interface is connected to a bridge interface (i.e. `ip link set master`). That bridge has a veth connected to it too (and at some point, it'll get multiple of them but nerdbox doesn't support multiple-containers-per-VM yet).

With TSO enabled, VM-to-VM connections are working correctly — packets are GSO'd on the sending side, and GRO'd on the receiving side. However, container-to-container connections are broken. After some debugging, I realized that disabling GSO on the sending veth fixed the issue and that was probably caused by invalid GSO fields in virtio-net headers.

I believe that with the above topology GRO is supposed to happen at the veth level (although I might be wrong), and missing GSO fields would cause `skb`s to have some fields unset, preventing GRO to happen.

With this change, my virtual switch is forwarding virtio-net headers sent by the sending side without recomputing L4 checksums (which was required for VM-to-VM scenarios to work), and it fixes container-to-container communications too. I also think that this should help with host-to-[VM|container] communications as it'd allow to send large frames with partial checksums.

> Please fix the mentioned confusion between stream/datagram sockets and features/flags. Also I guess it would be nice to also implement this for stream sockets too, but not strictly necessary.

Sorry for the nasty mistakes, let me fix them!

akerouanton:
@mtjhrc I added `include_vnet_header` to all network backends but to keep the bespoke 'frame length' header used by the unixstream backend, I had to extend `NetWorker`'s TX buffer by 4 bytes. The vnet header is now located at `tx_frame_buf[4..14]`.

`Unixstream` either writes the frame length to `0..4` or `10..14` dependening on whether the vnet header is included. Other backends have been modified to ignore the first four bytes.

The `Tap` backend also takes an `include_vnet_header` to keep it aligned with other backends. But since vnet headers have always been included, there's no way to make that configurable without introducing a breaking change in the API. Maybe something to fix in a major API version, or via a new API method, or maybe it's not worth it (in that case I can remove my change if you prefer).

Anyway, this should be in a better shape. PTAL.

slp:
@mtjhrc PTAAL

src/devices/src/virtio/net/worker.rs
mtjhrc: The offset here needs to be the same as used for `backend.write_frame` call, 
```suggestion
                .try_finish_write(vnet_hdr_len()  + FRAME_HEADER_LEN, &self.tx_frame_buf[..self.tx_frame_len])
```

mtjhrc:
Hi, sorry for the late reply.
I would have aproved and merged this, but this seemed to be buggy at least with with unixstream. Right now I'm working on https://github.com/containers/libkrun/issues/385, https://github.com/containers/libkrun/issues/405. I'll integrate this feature there, along with tests - the device side would be completely different - this removes the intermediary buffer entirely).

akerouanton:
@mtjhrc @slp Do you plan to get #530 ready and merge it any time soon (i.e. before the next minor)? Otherwise, maybe we can consider merging this PR first and then get #530 in a subsequent release?

> I would have aproved and merged this, but this seemed to be buggy at least with with unixstream.

I fixed that in a previous iteration, and fixed conflicts yesterday. I think it's good to go now.

slp:
> @mtjhrc @slp Do you plan to get #530 ready and merge it any time soon (i.e. before the next minor)? Otherwise, maybe we can consider merging this PR first and then get #530 in a subsequent release?

I think neither this one nor #530 qualify for a minor release. This will need to wait for 1.18, sorry.

akerouanton:
To clarify, when I said next minor, I meant 1.18 not 1.17.x 🙂

mtjhrc:
> To clarify, when I said next minor, I meant 1.18 not 1.17.x 🙂

We should be able to get #530 before 1.18 anyway, but just in case:

Regarding this PR:
I think `unixstream` is still incorect (things around `try_finish_write` are suspicious, but I am removing the buffer as part of #530 anyway).

Since you care about the `unixgram` backend, I propose you drop the `unixstream` implementation from this PR after all and I will implement that as part of #530 

The unixgram implementation looks OK.

slp:
> > To clarify, when I said next minor, I meant 1.18 not 1.17.x 🙂
> 
> We should be able to get #530 before 1.18 anyway, but just in case:

I'd prefer not touching net code in 1.17, but it'll definitely go in 1.18.

mtjhrc:
> > > To clarify, when I said next minor, I meant 1.18 not 1.17.x 🙂
> > 
> > 
> > We should be able to get #530 before 1.18 anyway, but just in case:
> 
> I'd prefer not touching net code in 1.17, but it'll definitely go in 1.18.

I meant merged into main before 1.18 release, not released as 1.17.x :sweat_smile:

PR #494, opened 2025-12-28
Rust API

jeromegn:
This PR adds a rudimentary Rust API for using libkrun. The idea is to make this new API the main one. The C API would use it entirely once this is done.

Ultimately, this new API should prevent unrepresentable states and allow usage from Rust in an ergonomic manner.

I am looking for early feedback before I continue my work. I'm trying not to touch much outside of the API surface (keeping other changes for when/if this PR is accepted).

Is the team open to this kind of change?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -888,6 +888,7 @@ dependencies = [
  "once_cell",
  "polly",
  "rand",
+ "thiserror 1.0.69",
  "utils",
  "vm-memory",
  "vmm",
@@ -1306,6 +1307,13 @@ dependencies = [
  "proc-macro2",
  "quote",
  "syn",
+]
+
+[[package]]
+name = "rust_api"
+version = "0.1.0"
+dependencies = [
+ "libkrun",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,5 @@
 [workspace]
-members = ["src/libkrun", "src/krun_input"]
+members = ["src/libkrun", "src/krun_input", "examples/rust_api"]
 exclude = ["examples/gtk_display"]
 resolver = "2"
 
--- a/examples/rust_api/Cargo.toml
+++ b/examples/rust_api/Cargo.toml
@@ -1,0 +1,8 @@
+[package]
+name = "rust_api"
+description = "Reference implementation of using the libkrun rust API"
+version = "0.1.0"
+edition = "2024"
+
+[dependencies]
+libkrun = { path = "../../src/libkrun" }
--- a/examples/rust_api/src/main.rs
+++ b/examples/rust_api/src/main.rs
@@ -1,0 +1,9 @@
+fn main() {
+    let mut builder = krun::Builder::new();
+
+    builder.set_root(&std::env::args().nth(1).expect("missing NEWROOT argument"));
+
+    let ctx = builder.build();
+
+    ctx.start_enter().unwrap()
+}
--- a/init/jsmn.h
+++ b/init/jsmn.h
@@ -237,7 +237,7 @@ static int jsmn_parse_string(jsmn_parser *parser, char *js, const size_t len,
             case 't':
                 break;
             /* Allows escaped symbol \uXXXX */
-            case 'u':
+            case 'u': {
                 char unicode[5];
                 long ascii;
 
@@ -270,6 +270,7 @@ static int jsmn_parse_string(jsmn_parser *parser, char *js, const size_t len,
                 parser->pos--;
                 js[parser->pos] = (char)ascii;
                 break;
+            }
             /* Unexpected symbol */
             default:
                 parser->pos = start;
--- a/src/libkrun/Cargo.toml
+++ b/src/libkrun/Cargo.toml
@@ -33,6 +33,7 @@ polly = { path = "../polly" }
 utils = { path = "../utils" }
 vmm = { path = "../vmm" }
 rand = "0.9.2"
+thiserror = "1"
 
 [target.'cfg(target_os = "macos")'.dependencies]
 hvf = { path = "../hvf" }
@@ -46,4 +47,4 @@ vm-memory = { version = ">=0.13", features = ["backend-mmap"] }
 
 [lib]
 name = "krun"
-crate-type = ["cdylib"]
+crate-type = ["cdylib", "lib"]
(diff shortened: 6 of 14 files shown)


ggoodman:
I understood from https://github.com/containers/libkrun/discussions/411 and some discussions on the Matrix channel that there is an interest in making a rust API a first class citizen in a v2.

jeromegn:
Didn’t know there was a matrix channel. I’ll join it if I can find it :)

I’m doing this because I definitely need to expose a more advanced API I
can use from rust. I’ve modified the virtio block device API to be a trait
and I’ve implemented my own custom device which enables great performance
for special use cases.

It makes no sense to expose that as a C API. Or rather: it’s a non goal for
me.

Would it still be acceptable in v1? It’s not a breaking change.

mtjhrc:
So I like the direction you took with the API design, it seems like a reasonable stepping stone between v1 and v2 libkrun. There are some other things we want in the v2 API such as to make everything more explicit - not adding or configuring the virtio devices based on heuristics and I think we also need a clear plan to interact with a running VM.

I am a bit concerned about 2 things:
 - breaking something when refactoring the whole API surface implementation (we can mitigate that if we are careful and test the new changes here)
 -  Do we want/need the Rust API to be stable until libkrun v2? 

> I’m doing this because I definitely need to expose a more advanced API I
can use from rust. I’ve modified the virtio block device API to be a trait

> It makes no sense to expose that as a C API. Or rather: it’s a non goal for me.

Yeah, this is currently problematic, I ran into this when I implemented the gpu display output support. We actually already expose some traits as a C API - specifically the traits in `krun_input` and `krun_display` crates, but it is very annoying (and error prone) to expose new traits like that.  

These traits are actually implemented by the Rust `gui_vm` example (via the `krun_gtk_display` example/reference implementation). The annoying part is that libkrun is written in Rust and the example is also written in Rust, but the function calls have to go through the C API. So the bindings are actually bidirectional (basically constructing a struct with function pointers from a trait + implementing the trait for the struct with function pointers). 

I am hoping to simplify this situation eventually with v2 - since we would have a native Rust API we could drop one side of the bindings, and also possibly partially automate the generation of the C bindings via macros. This could then also be used to expose the hooks you would implement for the  virtio block as a C API.

> I’ve modified the virtio block device API to be a trait
and I’ve implemented my own custom device which enables great performance
for special use cases.

This  to me seems quite difficult to get right, if we expose a very fine grained API, we could end up locking ourselves into how the internals of libkrun work, which could impede refactoring and performance optimizations within libkrun itself, so we have to be careful with that.

mtjhrc:
Ok, so we had a meeting today and I asked the other maintainers what they think. We concluded that we can proceed forward with this for v1 too.

I have some thoughts:  
 - please split changes into commits and don't mix in irrelevant changes (e.g. the formatting)
 - I feel like before changing the implementation of the C API functions, adding some simple unit tests that assert the state of the builder after calling the functions would be helpful. This could simplify the review process, and make sure we don't break anything. (especially the convoluted to-be-deprecated implicit APIs)
 - put the C API and Rust API in separate files
 - I like that you separated the step of building VM  and running the VM

Thanks!

jeromegn:
Thanks for bringing it up at your meeting! That all sounds reasonable to me. I don't know exactly when / if I'll get to completing this.

If I don't get to it, at least it gives others some idea of how to do it.

d-e-s-o:
Just an idea, but from my point of view the main hurdle I have is that I cannot consume the library from Rust/Cargo directly, but have to go through an "installed" C library. If we could just have something like
```patch
--- src/libkrun/Cargo.toml
+++ src/libkrun/Cargo.toml
@@ -46,4 +46,4 @@ vm-memory = { version = ">=0.13", features = ["backend-mmap"] }

 [lib]
 name = "krun"
-crate-type = ["cdylib"]
+crate-type = ["cdylib", "lib"]
```

that would already go a long way already and eliminate a bunch of hoops such as the entire `bindings` module dance in https://github.com/containers/krunvm/blob/main/src/bindings.rs, for example.

(this doesn't work right out the gate, because `init` has to be built, but that could [and IMO should] easily be done from a build script; I actually got all of that working already)

This is certainly a different direction than proposed, but to me the two seem orthogonal and not in conflict, with this small patch being relatively trivial to land (compared to stabilization of the entire Rust API surface). Just bringing this up as an additional direction to lower cost for Rust integration; not saying that it necessarily addresses OP's main use case.

slp:
> Just an idea, but from my point of view the main hurdle I have is that I cannot consume the library from Rust/Cargo directly, but have to go through an "installed" C library. If we could just have something like
> 
> ```diff
> --- src/libkrun/Cargo.toml
> +++ src/libkrun/Cargo.toml
> @@ -46,4 +46,4 @@ vm-memory = { version = ">=0.13", features = ["backend-mmap"] }
> 
>  [lib]
>  name = "krun"
> -crate-type = ["cdylib"]
> +crate-type = ["cdylib", "lib"]
> ```
> 
> that would already go a long way already and eliminate a bunch of hoops such as the entire `bindings` module dance in https://github.com/containers/krunvm/blob/main/src/bindings.rs, for example.

We can do this for 1.x until we do the full API revamp in 2.x. @mtjhrc WDYT?

mtjhrc:
> Just an idea, but from my point of view the main hurdle I have is that I cannot consume the library from Rust/Cargo directly, but have to go through an "installed" C library. If we could just have something like

Seems reasonable to me, user will have to compile the init.c themselves though (or provide an empty file)  https://github.com/containers/libkrun/pull/553 helps with that though. Ideally I we would build it in a `build.rs` I guess.

slp:
@d-e-s-o do you want to send a PR with your proposal?

d-e-s-o:
> @d-e-s-o do you want to send a PR with your proposal?

Yep, I can work on it. Thanks for the quick response.

ggoodman:
Would y'all be open to a PR that allows the init to be supplied as a memory range instead of baked in at build time?

slp:
> Would y'all be open to a PR that allows the init to be supplied as a memory range instead of baked in at build time?

Yes, please add it as a new function call in the API.

ggoodman:
@d-e-s-o, @slp here's what I had in mind: #593.

mtjhrc:
#718 has now landed in main branch :tada: and should cover this use case. Suggestions/PRs to improve the API are welcome :smiley:

PR #495, opened 2025-12-29
Fix external kernel memory region calculation

ericcurtin:
The external kernel payload type was not passing the firmware size to the architecture-specific memory region calculation function. This could lead to incorrect memory layout when using external kernels with firmware. The change ensures the firmware size is properly accounted for in the memory region setup. This inconsistency caused VM creation to fail with HV_DENIED (-85377017) on macOS when using direct kernel boot with the Hypervisor.framework.

### DIFF
--- a/src/vmm/src/builder.rs
+++ b/src/vmm/src/builder.rs
@@ -1450,9 +1450,13 @@ pub fn create_guest_memory(
                 };
             arch::arch_memory_regions(mem_size, Some(kernel_guest_addr), kernel_size, 0, None)
         }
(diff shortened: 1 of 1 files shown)


ericcurtin:
@slp @mtjhrc PTAL

slp:
TBH, I haven't considered the possibility of preloading a firmware AND a kernel, since usually the first is there just to load the second, but I guess there may be use cases where you want both (something we need to consider before dropping the EFI flavor, since the current API doesn't allow for it). Thanks @ericcurtin !

Issue #501, opened 2026-01-07
'error: expected expression' near 'char unicode[5];' in init/jsmn.h


mchassara:
Hi everyone,

While trying to compile libkrun through Homebrew, I noticed this error:

```
init/jsmn.h:241:17: error: expected expression
  241 |                 char unicode[5];
      |                 ^
init/jsmn.h:258:21: error: use of undeclared identifier 'unicode'
  258 |                     unicode[i] = js[parser->pos];
      |                     ^
init/jsmn.h:262:17: error: use of undeclared identifier 'unicode'
  262 |                 unicode[4] = '\0';
      |                 ^
init/jsmn.h:263:33: error: use of undeclared identifier 'unicode'
  263 |                 ascii = strtol(&unicode[0], NULL, 16);
```

[It seems](https://github.com/containers/libkrun/blob/main/init/jsmn.h#L240) that the whole statement for `case 'u':` is not properly considered as a statement, and variable declarations may break in the next lines of code.

Shouldn't the case be enclaused with braces, like so?

```c
    case 'u': {
        char unicode[5];
        long ascii;

        // [...]
    }
```

slp:
Fixed by #503

PR #503, opened 2026-01-07
Fix: Add block scope for variable declarations in switch case 'u'

mchassara:
## Problem

Related to https://github.com/containers/libkrun/issues/501

The `case 'u':` label in the JSON string parser's switch statement declares variables (`unicode` and `ascii`) directly after the case label. In C, this is invalid syntax because all case labels share the same scope, and variable declarations require a proper block scope.

## Solution

Wrapped the `case 'u':` block in curly braces `{ }` to create a proper scope for the variable declarations. This ensures the code compiles correctly and follows standard C syntax.

## Changes

- Added opening brace `{` after `case 'u':` on line 240
- Added closing brace `}` before the `default:` case on line 273

## Testing

The code should now compile without errors. The functionality remains unchanged - this is purely a syntax fix.

### DIFF
--- a/init/jsmn.h
+++ b/init/jsmn.h
@@ -237,7 +237,7 @@ static int jsmn_parse_string(jsmn_parser *parser, char *js, const size_t len,
             case 't':
                 break;
             /* Allows escaped symbol \uXXXX */
-            case 'u':
+            case 'u': {
                 char unicode[5];
                 long ascii;
 
@@ -270,6 +270,7 @@ static int jsmn_parse_string(jsmn_parser *parser, char *js, const size_t len,
                 parser->pos--;
                 js[parser->pos] = (char)ascii;
                 break;
+            }
             /* Unexpected symbol */
             default:
                 parser->pos = start;


slp (approved):
@mchassara thanks for the PR! It's interesting that some compilers doesn't complain about this. In any case, let's get this fixed.

mtjhrc:
Obviously one should always use a `{` `}` block to have sensible semantics for variables. I got a bit nerd-sniped why this was invalid though...

Turns out this is in fact valid C, but from C23 onwards 😆 (and it's also valid in C++)
In older versions the language expects a label to be followed by a statement (and variable declaration isn't a statement). 
So technically something like this is valid (after C99):
```
case 'u': {
    printf("I am a statment!");
    char unicode[5];
    long ascii;
    ...
}
```

We should probably specify a C version explicitly in the Makefile...

mchassara:
Thank you @mtjhrc for the investigation, I wasn't brave enough to check all the C versions and I've stopped coding C since few years now. 😅 

@slp For the record, I found the bug while trying to install https://github.com/nohajc/anylinuxfs (cc @nohajc)

nohajc:
@mchassara You're still running an older version of macOS, right? There's a libkrun bottle for Tahoe. I still provide one for Sequoia too. Otherwise it'll build from source.

mchassara:
> @mchassara You're still running an older version of macOS, right? There's a libkrun bottle for Tahoe. I still provide one for Sequoia too. Otherwise it'll build from source.

Running on Sonoma 14.8.3 indeed!

PR #504, opened 2026-01-08
Add debug logging for nitro enclaves

tylerfanelli:
This PR enables nitro enclave guest kernel and initramfs debug logs via `krun_set_log_level(KRUN_LOG_LEVEL_DEBUG)`. For those looking to try for themselves, a `--debug` flag was added to the nitro example. Enabling this flag will enable debug output from the enclave.

### DIFF
--- a/src/nitro/src/enclaves.rs
+++ b/src/nitro/src/enclaves.rs
@@ -21,12 +21,13 @@ use tar::HeaderMode;
 
 type Result<T> = std::result::Result<T, NitroError>;
 
-const ROOTFS_DIR_DENYLIST: [&str; 5] = [
+const ROOTFS_DIR_DENYLIST: [&str; 6] = [
     "proc", // /proc.
     "run",  // /run.
     "tmp",  // /tmp.
     "dev",  // /dev.
     "sys",  // /sys.
+    "usr/share/krun-nitro",
 ];
 
 /// Nitro Enclave data.


src/libkrun/src/lib.rs
```diff
@@ -403,16 +402,18 @@ impl TryFrom<ContextConfig> for NitroEnclave {
             return Err(-libc::EINVAL);
         };
 
+        let debug = KRUN_NITRO_DEBUG.lock().unwrap();
```
jakecorrenti, line 405: I think we can use [this macro](https://docs.rs/log/latest/log/macro.log_enabled.html) and avoid the mutex

jakecorrenti (commented):
Just one slight improvement that I think would improve overall readability

src/libkrun/src/lib.rs
tylerfanelli: This debug bool argument is used for the enclave, not libkrun itself. It's largely unrelated to the libkrun log level. It's a value representing if the **enclave** should be launched in debug mode. libkrun may be launched with a different log level (as I do for my testing, setting krun log level to error but launching the enclave in debug mode).

jakecorrenti: But isn't the only way you currently set the debug output to true is if the user selects the debug log level for all of libkrun?

If we want to keep the ability to have the debug level of the enclave different from the libkrun log level, then wouldn't we need to keep the start flags argument in the API?

tylerfanelli: Running libkrun with `RUST_LOG=...` can also achieve libkrun log levels AFAIU (at least, that's how I use it)

PR #513, opened 2026-01-13
virtio/fs/macos: keep a fd to unlinked files

slp:
On the macOS implementation of passthrough, we avoid keeping an FD open for each dirent entry by accessing the inode using the special directory '/.vol'. But this strategy doesn't work when the inode has been unlinked, and it's completely valid for userspace to keep accessing a file after unlinking it by keeping an FD to it.

To fix this without having to keep an FD open as the Linux implementation does (thus risking running out of open handles), here we change do_unlink() to open an FD and store it in InodeData->unlinked_fd. Then we change inode_to_path() to become inode_to_handle(), being able to return either a path accssible through '/.vol' or an FD. Finally, we update every user of inode_to_handle() to be able to operate both with a path and with an FD.

The FD stored in InodeData->unlinked_fd is released on forget_one(), which is called when the guest removes the entry from dirent.

Instead of wrapping the FD behind an OwnedFd or a File, we operate on it as an integer/RawFd. This allows us to 1) store the value as an atomic operation without a Mutex and 2) save a couple of syscalls everytime we use it.

### DIFF
--- a/src/devices/src/virtio/fs/macos/passthrough.rs
+++ b/src/devices/src/virtio/fs/macos/passthrough.rs
@@ -12,7 +12,7 @@ use std::mem::{self, MaybeUninit};
 use std::os::unix::io::{AsRawFd, FromRawFd, RawFd};
 use std::ptr::null_mut;
 use std::str::FromStr;
-use std::sync::atomic::{AtomicBool, AtomicI32, AtomicU64, Ordering};
+use std::sync::atomic::{AtomicBool, AtomicI32, AtomicI64, AtomicU64, Ordering};
 use std::sync::{Arc, Mutex, RwLock};
 use std::time::Duration;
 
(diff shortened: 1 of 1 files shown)


src/devices/src/virtio/fs/macos/passthrough.rs
```diff
@@ -478,8 +484,18 @@ impl PassthroughFs {
...
+        debug!("inode_to_handle: path={}", cstr.to_string_lossy());
+
+        let unlinked_fd = data.unlinked_fd.load(Ordering::Acquire);
+        if unlinked_fd >= 0 {
+            // Only use `unlinked_fd` when the inode path no longer exists.
+            let res = unsafe { libc::access(cstr.as_ptr(), libc::F_OK) };
+            if res != 0 {
+                return Ok(InodeHandle::Fd(unlinked_fd as RawFd));
+            }
+        }
```
mtjhrc: Why do you need to check if the `.vol` path exists here? If you already have a valid file descriptor can't you always just use that instead?

slp: In the initial iterations of this commit, many methods didn't work with file descriptors. Right now, only `link()` needs a path, so I've added a `supports_fd` boolean to `inode_to_handle` to deal with this particular method, and return `unlinked_fd` (if present) for every other one.

slp:
@mtjhrc PTAL

PR #514, opened 2026-01-13
devices: Refactor to make virtqueues fully created at transport layer

mtjhrc:


slp:
Code LGTM, but we're going to need some deep/long run testing. Please rebase it and let's merge at the beginning of the 1.18 development cycle to ensure it gets some testing before making in into a release.

PR #516, opened 2026-01-20
krun-sys: update bindgen dependency to v0.72.1

decathorpe:
It appears that this dependency was already updated in krun_display and krun_input, but not in krun-sys.

This is part of an upcoming effort to update packages that depend on bindgen in Fedora Linux to bindgen >= 0.72.1 to prevent breakage with LLVM 22 in Fedora 44+ (see https://fedoraproject.org/wiki/Changes/Bump_Minimum_Rust_Bindgen_To_v0.72 for context).

Note that there appears to be an issue with the current project layout (crates in subdirectories that *look* like workspace members but are not) - even running `cargo check` in the `krun-sys` subdirectory fails with recent Rust versions, so I'm not sure how this subproject is handled:

```
error: current package believes it's in a workspace when it's not:
current:   /home/deca/Workspace/forks/libkrun/krun-sys/Cargo.toml
workspace: /home/deca/Workspace/forks/libkrun/Cargo.toml

this may be fixable by adding `krun-sys` to the `workspace.members` array of the manifest located at: /home/deca/Workspace/forks/libkrun/Cargo.toml
Alternatively, to keep it out of the workspace, add the package to the `workspace.exclude` array, or add an empty `[workspace]` table to the package's manifest.
```

### DIFF
--- a/examples/Cargo.lock
+++ b/examples/Cargo.lock
@@ -86,38 +86,18 @@ checksum = "c08606f8c3cbf4ce6ec8e28fb0014a2c086708fe954eaa885384a6165172e7e8"
 
 [[package]]
 name = "bindgen"
-version = "0.69.5"
+version = "0.72.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "271383c67ccabffb7381723dea0672a673f292304fcb45c01cc648c7a8d58088"
+checksum = "993776b509cfb49c750f11b8f07a46fa23e0a1386ffc01fb1e7d343efc387895"
 dependencies = [
  "bitflags 2.9.1",
  "cexpr",
  "clang-sys",
- "itertools 0.12.1",
- "lazy_static",
- "lazycell",
+ "itertools",
  "proc-macro2",
  "quote",
  "regex",
- "rustc-hash 1.1.0",
- "shlex",
- "syn",
-]
-
-[[package]]
-name = "bindgen"
-version = "0.72.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4f72209734318d0b619a5e0f5129918b848c416e122a3c4ce054e03cb87b726f"
-dependencies = [
- "bitflags 2.9.1",
- "cexpr",
- "clang-sys",
- "itertools 0.13.0",
- "proc-macro2",
- "quote",
- "regex",
- "rustc-hash 2.1.1",
+ "rustc-hash",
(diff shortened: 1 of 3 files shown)


decathorpe:
It looks like the CI failed due to unrelated (?) issues. If they are *not* unrelated, then please tell me how to fix them. Because as far as I can tell, the current cargo "workspace but not actually workspace" setup is just broken. 🤷🏼

slp:
@decathorpe Thanks a lot for helping fix this. The DCO test is failing because the commit is missing a `Signed-off-by` signature (like the one added by `git commit -s`). The others are failing because `Cargo.lock` is out of date after updating `krun-sys/Cargo.toml`. Please add it to the commit too.

mtjhrc:
Thanks for the PR!

Yeah the workspace setup is confusing, but the setup is not broken per se, it could use some recognizing though. I guess we should at least exclude the `krun-sys` from the workspace explicitly.  

The CI failures are valid, the CI is failing because you haven't regenerated the Cargo.lock files to use the new dependency version.

decathorpe:
I have re-pushed with a sign-off commit.

> The others are failing because Cargo.lock is out of date after updating krun-sys/Cargo.toml.

I cannot do this. cargo refuses to update the lockfile (or do anything at all) with the error message I posted in my initial message.

mtjhrc:
> I cannot do this. cargo refuses to update the lockfile (or do anything at all) with the error message I posted in my initial message.

Note that, there shouldn't be a `Cargo.lock` inside the `krun-sys`. The problematic lock files are inside the `tests` and `examples` (these shouldn't be part of the workspace either, because they depend on libkrun being installed and available via pkg-config).

decathorpe:
> > I cannot do this. cargo refuses to update the lockfile (or do anything at all) with the error message I posted in my initial message.
> 
> Note that, there shouldn't be a `Cargo.lock` inside the `krun-sys`. The problematic lock files are inside the `tests` and `examples` (these shouldn't be part of the workspace either, because they depend on libkrun being installed and available via pkg-config).

Oh, I didn't realize that there were multiple Cargo.lock file in the repository ... that is quite cursed.

I've pushed again, this time updating the Cargo.lock files in examples/ and tests/.

slp:
Please run something like `cargo build` in both `examples` and `tests` to update their respective `Cargo.lock` files, and add them to the commit.

decathorpe:
Thanks!

PR #517, opened 2026-01-21
nitro: Add CID-specific vsock ports, shutdown status reader, signal handling proxy, device modularization

tylerfanelli:


init/nitro/main.c
```diff
@@ -369,12 +369,37 @@ static int nsm_exit(int nsm_fd)
...
+    unsigned int cid;
+    int ret, fd;
+
+    fd = open("/dev/vsock", O_RDONLY);
+    if (fd < 0) {
+        perror("unable to open /dev/vsock to fetch enclave CID:");
+        return 0;
+    }
+
+    ret = ioctl(fd, IOCTL_VM_SOCKETS_GET_LOCAL_CID, &cid);
```
jakecorrenti, line 287: non-blocker: we could just immediately close the fd after the ioctl call. would let us avoid writing it twice

src/nitro/src/enclaves.rs
```diff
@@ -30,6 +30,16 @@ const ROOTFS_DIR_DENYLIST: [&str; 6] = [
     "usr/share/krun-nitro",
 ];
 
+#[repr(u32)]
+pub enum VsockPortOffset {
+    ArgsReader = 1,
+    Net = 2,
+    AppOutput = 3,
+
+    // Not set by krun-nitro.
```
jakecorrenti: non-blocker: I think it would be helpful to put the link to the docs where you found this so it's easy for contributors to find the reference materials

src/nitro/src/device/mod.rs
```diff
@@ -0,0 +1,5 @@
+// SPDX-License-Identifier: Apache-2.0
+
+mod devices;
```
jakecorrenti: non-blocker: I'm not thrilled about the name since we're not really implementing any device emulation... I don't have a better name, though.

(`proxy` module maybe?)

src/nitro/src/enclave/args_writer.rs
```diff
@@ -21,7 +24,43 @@ pub struct EnclaveArgsWriter<'a> {
...
+            .map(|s| s.to_string())
+            .collect();
+
+        let envp: Vec<String> = envp_str
+            .replace("\"", "")
+            .split(' ')
+            .map(|s| s.to_string())
+            .collect();
+
+        args.append(&mut vec![
```
jakecorrenti, line 49: non-blocker: rather than defining `args` at the top of the function and then appending an entire vec right here, just wait to define the variable until after we define `argv` and `envp` and just do `let mut args = vec![...]`

init/nitro/main.c
```diff
@@ -524,20 +575,12 @@ int main(int argc, char *argv[])
...
-         * network proxy that the vsock and TUN file descriptors have been
-         * closed before exiting the parent process (i.e. the VM in general).
-         */
-        sleep(5);
+        ret = app_ret_write(ret_code, cid);
 
-        ret = app_status;
-        break;
+        // Allow the host to read the return code before exiting the enclave.
+        sleep(1);
```
jakecorrenti, line 557: non-blocker: wonder if there's a way to handle this without doing a sleep

PR #520, opened 2026-01-27
nitro: Reorganize and document modules

tylerfanelli:
cc/ @jakecorrenti

### DIFF
--- a/src/nitro/src/enclave/proxy/proxies/signal_handler.rs
+++ b/src/nitro/src/enclave/proxy/proxies/signal_handler.rs
@@ -6,7 +6,7 @@ use crate::enclave::{
 };
 use signal_hook::consts::SIGTERM;
 use std::{
-    io::{Read, Write},
+    io::{ErrorKind, Read, Write},
     sync::{
         atomic::{AtomicBool, Ordering},
         Arc,
@@ -61,9 +61,15 @@ impl DeviceProxy for SignalHandler {
         }
 
         let sig = libc::SIGTERM;
-        vsock.write(&sig.to_ne_bytes()).map_err(Error::VsockWrite)?;
-
-        Ok(0)
+        match vsock.write(&sig.to_ne_bytes()) {
+            Ok(size) => Ok(size),
+            /*
+             * If connection was already closed by enclave, return zero bytes written in order to
+             * listen for receiver shutdown signal.
+             */
+            Err(e) if e.kind() == ErrorKind::BrokenPipe => Ok(0),
+            Err(e) => Err(Error::VsockWrite(e)),
+        }
     }
 
     /// Establish the proxy's vsock connection.


src/nitro/src/enclave/device/devices/net.rs
```diff
@@ -102,12 +102,12 @@ impl DeviceProxy for NetProxy {
                                 if e == RecvTimeoutError::Timeout {
                                     continue;
                                 } else {
-                                    panic!();
+                                    return Err(Error::ShutdownSignalReceive(e))?;
                                 }
```
jakecorrenti:
```suggestion
                                if e != RecvTimeoutError::Timeout {
                                    return Err(Error::ShutdownSignalReceive(e))?;
                                }
```

src/nitro/src/enclave/device/devices/net.rs
```diff
@@ -67,18 +67,18 @@ impl DeviceProxy for NetProxy {
                         .write_all(&vsock_buf[..size])
                         .map_err(Error::UnixWrite)?;
                 } else {
-                    tx.send(()).unwrap();
+                    let _ = tx.send(());
```
jakecorrenti: Any particular reason why we want to ignore the error here?

src/nitro/src/enclave/device/devices/net.rs
jakecorrenti: maybe something along these lines
```suggestion
if size < 1 {
    break;
}
                        
if let Err(e) = vsock_stream.0.write_all(&unix_buf[..size])) {
    log::error("failed to write bytes over vsock: {}", e);
}
```

src/nitro/src/enclave/mod.rs
```diff
@@ -89,30 +90,35 @@ impl NitroEnclave {
...
         }
 
         Ok(())
     }
 
-    fn start(&mut self) -> Result<(u32, PollTimeout)> {
-        let eif = eif()?;
+    fn start(&mut self) -> Result<(u32, PollTimeout), StartError> {
+        let path = env::var("KRUN_NITRO_EIF_PATH")
+            .unwrap_or("/usr/share/krun-nitro/krun-nitro.eif".to_string());
```
jakecorrenti: These hard-coded strings should be constants somewhere

src/nitro/src/enclave/mod.rs
```diff
@@ -123,16 +129,16 @@ impl NitroEnclave {
...
-            .map_err(NitroError::VmStart)?;
+            .map_err(StartError::VmStart)?;
 
-        Ok((cid.try_into().unwrap(), timeout)) // Safe to unwrap.
+        // Safe to unwrap.
+        Ok((cid.try_into().unwrap(), timeout))
     }
 
-    fn devices(&self) -> Result<DeviceProxyList> {
+    fn devices(&self) -> Result<DeviceProxyList, device::Error> {
```
jakecorrenti: above you alias the other modules' error types. Here I would either alias this as `DeviceError` or change the others to use the module path, i.e. `start::Error` instead of `StartError`

src/nitro/src/enclave/mod.rs
```diff
@@ -144,7 +150,7 @@ impl NitroEnclave {
         Ok(DeviceProxyList(proxies))
     }
 
-    fn rootfs_archive(&self) -> Result<Vec<u8>> {
+    fn rootfs_archive(&self) -> result::Result<Vec<u8>, io::Error> {
```
jakecorrenti: s/`result::Result`/`Result`

src/libkrun/src/lib.rs
```diff
@@ -381,13 +381,12 @@ impl TryFrom<ContextConfig> for NitroEnclave {
...
-                            error!("unable to configure network device: {:?}", e);
-                            return Err(-libc::EINVAL);
-                        }
-                    }
+                    let fd = match device.cfg_backend {
+                        VirtioNetBackend::UnixstreamFd(fd) => RawFd::from(fd),
+                        _ => return Err(libc::EINVAL),
+                    };
+
+                    Some(fd)
```
jakecorrenti, line 389:
```suggestion
                    match device.cfg_backend {
                        VirtioNetBackend::UnixstreamFd(fd) => Some(RawFd::from(fd)),
                        _ => Err(libc::EINVAL),
                    }
```

src/nitro/src/enclave/device/devices/net.rs
```diff
@@ -5,129 +5,84 @@ use crate::enclave::{
...
-    thread::{self, JoinHandle},
     time::Duration,
 };
-use vsock::{VsockAddr, VsockListener, VMADDR_CID_ANY};
+use vsock::{VsockAddr, VsockListener, VsockStream, VMADDR_CID_ANY};
 
-#[derive(Clone)]
 pub struct NetProxy {
-    fd: RawFd,
+    buf: [u8; 1500],
```
jakecorrenti: What's the significance of the number 1500 here?

src/nitro/src/enclave/proxy/proxies/net.rs
```diff
@@ -5,129 +5,84 @@ use crate::enclave::{
...
-            .map_err(Error::VsockBind)?;
-
-        let mut vsock_stream = vsock_listener.accept().map_err(Error::VsockAccept)?;
-
-        let mut vsock_stream_clone = vsock_stream.0.try_clone().map_err(Error::VsockClone)?;
+        Ok(Some(Box::new(Self {
+            buf: self.buf,
+            unix,
+        })))
+    }
```
jakecorrenti, line 56: missing the whitespace between method definitions in this commit

src/nitro/src/enclave/proxy/proxies/net.rs
```diff
@@ -5,129 +5,84 @@ use crate::enclave::{
...
 
-        // vsock
-        let vsock_thread: JoinHandle<Result<()>> = thread::spawn(move || {
-            let mut vsock_buf = [0u8; 1500];
-            loop {
-                let size = vsock_stream_clone
-                    .read(&mut vsock_buf)
-                    .map_err(Error::VsockRead)?;
+        Ok(size)
+    }
```
jakecorrenti, line 68: missing the whitespace between method definitions in this commit

src/nitro/src/enclave/proxy/proxies/net.rs
```diff
@@ -5,129 +5,84 @@ use crate::enclave::{
...
+        match self.unix.read(&mut self.buf) {
+            Ok(size) => {
                 if size > 0 {
-                    unix_stream_clone_write
-                        .write_all(&vsock_buf[..size])
-                        .map_err(Error::UnixWrite)?;
-                } else {
-                    let _ = tx.send(());
-                    break;
+                    let _ = vsock.write_all(&self.buf[..size]);
```
jakecorrenti, line 75: should we at least log any sorts of errors returned from the write op?

src/nitro/src/enclave/proxy/proxies/output.rs
```diff
@@ -18,6 +18,7 @@ type Result<T> = std::result::Result<T, Error>;
 pub struct OutputProxy {
     file: File,
     debug: bool,
+    buf: [u8; 1500],
```
jakecorrenti: Seems like we've used the buffer size 1500 in multiple spots now; we should make this a constant so we only have to make changes to the value in one spot if necessary.

src/nitro/src/enclave/device/devices/output.rs
```diff
@@ -28,54 +29,54 @@ impl OutputProxy {
...
+        let buf = [0u8; 1500];
+
+        Ok(Self { file, debug, buf })
     }
 }
 
 impl DeviceProxy for OutputProxy {
+    fn clone(&self) -> Result<Option<Box<dyn DeviceProxy>>> {
+        Ok(None)
+    }
```
jakecorrenti: missing white space between all method definitions in this commit

src/nitro/src/enclave/proxy/proxies/output.rs
```diff
@@ -28,54 +29,54 @@ impl OutputProxy {
...
+        }
 
-    fn _start(&mut self, vsock_port: u32) -> Result<()> {
-        let mut vsock_stream = if self.debug {
-            VsockStream::connect(&VsockAddr::new(VMADDR_CID_HYPERVISOR, vsock_port))
+        Ok(size)
+    }
+    fn send(&mut self, _vsock: &mut VsockStream) -> Result<usize> {
+        Ok(0)
+    }
```
jakecorrenti, line 79: we could probably make this the default implementation of the method in the trait definition

src/nitro/src/enclave/device/devices/signal_handler.rs
```diff
@@ -9,77 +9,58 @@ use std::{
...
+        signal_hook::flag::register(SIGTERM, Arc::clone(&sig)).map_err(Error::SignalRegister)?;
+
+        let buf = [0u8; 1];
+
+        Ok(Self { sig, buf })
+    }
+}
 
 impl DeviceProxy for SignalHandler {
+    fn clone(&self) -> Result<Option<Box<dyn DeviceProxy>>> {
```
jakecorrenti: missing whitespace between all methods in this commit

src/nitro/src/enclave/proxy/proxies/signal_handler.rs
```diff
@@ -9,77 +9,58 @@ use std::{
...
 
-            Ok(())
-        });
+        let sig = libc::SIGTERM;
+        vsock.write(&sig.to_ne_bytes()).map_err(Error::VsockWrite)?;
 
-        if let Ok(Err(e)) = signal_handler.join() {
-            log::error!("error in signal handler proxy: {e}");
-        }
+        Ok(0)
```
jakecorrenti: Should we return the bytes written instead of 0?

src/nitro/src/enclave/proxy/mod.rs
```diff
@@ -7,32 +7,81 @@ pub use devices::*;
...
+                                        } else {
+                                            return Err(Error::ShutdownSignalReceive(e))?;
+                                        }
+                                    }
+                                }
+                            }
+                        }
+                    }
+
+                    Ok(())
```
jakecorrenti, line 114: Could do something like this to reduce the nesting 
```suggestion
                    let mut sender = clone.unwrap_or(return Ok(()));
                    loop {
                        let size = sender.send(&mut vsock_send)?;
                        if size != 0 {
                            continue;
                        }
                        match rx.recv_timeout(Duration::from_micros(500)) {
                            Ok(_) => break,
                            Err(e) => {
                                if e != RecvTimeoutError::Timeout {
                                    return Err(Error::ShutdownSignalReceive(e))?;
                                }
                            }
                        }
                    }

                    Ok(())
```

src/nitro/src/enclave/proxy/proxies/net.rs
```diff
@@ -35,6 +35,9 @@ impl TryFrom<RawFd> for NetProxy {
 }
 
 impl DeviceProxy for NetProxy {
+    fn arg(&self) -> Option<EnclaveArg<'_>> {
+        Some(EnclaveArg::NetworkProxy)
+    }
```
jakecorrenti, line 46: missing whitespace between method in this commit

src/nitro/src/enclave/proxy/proxies/output.rs
```diff
@@ -36,15 +36,15 @@ impl OutputProxy {
...
-        Ok(None)
-    }
-    fn enclave_arg(&self) -> Option<EnclaveArg<'_>> {
+    fn arg(&self) -> Option<EnclaveArg<'_>> {
         match self.debug {
             true => Some(EnclaveArg::Debug),
             false => None,
         }
     }
+    fn clone(&self) -> Result<Option<Box<dyn DeviceProxy>>> {
```
jakecorrenti, line 60: missing the whitespace between the method in this commit.

Also: we could make this the default implementation of the method in the trait definition, that way it's reusable for other implementations and we only have to write it once.

src/nitro/src/enclave/proxy/proxies/signal_handler.rs
```diff
@@ -32,12 +32,12 @@ impl SignalHandler {
 }
 
 impl DeviceProxy for SignalHandler {
+    fn arg(&self) -> Option<EnclaveArg<'_>> {
+        None
```
jakecorrenti, line 42: Whitespace in this commit as well.

Same comment as above w/r/t this being the default implementation for the trait

src/nitro/src/enclave/mod.rs
```diff
@@ -58,10 +60,11 @@ pub struct NitroEnclave {
 }
 
 impl NitroEnclave {
-    /// Run the enclave.
+    //// Run an application within a nitro enclave.
```
jakecorrenti: extra `/`

init/nitro/main.c
```diff
@@ -340,6 +340,13 @@ static int app_ret_write(int code, unsigned int cid)
...
+    ret = read(sock_fd, (void *)&code, sizeof(int));
+    if (ret < sizeof(int)) {
+        perror("unable to read close signal from application return vsock");
+        close(sock_fd);
+        return -errno;
+    }
+
     close(sock_fd);
 
     return 0;
```
jakecorrenti, line 324: this is totally acceptable as is, but you could also do this I believe
```suggestion
    ret = read(sock_fd, (void *)&code, sizeof(int));
    if (ret < sizeof(int))
        perror("unable to read close signal from application return vsock");

    close(sock_fd);
    return -errno;
```

src/nitro/src/enclave/proxy/proxies/net.rs
tylerfanelli: Addressed in later commit.

src/nitro/src/enclave/device/devices/net.rs
tylerfanelli: Sometimes the sender thread closes the connection before the message is received. In this case an error is thrown, although it's not necessarily an error.

src/nitro/src/enclave/device/devices/net.rs
tylerfanelli: Addressed in later commit.

src/nitro/src/enclave/device/devices/net.rs
tylerfanelli: Addressed in later commit.

src/nitro/src/enclave/proxy/proxies/signal_handler.rs
tylerfanelli: Addressed in later commit.

src/nitro/src/enclave/mod.rs
tylerfanelli: Addressed in https://github.com/containers/libkrun/pull/520/commits/9058a5a812b58c2be99d5aae5690e401b58b3898

src/nitro/src/enclave/mod.rs
tylerfanelli: Addressed in https://github.com/containers/libkrun/pull/520/changes/f9d8260eac5be96ea942726fe4f2a9a703e50fd5

src/libkrun/src/lib.rs
tylerfanelli: This doesn't immediately return the error, which is what we would like to do.

src/nitro/src/enclave/device/devices/net.rs
tylerfanelli: It represents the ethernet MTU for the enclave VM. At some point, I'll likely make an addition that the enclave will communicate this value to the host.

src/nitro/src/enclave/proxy/proxies/net.rs
tylerfanelli: I don't think so, because it largely only happens when the proxy writes something after the enclave has already closed the vsock. This isn't really an error per se. We expect that the enclave closed the socket for a reason (exiting the VM, for example).

src/nitro/src/enclave/proxy/proxies/output.rs
tylerfanelli: It's not going to be a constant in the network proxy at some point, so I think that can be addressed at a later point.

src/nitro/src/enclave/proxy/proxies/output.rs
tylerfanelli: I'd rather make it explicit and clear when writing each device proxy.

src/nitro/src/enclave/proxy/proxies/output.rs
tylerfanelli: I'd rather make it explicit and clear when writing each device proxy. This way, a comment can be made exactly **why** there is no send method.

src/nitro/src/enclave/proxy/proxies/signal_handler.rs
tylerfanelli: In this case no, because we'd like the proxy to exit the send method immediately after this (there is nothing else left to write).

init/nitro/main.c
tylerfanelli: I don't believe we can make an assumption of  the value of errno at that point. Or at least it looks a bit more confusing to me, even if we assume errno is `0`.

src/nitro/src/enclave/proxy/mod.rs
tylerfanelli: Clippy complained about this approach because of:
```
let mut sender = clone.unwrap_or(return Ok(()));
```
If that unwrap is taken, clippy warns that the rest of the variables/methods are unused. This persists even if I use
```
#[allow(unused_variables)]
```
With this, I think it's fine to leave the nesting as is at the moment.

src/nitro/src/enclave/proxy/proxies/signal_handler.rs
tylerfanelli: I've instead addressed this in this commit: https://github.com/containers/libkrun/pull/520/commits/84fb54e852d9a93e431edbb2ce3f4756aa89b90f

Essentially, we just want the signal handling proxy to return zero bytes written (and wait for a receiver timeout signal) if the vsock was already closed by the enclave.

src/libkrun/src/lib.rs
jakecorrenti: yep, you're right. Apologies.

PR #521, opened 2026-02-02
Nitro reformatting + rootfs writing update

tylerfanelli:


tylerfanelli:
cc @jakecorrenti

jakecorrenti:
It was my understanding that we used "nitro", "tdx", and "cca" as the CC feature names to avoid verbosity introduced by SNP ("amd-snp"). Has this changed? In other words, should CCA and TDX get renamed as well? 

I think that we need to come to a consensus on the naming scheme that we want to use here

tylerfanelli:
Those acronyms have specific meanings more than "nitro" does. I think they can remain the same.

init/aws-nitro/fs.c
```diff
@@ -126,9 +126,45 @@ int filesystem_init()
...
+    if (ret < 0) {
+        perror("mkdir /sys");
+        return -errno;
+    }
+
+    ret = mkdir("/tmp", 0755);
+    if (ret < 0) {
+        perror("mkdir /tmp");
+        return -errno;
+    }
```
jakecorrenti:
```suggestion
    char *dirs[] = {"/dev", "/proc", "run", "/sys", "/tmp"};
    for (size_t i = 0; i <  sizeof(dirs) / sizeof(dirs[0]); i++) {
        char *dir = dirs[i];
        ret = mkdir(dir, 0755);
        if (ret < 0) {
            fprintf(stderr, "mkdir %s: %s\n", dir, strerror(errno));
            return -errno;
        }
    }

File: init/aws-nitro/fs.c [hunk 2]
```diff
@@ -156,13 +160,47 @@ int filesystem_init()
...
         perror("mount /dev");
         return -errno;
     }
 
-    // Create the /proc filesystem.
+    /*
+     * Initialize the /dev/shm used for shared memory between running
+     * applications.
+     */
+    ret = mkdir("/dev/shm", 0755);
```
jakecorrenti: can we create these directories at the same time as the ones above? Could tie it into the suggestion I made previously with the array of dirs.

src/aws_nitro/src/enclave/mod.rs
```diff
@@ -18,23 +18,23 @@ use std::{
...
     os::fd::RawFd,
-    path::PathBuf,
+    path::{Path, PathBuf},
 };
 use tar::HeaderMode;
 use vsock::{VsockAddr, VsockListener, VMADDR_CID_ANY};
 
 const KRUN_NITRO_EIF_PATH_ENV_VAR: &str = "KRUN_NITRO_EIF_PATH";
-const KRUN_NITRO_EIF_PATH_DEFAULT: &str = "/usr/share/krun-nitro/krun-nitro.eif";
+const KRUN_NITRO_EIF_PATH_DEFAULT: &str = "/krun-awsnitro/krun-awsnitro.eif";
```
jakecorrenti, line 27: We shouldn't be putting anything in the root directory. The XDG standard says cached/non-essential data for a user application should go in `$HOME/.cache/krun-awsnitro/`

tylerfanelli: It's only within the root directory for container images, and even then it is only used to access the cached EIF, not writing to the application itself. If a user wants to store it elsewhere, they can specify the path with `KRUN_NITRO_EIF_PATH=...`

Putting it in the root directory makes writing the rootfs much simpler when running inside of a container environment.

init/aws-nitro/fs.c
tylerfanelli: They cannot be created until `/dev` is mounted.

src/aws_nitro/src/enclave/mod.rs
jakecorrenti: I'm not sure I understand. I thought, in this PR's state, either the `nitro-cli` or `krun-awsnitro-eif-ctl` needs to create an EIF file and either put it at `/krun-awsnitro` or have the user put it there in order for libkrun to find it by default so that we can run the enclave.

In that case we'd still be putting application-specific data in the root directory of the parent EC2 instance, or at least making the assumption that something could be there

tylerfanelli: That is the default location, which should only be used when sharing a volume with a container. For running on a host EC2 instance, you can specify `KRUN_NITRO_EIF_PATH=...` and it will instead use that path.

You aren't required to actually put anything in the EC2 host instances root directory, but for container images you can share the volume at that directory and ensure it is not written within the enclave itself.

PR #525, opened 2026-02-03
Introduce krun_has_feature() API

mtjhrc:
Introduce a function to query at runtime whether a specific feature was enabled at build time. This allows applications to check for optional capabilities before attempting to use them. This is especially useful for our tests and examples.

### DIFF
--- a/include/libkrun.h
+++ b/include/libkrun.h
@@ -977,6 +977,35 @@ int32_t krun_set_nested_virt(uint32_t ctx_id, bool enabled);
  */
 int32_t krun_check_nested_virt(void);
 
+/* Feature constants for krun_has_feature() */
+#define KRUN_FEATURE_NET 0
+#define KRUN_FEATURE_BLK 1
+#define KRUN_FEATURE_GPU 2
+#define KRUN_FEATURE_SND 3
+#define KRUN_FEATURE_INPUT 4
+#define KRUN_FEATURE_EFI 5
+#define KRUN_FEATURE_TEE 6
+#define KRUN_FEATURE_AMD_SEV 7
+#define KRUN_FEATURE_INTEL_TDX 8
+#define KRUN_FEATURE_AWS_NITRO 9
+#define KRUN_FEATURE_VIRGL_RESOURCE_MAP2 10
+
+/**
+ * Checks if a specific feature was enabled at build time.
+ *
+ * Arguments:
+ *  "feature" - one of the KRUN_FEATURE_* constants.
+ *
+ * Returns:
+ *  1 if the feature is supported, 0 if not supported, or a negative error
+ *  number on failure (e.g., -EINVAL for invalid/unknown feature constant).
+ *
+ * Notes:
+ *  When linking against an older version of libkrun, this function may
+ *  return -EINVAL for feature constants that were added in newer versions.
+ */
+int32_t krun_has_feature(uint64_t feature);
+
 /**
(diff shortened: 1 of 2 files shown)


mtjhrc:
Is this even a good idea? I mean an alternative could be checking if the related symbol names exists with dlysm, but we are not always constituent on whether the function exists or just returns ENOTSUP or doesn't even return any error at all e.g. krun_set_gpu_options

src/libkrun/src/lib.rs
```diff
@@ -1844,6 +1844,36 @@ pub unsafe extern "C" fn krun_check_nested_virt() -> i32 {
...
+        "NET" => cfg!(feature = "net"),
+        "BLK" => cfg!(feature = "blk"),
+        "GPU" => cfg!(feature = "gpu"),
+        "SND" => cfg!(feature = "snd"),
+        "INPUT" => cfg!(feature = "input"),
+        "EFI" => cfg!(feature = "efi"),
+        "TEE" => cfg!(feature = "tee"),
+        "SEV" => cfg!(feature = "amd-sev"),
+        "TDX" => cfg!(feature = "tdx"),
+        "NITRO" => cfg!(feature = "nitro"),
```
tylerfanelli: #521 changes `NITRO` to `AWS_NITRO`, perhaps that could be modified here?

include/libkrun.h
```diff
@@ -977,6 +977,20 @@ int32_t krun_set_nested_virt(uint32_t ctx_id, bool enabled);
  */
 int32_t krun_check_nested_virt(void);
 
+/**
+ * Checks if a specific feature was enabled at build time.
+ *
+ * Arguments:
+ *  "feature" - a null-terminated string representing the feature name.
+ *              Supported feature names: "NET", "BLK", "GPU", "SND", "INPUT",
+ *              "EFI", "TEE", "SEV", "TDX", "NITRO", "VIRGL_RESOURCE_MAP2"
```
tylerfanelli:
```suggestion
 *              "EFI", "TEE", "SEV", "TDX", "AWS_NITRO", "VIRGL_RESOURCE_MAP2"
```

tylerfanelli (changes requested):

src/libkrun/src/lib.rs
mtjhrc: Yeah, that is nicer.

Hmm, perhaps we should we also change these for consistency?
`TDX` -> `INTEL_TDX`
`SEV` -> `AMD_SEV`

I mean we should probably change the name in the Makefile too, but this seems like a more user friendly naming convention.

mtjhrc:
> Rather than using strings, I wonder if we should do something similar to KVM where we have something along these lines:

Yeah, I've changed it to use integers.

include/libkrun.h
```diff
@@ -977,6 +977,35 @@ int32_t krun_set_nested_virt(uint32_t ctx_id, bool enabled);
...
+ *
+ * Returns:
+ *  1 if the feature is supported, 0 if not supported, or a negative error
+ *  number on failure (e.g., -EINVAL for invalid/unknown feature constant).
+ *
+ * Notes:
+ *  When linking against an older version of libkrun, this function may
+ *  return -EINVAL for feature constants that were added in newer versions.
+ */
+int krun_has_feature(uint64_t feature);
```
tylerfanelli:
```suggestion
int32_t krun_has_feature(uint64_t feature);
```

PR #527, opened 2026-02-04
Add vhost-user support with RNG device implementation

dorindabassey:
This PR adds vhost-user frontend support to libkrun, enabling virtio devices to run in separate processes using rust-vmm's vhost-device backends for improved isolation and flexibility. The RNG frontend is implemented as the initial use case, and is designed to easily support additional devices (sound, GPU, can, etc).

### DIFF
--- a/include/libkrun.h
+++ b/include/libkrun.h
@@ -718,6 +718,76 @@ int krun_add_input_device_fd(uint32_t ctx_id, int input_fd);
 int32_t krun_set_snd_device(uint32_t ctx_id, bool enable);
 
 /**
+ * Vhost-user device types.
+ * These correspond to virtio device type IDs for devices.
+ */
+#define KRUN_VIRTIO_DEVICE_RNG 4
+#define KRUN_VIRTIO_DEVICE_SND 25
+#define KRUN_VIRTIO_DEVICE_CAN 36
+
+/**
+ * Vhost-user RNG device default queue configuration.
+ * Use these when you want explicit defaults instead of auto-detection.
+ */
+#define KRUN_VHOST_USER_RNG_NUM_QUEUES 1
+#define KRUN_VHOST_USER_RNG_QUEUE_SIZES ((uint16_t[]){256})
+
+/**
+ * Add a vhost-user device to the VM.
+ *
+ * This function adds a vhost-user device by connecting to an external
+ * backend process (e.g., vhost-device-rng, vhost-device-snd). The backend
+ * must be running and listening on the specified socket before starting the VM.
+ *
+ * This API is designed for devices like RNG, sound, and CAN.
+ *
+ * Arguments:
+ *  "ctx_id"       - the configuration context ID.
+ *  "device_type"  - type of vhost-user device (e.g., KRUN_VHOST_USER_DEVICE_RNG).
+ *  "socket_path"  - path to the vhost-user Unix domain socket (e.g., "/tmp/vhost-rng.sock").
+ *  "name"         - device name for logging/debugging (e.g., "vhost-rng", "vhost-snd").
+ *                   NULL = auto-generate from device_type ("vhost-user-4", "vhost-user-25", etc.)
+ *  "num_queues"   - number of virtqueues.
+ *                   0 = auto-detect from backend (requires backend MQ support).
+ *                   >0 = explicit queue count.
+ *                   Or use device-specific constants like KRUN_VHOST_USER_RNG_NUM_QUEUES.
+ *  "queue_sizes"  - array of queue sizes for each queue.
+ *                   NULL = use default size (256) for all queues.
+ *                   When num_queues=0 (auto-detect): array must be 0-terminated (sentinel).
+ *                   When num_queues>0 (explicit): array must have exactly num_queues elements.
+ *                   Use device-specific constants like KRUN_VHOST_USER_RNG_QUEUE_SIZES for defaults.
+ *
+ * Examples:
+ *  // Auto-detect queue count, use default size (256)
+ *  krun_add_vhost_user_device(ctx, KRUN_VHOST_USER_DEVICE_RNG, "/tmp/rng.sock", NULL, 0, NULL);
+ *
+ *  // Auto-detect queue count, use custom size (512) for all queues
+ *  uint16_t custom_size[] = {512, 0};  // 0 = sentinel terminator
+ *  krun_add_vhost_user_device(ctx, KRUN_VHOST_USER_DEVICE_RNG, "/tmp/rng.sock", NULL, 0, custom_size);
+ *
+ *  // Explicit defaults using #define constants
+ *  krun_add_vhost_user_device(ctx, KRUN_VHOST_USER_DEVICE_RNG, "/tmp/rng.sock", "vhost-rng",
+ *                             KRUN_VHOST_USER_RNG_NUM_QUEUES,
+ *                             KRUN_VHOST_USER_RNG_QUEUE_SIZES);
+ *
+ *  // Explicit queue count with custom sizes
+ *  uint16_t sizes[] = {256, 512};
+ *  krun_add_vhost_user_device(ctx, KRUN_VHOST_USER_DEVICE_SND, "/tmp/snd.sock", "vhost-snd", 2, sizes);
+ *
+ * Returns:
+ *  Zero on success or a negative error number on failure.
+ *  -EINVAL  - Invalid parameters (device_type, socket_path, etc.)
+ *  -ENOENT  - Context doesn't exist
+ *  -ENOTSUP - vhost-user support not compiled in
+ */
+int32_t krun_add_vhost_user_device(uint32_t ctx_id,
+                                   uint32_t device_type,
+                                   const char *socket_path,
+                                   const char *name,
+                                   uint16_t num_queues,
+                                   const uint16_t *queue_sizes);
+
+/**
  * Configures a map of rlimits to be set in the guest before starting the isolated binary.
  *
  * Arguments:
--- a/src/devices/Cargo.toml
+++ b/src/devices/Cargo.toml
@@ -18,6 +18,7 @@ input = ["zerocopy", "krun_input"]
 virgl_resource_map2 = []
 aws-nitro = []
 test_utils = []
+vhost-user = ["vhost"]
 
 [dependencies]
 bitflags = "1.2.0"
@@ -29,6 +30,7 @@ nix = { version = "0.30.1", features = ["ioctl", "net", "poll", "socket", "fs"] 
 pw = { package = "pipewire", version = "0.8.0", optional = true }
 rand = "0.9.2"
 thiserror = { version = "2.0", optional = true }
+vhost = { version = "0.14", optional = true, features = ["vhost-user-frontend"] }
 virtio-bindings = "0.2.0"
 vm-memory = { version = "0.17", features = ["backend-mmap"] }
 zerocopy = { version = "0.8.26", optional = true, features = ["derive"] }
--- a/src/devices/Cargo.toml
+++ b/src/devices/Cargo.toml
@@ -30,7 +30,7 @@ nix = { version = "0.30.1", features = ["ioctl", "net", "poll", "socket", "fs"] 
 pw = { package = "pipewire", version = "0.8.0", optional = true }
 rand = "0.9.2"
 thiserror = { version = "2.0", optional = true }
-vhost = { version = "0.14", optional = true, features = ["vhost-user-frontend"] }
+vhost = { version = "0.16", optional = true, features = ["vhost-user-frontend"] }
 virtio-bindings = "0.2.0"
 vm-memory = { version = "0.17", features = ["backend-mmap"] }
 zerocopy = { version = "0.8.26", optional = true, features = ["derive"] }
@@ -50,8 +50,8 @@ lru = ">=0.9"
 [target.'cfg(target_os = "linux")'.dependencies]
 rutabaga_gfx = { path = "../rutabaga_gfx", features = ["x"], optional = true }
 caps = "0.5.5"
-kvm-bindings = { version = "0.12", features = ["fam-wrappers"] }
-kvm-ioctls = "0.22"
+kvm-bindings = { version = "0.14", features = ["fam-wrappers"] }
+kvm-ioctls = "0.24"
 
 [target.'cfg(any(target_arch = "aarch64", target_arch = "riscv64"))'.dependencies]
 vm-fdt = ">= 0.2.0"
--- a/src/devices/src/virtio/vhost_user/device.rs
+++ b/src/devices/src/virtio/vhost_user/device.rs
@@ -1,0 +1,548 @@
+// Copyright 2026, Red Hat Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Generic vhost-user device wrapper.
+//!
+//! This module provides a wrapper around the vhost crate's Frontend,
+//! adapting it to work with libkrun's VirtioDevice trait.
+
+use std::io::{self, ErrorKind, Result as IoResult};
+use std::os::fd::AsRawFd;
+use std::os::unix::net::UnixStream;
+use std::sync::{Arc, Mutex};
+
+use log::{debug, error, warn};
+use polly::event_manager::{EventManager, Subscriber};
+use utils::epoll::{EpollEvent, EventSet};
+use utils::eventfd::{EventFd, EFD_NONBLOCK};
+use vhost::vhost_user::{Frontend, VhostUserFrontend, VhostUserProtocolFeatures};
+use vhost::{VhostBackend, VhostUserMemoryRegionInfo, VringConfigData};
+use vm_memory::{Address, GuestMemory, GuestMemoryMmap, GuestMemoryRegion};
+
+use crate::virtio::{
+    ActivateError, ActivateResult, DeviceQueue, DeviceState, InterruptTransport, QueueConfig,
+    VirtioDevice,
+};
+
+/// VHOST_USER_F_PROTOCOL_FEATURES (bit 30) is a backend-only feature
+/// that enables vhost-user protocol extensions. It's not a virtio feature.
+const VHOST_USER_F_PROTOCOL_FEATURES: u64 = 1 << 30;
+
+/// Generic vhost-user device wrapper.
+///
+/// This wraps a vhost-user backend connection and implements the VirtioDevice
+/// trait, allowing it to be used like any other virtio device in libkrun.
+pub struct VhostUserDevice {
+    /// Vhost-user frontend connection
+    frontend: Arc<Mutex<Frontend>>,
+
+    /// Device type (e.g., VIRTIO_ID_RNG = 4)
+    device_type: u32,
+
+    /// Device name for logging
+    device_name: String,
+
+    /// Queue configurations
+    queue_configs: Vec<QueueConfig>,
+
+    /// Available features from the backend
+    avail_features: u64,
+
+    /// Whether the backend supports protocol features
+    has_protocol_features: bool,
+
+    /// Acknowledged features
+    acked_features: u64,
+
+    /// Device state
+    device_state: DeviceState,
+
+    /// Activation event (registered with EventManager)
+    activate_evt: EventFd,
+
+    /// Vring call event (backend->VMM interrupt notification)
+    vring_call_event: Option<EventFd>,
+}
+
+impl VhostUserDevice {
+    /// Create a new vhost-user device by connecting to a socket.
+    ///
+    /// # Arguments
+    ///
+    /// * `socket_path` - Path to the vhost-user Unix domain socket
+    /// * `device_type` - Virtio device type ID
+    /// * `device_name` - Human-readable device name for logging
+    /// * `num_queues` - Number of queues (0 = query backend via MQ protocol)
+    /// * `queue_sizes` - Size for each queue (empty = use default 256)
+    ///
+    /// # Returns
+    ///
+    /// A new VhostUserDevice or an error if connection fails.
+    pub fn new(
+        socket_path: &str,
+        device_type: u32,
+        device_name: String,
+        num_queues: u16,
+        queue_sizes: &[u16],
+    ) -> IoResult<Self> {
+        debug!("Connecting to vhost-user backend at {}", socket_path);
+
+        // Connect to the vhost-user backend
+        let stream = UnixStream::connect(socket_path)?;
+        // NOTE: `num_queues` could be 0 here, but this is actually fine
+        // because if `VhostUserProtocolFeatures::MQ` is supported the negotiated
+        // value will be used automatically by Frontend
+        let mut frontend = Frontend::from_stream(stream, num_queues as u64);
+
+        // Get available features from backend
+        let avail_features = frontend
+            .get_features()
+            .map_err(|e| io::Error::new(ErrorKind::Other, e))?;
+
+        debug!("{}: backend features: 0x{:x}", device_name, avail_features);
+
+        // Strip the vhost specific bit to leave only standard virtio features
+        let has_protocol_features = avail_features & VHOST_USER_F_PROTOCOL_FEATURES != 0;
+        let avail_features = avail_features & !VHOST_USER_F_PROTOCOL_FEATURES;
+
+        if has_protocol_features {
+            let protocol_features = frontend
+                .get_protocol_features()
+                .map_err(|e| io::Error::new(ErrorKind::Other, e))?;
+
+            let mut our_protocol_features = VhostUserProtocolFeatures::empty();
+            if protocol_features.contains(VhostUserProtocolFeatures::CONFIG) {
+                our_protocol_features |= VhostUserProtocolFeatures::CONFIG;
+            }
+            if protocol_features.contains(VhostUserProtocolFeatures::MQ) {
+                our_protocol_features |= VhostUserProtocolFeatures::MQ;
+            }
+
+            frontend
+                .set_protocol_features(our_protocol_features)
+                .map_err(|e| io::Error::new(ErrorKind::Other, e))?;
+        }
+
+        // Determine actual queue count - may require protocol feature negotiation
+        let actual_num_queues = if num_queues == 0 {
+            if has_protocol_features {
+                let backend_queue_num = frontend
+                    .get_queue_num()
+                    .map_err(|e| io::Error::new(ErrorKind::Other, e))?;
+
+                debug!(
+                    "{}: backend reports {} queues available",
+                    device_name, backend_queue_num
+                );
+
+                backend_queue_num as usize
+            } else {
+                return Err(io::Error::new(
+                    ErrorKind::InvalidInput,
+                    "Backend doesn't support protocol features, must specify queue count",
+                ));
+            }
+        } else {
+            num_queues as usize
+        };
+
+        debug!(
+            "{}: using {} queues (requested: {}, sizes provided: {})",
+            device_name,
+            actual_num_queues,
+            num_queues,
+            queue_sizes.len()
+        );
+
+        let default_size = queue_sizes.last().copied().unwrap_or(256);
+        let queue_configs: Vec<_> = (0..actual_num_queues)
+            .map(|i| {
+                let size = queue_sizes.get(i).copied().unwrap_or(default_size);
+                QueueConfig::new(size)
+            })
+            .collect();
+
+        Ok(VhostUserDevice {
+            frontend: Arc::new(Mutex::new(frontend)),
+            device_type,
+            device_name,
+            queue_configs,
+            avail_features,
+            has_protocol_features,
+            acked_features: 0,
+            device_state: DeviceState::Inactive,
+            activate_evt: EventFd::new(EFD_NONBLOCK)?,
+            vring_call_event: None,
+        })
+    }
+
+    /// Activate the vhost-user device by setting up memory and vrings.
+    fn activate_vhost_user(
+        &mut self,
+        mem: &GuestMemoryMmap,
+        queues: &[DeviceQueue],
+    ) -> IoResult<()> {
+        let mut frontend = self.frontend.lock().unwrap();
+
+        debug!("{}: activating vhost-user device", self.device_name);
+
+        // Combine guest-acked features with backend-only features (QEMU approach)
+        let backend_feature_bits = if self.has_protocol_features {
+            self.acked_features | VHOST_USER_F_PROTOCOL_FEATURES
+        } else {
+            self.acked_features
+        };
+
+        frontend
+            .set_owner()
+            .map_err(|e| io::Error::new(ErrorKind::Other, e))?;
+
+        // Only share memory regions that have file backing (memfd)
+        let regions: Vec<VhostUserMemoryRegionInfo> = mem
+            .iter()
+            .filter_map(|region| {
+                if region.file_offset().is_some() {
+                    Some(VhostUserMemoryRegionInfo::from_guest_region(region))
+                } else {
+                    None
+                }
+            })
+            .collect::<Result<Vec<_>, _>>()
+            .map_err(|e| {
+                error!(
+                    "{}: failed to convert memory regions: {:?}",
+                    self.device_name, e
+                );
+                io::Error::new(ErrorKind::Other, e)
+            })?;
+
+        debug!(
+            "{}: sharing {} file-backed regions with backend",
+            self.device_name,
+            regions.len()
+        );
+
+        frontend.set_mem_table(&regions).map_err(|e| {
+            error!("{}: set_mem_table failed: {:?}", self.device_name, e);
+            io::Error::new(ErrorKind::Other, e)
+        })?;
+
+        // If protocol features not negotiated, this triggers automatic ring enabling
+        frontend
+            .set_features(backend_feature_bits)
+            .map_err(|e| io::Error::new(ErrorKind::Other, e))?;
+
+        let vring_call_event = EventFd::new(EFD_NONBLOCK)?;
+
+        for (queue_index, device_queue) in queues.iter().enumerate() {
+            let queue = &device_queue.queue;
+
+            frontend
+                .set_vring_num(queue_index, queue.actual_size())
+                .map_err(|e| io::Error::new(ErrorKind::Other, e))?;
+
+            // Set vring base
+            frontend
+                .set_vring_base(queue_index, 0)
+                .map_err(|e| io::Error::new(ErrorKind::Other, e))?;
+
+            // Vring addresses in queue are GPAs, but vhost-user protocol expects VMM VAs
+            let desc_table_gpa = queue.desc_table.0;
+            let avail_ring_gpa = queue.avail_ring.0;
+            let used_ring_gpa = queue.used_ring.0;
+
+            let desc_table_vmm =
+                mem.get_host_address(Address::new(desc_table_gpa))
+                    .map_err(|_| {
+                        io::Error::new(
+                            ErrorKind::InvalidInput,
+                            format!("GPA 0x{:x} not found in any memory region", desc_table_gpa),
+                        )
+                    })? as u64;
+            let avail_ring_vmm =
+                mem.get_host_address(Address::new(avail_ring_gpa))
+                    .map_err(|_| {
+                        io::Error::new(
+                            ErrorKind::InvalidInput,
+                            format!("GPA 0x{:x} not found in any memory region", avail_ring_gpa),
+                        )
+                    })? as u64;
+            let used_ring_vmm = mem
+                .get_host_address(Address::new(used_ring_gpa))
+                .map_err(|_| {
+                    io::Error::new(
+                        ErrorKind::InvalidInput,
+                        format!("GPA 0x{:x} not found in any memory region", used_ring_gpa),
+                    )
+                })? as u64;
+
+            let vring_config = VringConfigData {
+                flags: 0,
+                queue_max_size: queue.get_max_size(),
+                queue_size: queue.actual_size(),
+                desc_table_addr: desc_table_vmm,
+                used_ring_addr: used_ring_vmm,
+                avail_ring_addr: avail_ring_vmm,
+                log_addr: None,
+            };
+
+            frontend
+                .set_vring_addr(queue_index, &vring_config)
+                .map_err(|e| {
+                    error!("{}: set_vring_addr failed: {:?}", self.device_name, e);
+                    io::Error::new(ErrorKind::Other, e)
+                })?;
+
+            frontend
+                .set_vring_kick(queue_index, &device_queue.event)
+                .map_err(|e| {
+                    error!("{}: set_vring_kick failed: {:?}", self.device_name, e);
+                    io::Error::new(ErrorKind::Other, e)
+                })?;
+
+            frontend
+                .set_vring_call(queue_index, &vring_call_event)
+                .map_err(|e| {
+                    error!("{}: set_vring_call failed: {:?}", self.device_name, e);
+                    io::Error::new(ErrorKind::Other, e)
+                })?;
+
+            // Per QEMU vhost.c: when VHOST_USER_F_PROTOCOL_FEATURES is not negotiated,
+            // the rings start directly in the enabled state, and set_vring_enable will fail.
+            if self.has_protocol_features {
+                frontend
+                    .set_vring_enable(queue_index, true)
+                    .map_err(|e| io::Error::new(ErrorKind::Other, e))?;
+            } else {
+                debug!(
+                    "{}: vring {} already enabled (protocol features not negotiated)",
+                    self.device_name, queue_index
+                );
+            }
+        }
+
+        self.vring_call_event = Some(vring_call_event);
+
+        debug!(
+            "{}: vhost-user device activated successfully",
+            self.device_name
+        );
+
+        Ok(())
+    }
+}
+
+impl VirtioDevice for VhostUserDevice {
+    fn device_type(&self) -> u32 {
+        self.device_type
+    }
+
+    fn device_name(&self) -> &str {
+        &self.device_name
+    }
+
+    fn queue_config(&self) -> &[QueueConfig] {
+        &self.queue_configs
+    }
+
+    fn avail_features(&self) -> u64 {
+        self.avail_features
+    }
+
+    fn acked_features(&self) -> u64 {
+        self.acked_features
+    }
+
+    fn set_acked_features(&mut self, acked_features: u64) {
+        self.acked_features = acked_features;
+    }
+
+    fn read_config(&self, offset: u64, data: &mut [u8]) {
+        // For now, configuration space reads are not supported
+        // This can be extended using VHOST_USER_GET_CONFIG
+        debug!(
+            "{}: config read at offset {} (not yet implemented)",
+            self.device_name, offset
+        );
+        data.fill(0);
+    }
+
+    fn write_config(&mut self, offset: u64, _data: &[u8]) {
+        // For now, configuration space writes are not supported
+        // This can be extended using VHOST_USER_SET_CONFIG
+        debug!(
+            "{}: config write at offset {} (not yet implemented)",
+            self.device_name, offset
+        );
+    }
+
+    fn activate(
+        &mut self,
+        mem: GuestMemoryMmap,
+        interrupt: InterruptTransport,
+        queues: Vec<DeviceQueue>,
+    ) -> ActivateResult {
+        if let Err(e) = self.activate_vhost_user(&mem, &queues) {
+            error!(
+                "{}: failed to activate vhost-user device: {}",
+                self.device_name, e
+            );
+            return Err(ActivateError::BadActivate);
+        }
+
+        self.device_state = DeviceState::Activated(mem, interrupt);
+
+        if let Err(e) = self.activate_evt.write(1) {
+            error!(
+                "{}: failed to write activate event: {}",
+                self.device_name, e
+            );
+            return Err(ActivateError::BadActivate);
+        }
+
+        Ok(())
+    }
+
+    fn is_activated(&self) -> bool {
+        matches!(self.device_state, DeviceState::Activated(_, _))
+    }
+
+    fn reset(&mut self) -> bool {
+        debug!("{}: resetting vhost-user device", self.device_name);
+
+        // Disable all vrings
+        if let Ok(mut frontend) = self.frontend.lock() {
+            for queue_index in 0..self.queue_configs.len() {
+                if let Err(e) = frontend.set_vring_enable(queue_index, false) {
+                    debug!(
+                        "{}: failed to disable vring {} during reset: {}",
+                        self.device_name, queue_index, e
+                    );
+                }
+            }
+        }
+
+        self.vring_call_event = None;
+        self.device_state = DeviceState::Inactive;
+        true
+    }
+}
+
+impl VhostUserDevice {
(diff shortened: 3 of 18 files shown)


src/devices/src/virtio/vhost_user/device.rs
```diff
@@ -0,0 +1,444 @@
...
+        }
+
+        debug!(
+            "{}: vhost-user device activated successfully",
+            self.device_name
+        );
+
+        Ok(())
+    }
+}
```
mtjhrc: Having to have monitoring threads to forward the events seems unfortunate , maybe we could avoid it  with better API somewhere else (I don't know), I guess currently the problem is that `interrupt.signal_used_queue` does more than than just signal the `eventfd` (otherwise we could just pass along the `eventfd` to the frontend).

Anyways I think a simple improvement can be made to only have 1 monitoring thread:
- you only create 1 `vring_call_event` for all queues and pass that fd for all the queues to the `frontend`
- in the monitoring thread you do the same - just wait for the eventfd and  then `signal_used_queue()`.

dorindabassey: Right! The monitoring thread is the bridge between the vhost backend and the MMIO interrupt mechanism, considering the fact that we can't pass the interrupt eventfd directly to the vhost frontend. Agreed, we can use a single shared `vring_call_event` for all queues instead of one per queue, reducing it to just 1 monitoring thread. I think that's the best we can do with MMIO.

src/devices/src/virtio/vhost_user/rng.rs
```diff
@@ -95,3 +95,22 @@ impl VirtioDevice for VhostUserRng {
...
+mod tests {
+    use super::{VhostUserRng, NUM_QUEUES, QUEUE_SIZES, RNG_DEV_ID, VIRTIO_ID_RNG};
+
+    #[test]
+    fn test_rng_constants() {
+        assert_eq!(RNG_DEV_ID, "vhost_user_rng");
+        assert_eq!(NUM_QUEUES, 1);
+        assert_eq!(QUEUE_SIZES[0], 256);
+        assert_eq!(VIRTIO_ID_RNG, 4);
+    }
```
mtjhrc: This test doesn't seem helpful, it only asserts the values of the constants. I would just remove it.

src/devices/src/virtio/vhost_user/device.rs
```diff
@@ -0,0 +1,532 @@
...
+
+    #[test]
+    fn test_protocol_feature_flags() {
+        let mut features = VhostUserProtocolFeatures::empty();
+        features |= VhostUserProtocolFeatures::CONFIG;
+        features |= VhostUserProtocolFeatures::MQ;
+
+        assert!(features.contains(VhostUserProtocolFeatures::CONFIG));
+        assert!(features.contains(VhostUserProtocolFeatures::MQ));
+    }
```
mtjhrc: This only asserts on functionality of the `bitflags` crate, it doesn't assert on anything to do with the vhost user functionality, I would just remove the test.

src/devices/src/virtio/vhost_user/device.rs
```diff
@@ -0,0 +1,532 @@
...
+        let queue_sizes = vec![64, 128, 256, 512];
+        let queues: Vec<VirtQueue> = queue_sizes
+            .iter()
+            .map(|&size| VirtQueue::new(size))
+            .collect();
+
+        assert_eq!(queues.len(), 4);
+        assert_eq!(queues[0].get_max_size(), 64);
+        assert_eq!(queues[3].get_max_size(), 512);
+    }
```
mtjhrc: This only asserts on functionality of the vector from stdlib (which we surely don't need to test) and the `get_max_size()` getter, for which an unit test should go to the virtio queue module.
This doesn't assert on anything to do with the vhost user functionality, I would just remove this test.

src/devices/src/virtio/vhost_user/device.rs
```diff
@@ -0,0 +1,532 @@
...
+    fn test_feature_separation() {
+        let backend_features = 0x1234_5678 | VHOST_USER_F_PROTOCOL_FEATURES;
+        let virtio_features = backend_features & !VHOST_USER_F_PROTOCOL_FEATURES;
+        let backend_only = backend_features & VHOST_USER_F_PROTOCOL_FEATURES;
+
+        assert_eq!(virtio_features, 0x1234_5678);
+        assert_eq!(backend_only, VHOST_USER_F_PROTOCOL_FEATURES);
+        assert_eq!(virtio_features & (1 << 30), 0);
+        assert_eq!(backend_only & (1 << 30), 1 << 30);
+    }
```
mtjhrc: Again, this doesn't seem to test anything relevant to the vhost-user device.

src/devices/src/virtio/vhost_user/device.rs
```diff
@@ -0,0 +1,532 @@
...
+        assert!(test_gpa_to_vmm_va(&mem, 0xfff).is_ok());
+        assert!(test_gpa_to_vmm_va(&mem, 0x1000).is_err());
+        assert!(test_gpa_to_vmm_va(&mem, 0x10000).is_ok());
+        assert!(test_gpa_to_vmm_va(&mem, 0x11fff).is_ok());
+        assert!(test_gpa_to_vmm_va(&mem, 0x12000).is_err());
+
+        // Test gaps and invalid addresses
+        assert!(test_gpa_to_vmm_va(&mem, 0x5000).is_err());
+        assert!(test_gpa_to_vmm_va(&mem, 0x200000).is_err());
+    }
```
mtjhrc: This seems like it was meant to test the `gpa_to_vmm_va` method (which would be good!), but this actually tests the function `test_gpa_to_vmm_va`, which is defined only in the tests.

The `VhostUserDevice::gpa_to_vmm_va` takes in `&self`, but it really doesn't need to, so you can actually just remove that parameter and test it easily here.

src/devices/src/virtio/vhost_user/device.rs
```diff
@@ -0,0 +1,532 @@
...
+        // Without protocol features
+        let backend_no_proto: u64 = 0x1234_5678;
+        let bits = guest_acked | (backend_no_proto & VHOST_USER_F_PROTOCOL_FEATURES);
+        assert_eq!(bits, guest_acked);
+
+        // With protocol features
+        let backend_with_proto = 0x1234_5678 | VHOST_USER_F_PROTOCOL_FEATURES;
+        let bits = guest_acked | (backend_with_proto & VHOST_USER_F_PROTOCOL_FEATURES);
+        assert_eq!(bits, guest_acked | VHOST_USER_F_PROTOCOL_FEATURES);
+    }
```
mtjhrc: This isn't clear to me what this tests (doesn't seem like it does anything besides bit-wise math?). If there is some specific relationship between constants that needs to hold, than that could be nice to assert (but maybe add a comment, because I am confused), otherwise I'd just remove this test too.

mtjhrc (changes requested):
I had a look at the tests, and honestly I don't see much value in them, we don't have any arbitrary coverage percent metric for merging stuff, so I would just remove most of them.

The rest of the code looks good, but I've only had a quick look so far, when this was still a draft, I'll try running this and have another look later. 

Having vhost-user support in libkrun seems pretty cool, thanks!

src/devices/src/virtio/vhost_user/device.rs
dorindabassey: yeah, I can remove them, it's just that the actual feature separation happens in `VhostUserDevice::new()`, but we can't test that without a real backend.

src/devices/src/virtio/vhost_user/device.rs
dorindabassey: Right! I'll remove these test. But now looking at what's left now, I'll just remove all the tests. We can test the implementation with real backends like vhost-device-rng.

src/devices/src/virtio/vhost_user/device.rs
mtjhrc: @dorindabassey  This test was actually pretty good, if you do what I mentioned in comment, you can avoid instantiating the device, and can add the this test back!

src/devices/src/virtio/vhost_user/device.rs
mtjhrc: > it's just that the actual feature separation happens in VhostUserDevice::new(),

Yeah, that is kind of the general problem with unit testing the devices, it's quite hard to test the actual implementation :pensive: 

I think it's definitely achievable to write a true unit test for the real `VhostUserDevice` impl, it's just annoying/hard to do. To do it, I'd say you need to actually provide a mock `frontend` to the device, which e.g. returns hardcoded features and asserts on that in the test. But for that you need some test utils infrastructure to be able to create a mock of something like that. For truly a unit test you could intercept it at a trait level - I think the `mockall` crate works for this nicely. Easier thing to do, would probably be to create a really simple mock vhost-user-device using the actuall rust-vmm crates and test against that. 

Not saying you need to something like that as part of this PR, I'm just saying how I think It needs to be approached if we want unit test.

dorindabassey: Yeah, I agree that mocking the frontend would be the right approach for a proper unit test. For now, I just removed all the tests since they weren't actually testing the vhost-user functionality. If we want to add proper unit testing infrastructure in the future like the mocked frontend, etc, I agree that could be a separate effort/PR that benefits all vhost-user device types.

src/devices/src/virtio/vhost_user/device.rs
dorindabassey: I'm personally more of an "all or nothing" person when it comes to test coverage. Having just this one small test would give us maybe 5-10% coverage of the vhost-user implementation, while the core functionality (feature negotiation, vring setup, memory region sharing, interrupt handling) is untested. I'd prefer to bring it back in a more comprehensive test suite like we talked about the mock frontend. But for now maybe it's better to ship it without the test, and validate it through actual usage.

tylerfanelli:
Not really a review, just a comment:

@dorindabassey This is some great work and a huge addition to the project.

@slp passt uses vhost-user to provide near-native network performance in user mode for QEMU. Perhaps we should consider doing the same here, especially as the main network driver for the v2 API.

sbrivio-rh:
> @slp passt uses vhost-user to provide near-native network performance in user mode for QEMU. Perhaps we should consider doing the same here, especially as the main network driver for the v2 API.

Right, that would be great to have! It's mostly about throughput and latency, but also, perhaps, one day, to implement live migration of TCP connections, as it's only available via vhost-user interface.

I'm not sure how much effort networking support would take on top of this pull request, and whether it's beyond its scope or not, but let me share a couple of pointers just in case. The part of API that's implemented by passt is this:

https://passt.top/passt/tree/vhost_user.c?id=af7b81b5408da8c56bb22dd11679f2b4024a45c8#n1128

And there's a bit of documentation about how it can be used with QEMU here:

https://www.qemu.org/docs/master/system/devices/net.html#using-passt-as-the-user-mode-network-stack

Let me know if you have any question! I'll also tag @vivier, the author of the vhost-user implementation in passt, for good measure.

tylerfanelli:
> I'm not sure how much effort networking support would take on top of this pull request, and whether it's beyond its scope or not

It's fine for this PR to stand as-is. This can be used as a base for another series adding the network implementation.

include/libkrun.h
```diff
@@ -717,6 +717,28 @@ int krun_add_input_device_fd(uint32_t ctx_id, int input_fd);
...
+ *  "ctx_id"      - the configuration context ID.
+ *  "socket_path" - path to the vhost-user Unix domain socket (e.g., "/tmp/vhost-rng.sock").
+ *
+ * Returns:
+ *  Zero on success or a negative error number on failure.
+ *  -EINVAL  - Invalid socket_path
+ *  -ENOENT  - Context doesn't exist
+ *  -ENOTSUP - vhost-user support not compiled in
+ */
+int32_t krun_set_vhost_user_rng(uint32_t ctx_id, const char *socket_path);
```
tylerfanelli: With possible extensions in mind, is there any way this can be generalized to something like:

```
int32_t krun_add_vhost_user_device(uint32_t ctx_id, ...);
```

tylerfanelli (changes requested):

include/libkrun.h
mtjhrc: Hmm, I think this is a good point actually. For things like virtio-rng, and the mentioned virtio-net (passt) I think this should be rather simple - basically just configuring the number and sizes of queues and the device ID actually.

(some things would be more tricky to generalize, I can think of virtio-gpu, any device that requires shared memory regions ( that is being worked on: https://github.com/rust-vmm/vhost/pull/339 ).

mtjhrc:
> passt uses vhost-user to provide near-native network performance in user mode for QEMU. Perhaps we should consider doing the same here, especially as the main network driver for the v2 API.

> I'm not sure how much effort networking support would take on top of this pull request, and whether it's beyond its scope or not

Adding support for virtio-net over vhost doesn't require any changes to any of our networking code. Our `virtio-net` is not attached by default.

All the user would have to do to enable it is:
1) Disable the implicit vsock device to get rid of TSI (`krun_disable_implicit_vsock`). Optionally add a vsock device back (`krun_add_vsock`) to use vsock for other purpose
2) Use something like `krun_add_vhost_user_device` to attach the net device

include/libkrun.h
tylerfanelli: Perhaps we can define a `struct vhost_user_params` that is essentially a generic pointer to some data based on the device ID? This can provide the API extensibility needed for eventually adding more complicated devices.

dorindabassey: I'm currently refactoring it to do something like this:
```
int32_t krun_add_vhost_user_device(uint32_t ctx_id,
                                   krun_vhost_user_device_type device_type,
                                   const char *socket_path,
                                   uint16_t num_queues,
                                   uint16_t queue_size);
```
For RNG specifically, you'd call it like: `krun_add_vhost_user_device(ctx, KRUN_VHOST_USER_DEVICE_RNG, "/path/to/rng.sock", 0, 0);`
The 0s would use defaults (1 queue, 256 size for RNG).                                                 
Other devices like sound, net, can, etc. could be added later with device-specific APIs as needed, for more complex device like gpu we could extend this with a `void *device_config` parameter for device-specific config.

mtjhrc: > I'm currently refactoring it to do something like this:
> 
> ```
> int32_t krun_add_vhost_user_device(uint32_t ctx_id,
>                                    krun_vhost_user_device_type device_type,
>                                    const char *socket_path,
>                                    uint16_t num_queues,
>                                    uint16_t queue_size);
> ```
> 

I think it would be nice to support configuring different queue size for each queue. So just have `uint16_t *queue_sizes` I guess. 

Another thing I would consider adding is a `const char *name` that would correspond to the `VirtioDevice::device_name` - seems like it would be helpful for debugging (we currently just use it for logging).

dorindabassey: yeah make sense, I'll add those Thanks!

tylerfanelli: I'd probably remove the default (1 queue, 256 size) values and just make it required for the user to configure the virtqueues explicitly.

```
krun_add_vhost_user_device(ctx, KRUN_VHOST_USER_DEVICE_RNG, "/path/to/rng.sock", 1, 256);
```

If they submit something like:
```
krun_add_vhost_user_device(ctx, KRUN_VHOST_USER_DEVICE_RNG, "/path/to/rng.sock", 0, 0);
```
It should fail with something like `Invalid virtqueue configuration` or something of the sort.

dorindabassey: > I think it would be nice to support configuring different queue size for each queue. So just have uint16_t *queue_sizes I guess.

I've implemented the `queue_sizes` array change as suggested, so each queue can have a different size.

> Another thing I would consider adding is a const char *name that would correspond to the VirtioDevice::device_name - seems like it would be helpful for debugging (we currently just use it for logging).

Done! Added the `name` parameter (NULL means auto-generate from device_type)

> I'd probably remove the default (1 queue, 256 size) values and just make it required for the user to configure the virtqueues explicitly.

I'd rather keep the defaults (0, NULL = use device defaults) reasons being
1. Most users won't know what queue configuration to use. For simple devices like `rng`, forcing them to specify `krun_add_vhost_user_device(ctx, KRUN_VHOST_USER_DEVICE_RNG, socket, 1, sizes)` means they need to know RNG uses 1 queue, and that 256 is a reasonable queue size, and also have to write boilerplate for every device.
2. When we add more device types e.g (sound, CAN), each will have different default queue configs, adding these in documentation users would need to look up and specify the correct values for each device type in their code.

The current API supports both use cases - simple with defaults, and advanced with full control. @tylerfanelli WDYT?

mtjhrc: > I've implemented the queue_sizes array change as suggested, so each queue can have a different size.

:+1: 

> Done! Added the name parameter (NULL means auto-generate from device_type)

:+1: 

> Most users won't know what queue configuration to use. For simple devices like rng, forcing them to specify krun_add_vhost_user_device(ctx, KRUN_VHOST_USER_DEVICE_RNG, socket, 1, sizes) means they need to know RNG uses 1 queue, and that 256 is a reasonable queue size, and also have to write boilerplate for every device.

I agree that having user know that is a bit annoying (and e.g. QEMU doesn't force the user to know that). However I quite dislike using `0` for the number of queues and NULL for sizes. Instead I would much prefer to `#define` the suggested defaults in the header. 

This way its quite easy for the user  to see what the default value is without studying the docs or even worse having to look it up in libkrun source code. I think this is a reasonable middle ground. 

```
#define KRUN_DEVICE_RNG_NUM_QUEUES 1
#define KRUN_DEVICE_RNG_DEFAULT_QUEUE_SIZES (uint16_t[]){256}


krun_add_vhost_user_device(ctx,
  KRUN_VHOST_USER_DEVICE_RNG, 
  "vhost-user-rng", 
  socket, 
  KRUN_DEVICE_RNG_NUM_QUEUES, 
  KRUN_DEVICE_RNG_DEFAULT_QUEUE_SIZES
);
```

src/devices/src/virtio/vhost_user/rng.rs
mtjhrc, line 1: Since we switched to a generic API for attaching vhost-user devices, at this point I think we I would remove the specific `VhostUserRng` altogether. Instead just instantiate the `VhostUserDevice` with the arguments obtained from user. 

While it's possible we might eventually need some device specific front-ends, for devices that don't require anything special this just introduces unnecessary boilerplate.

include/libkrun.h
```diff
@@ -717,6 +717,60 @@ int krun_add_input_device_fd(uint32_t ctx_id, int input_fd);
...
 
+/**
+ * Vhost-user device types.
+ * These correspond to virtio device type IDs for devices.
+ */
+typedef enum {
+    KRUN_VHOST_USER_DEVICE_RNG = 4,
+    KRUN_VHOST_USER_DEVICE_SND = 25,
+    KRUN_VHOST_USER_DEVICE_CAN = 36,
+} krun_vhost_user_device_type;
```
mtjhrc: Up to this point we have avoided using C enums in the API. 
Technically they don't have a very stable ABI, the size of the backing integer is up to the compiler and compiler options - this could be anything from uint8_t to int64_t ( though unlikely, and typically its just a 32-bit "int") .

I would suggest to stay on the safe side and just `#define` the values

Btw.:
C++, C23 and Rust allow you to specify the backing integer type for enums, but that's not an option for us

src/vmm/src/builder.rs
```diff
@@ -2246,6 +2409,62 @@ fn attach_rng_device(
...
+                device_manager::mmio::Error::VhostUserDevice(std::io::Error::new(
+                    std::io::ErrorKind::InvalidInput,
+                    format!(
+                        "Unknown vhost-user device type {}",
+                        device_config.device_type
+                    ),
+                )),
+            ));
+        }
+    }
```
mtjhrc: I would say we shouldn't care what the  `VIRTIO_ID_*` is at all except for generating the string name for the device, and we should just pass along whatever the user gives us to a generic vhost-user device wrapper.
If we don't know what the device ID is for and user didn't provide a name, I would just generate a name for the device - something like "vhost-user-uknown-40"  (where 40 is the user provided id). 

Also, I think we should just allow unknown/potentially invalid device IDs to be passed - the user might want to run a custom kernel in the guest which understands their custom device and I don't see a good reason to prevent them.

src/devices/src/virtio/vhost_user/device.rs
```diff
@@ -0,0 +1,426 @@
...
+            // We want CONFIG and MQ support at minimum
+            let mut our_protocol_features = VhostUserProtocolFeatures::empty();
+
+            if protocol_features.contains(VhostUserProtocolFeatures::CONFIG) {
+                our_protocol_features |= VhostUserProtocolFeatures::CONFIG;
+            }
+
+            if protocol_features.contains(VhostUserProtocolFeatures::MQ) {
+                our_protocol_features |= VhostUserProtocolFeatures::MQ;
+            }
```
mtjhrc: I looked up the spec and I feel like this is not fully correct. Our frontend advertises that it supports  `VHOST_USER_PROTOCOL_F_MQ`, that is fine. But that feature means the frontend should ask the backend for the number of queues (`VhostUserFrontend::get_queue_num`), problem is that the backend might not support this feature so we can't rely on this.

This opens up some problems: 
1) if the user provided `0` for the number of queues then that's easy I think -  we should create however many queues the device requests (if the device doesn't support this feature than this should be an error as we can't know...)
2) if the user said the device should have e.g. 2 queues and the device says there should be e.g. 3, then I guess this should be an error? Maybe we could just create `3` queues and `log::warn!`, but that's a bit sketchy
3)  I'm not sure what should be the sizes of the created queues. This doesn't neatly fit into the user providing an array. Does anyone have some suggestion how to tackle that?

include/libkrun.h
tylerfanelli: +1 to @mtjhrc's suggestion. With that, the user still has to explicitly configure the number of queues and their size, but they are still "known" default values that the user doesn't need prior knowledge of. I agree that is a good middle ground.

src/devices/src/virtio/vhost_user/device.rs
dorindabassey: Regarding this, The vhost-device-rng backend actually advertises `VhostUserProtocolFeatures::MQ` and implements num_queues() (returning 1 for RNG). So we should actually call `get_queue_num()` after negotiating MQ support. I see 3 ways we can handle the queue count issue:
1. If the user provides 0 queues + Backend supports MQ: Call get_queue_num() and use backend's value
2. If the user provides 0 queues + Backend doesn't support MQ: Error - we have no way to know the queue count
3. If the user provides N queues: Use N directly (assumes the user knows their backend's configuration)

For queue sizes, if the user provides fewer sizes than queues needed, we repeat the last provided size or default to 256 for all queues if none is provided. JFYI I'm currently making these changes.

include/libkrun.h
dorindabassey: make sense, I settled for giving users two options (using defines and auto-detecting from the backend): https://github.com/containers/libkrun/pull/527/changes#diff-2d2ad6720e07dcce279bd688772adf5b96526a8679a9fbca186557239b1d49f9R757-R764 
The auto-detect pattern (0/NULL) is useful when the backend implements the MQ protocol feature and can report its own queue count via `get_queue_num()`. For example, vhost-device-rng advertises MQ support, so passing 0 will query the backend and use whatever it reports

src/libkrun/src/lib.rs
```diff
@@ -1773,6 +1773,82 @@ pub unsafe extern "C" fn krun_set_snd_device(ctx_id: u32, enable: bool) -> i32 {
...
+            Ok(s) if !s.is_empty() => Some(s.to_string()),
+            _ => None,
+        }
+    };
+
+    let queue_sizes_vec = if queue_sizes.is_null() || num_queues == 0 {
+        Vec::new()
+    } else {
+        std::slice::from_raw_parts(queue_sizes, num_queues as usize).to_vec()
+    };
```
mtjhrc: This is bit of an oversight. If the user provides `0`, they want to have the number of queues autodetected (as per the documentation), but they also want to provide e.g. 2 specific queue sizes - this would construct a vector of size `0` and ignore the user requested sizes.

I see 3 options to deal with this: 
- introduce a sentinel value (`0`), make the user terminate the queue sizes array
- introduce another parameter that is the size of the queues array, but not necessarily the number of queues (I don't really prefer this - seems confusing)
- introduce another `flags` parameter, currently we would have only 1 flag - the  `MQ` feature (in that case the `num_queues` is optional and only for the queue_sizes array)

include/libkrun.h
```diff
@@ -717,6 +717,69 @@ int krun_add_input_device_fd(uint32_t ctx_id, int input_fd);
  */
 int32_t krun_set_snd_device(uint32_t ctx_id, bool enable);
 
+/**
+ * Vhost-user device types.
+ * These correspond to virtio device type IDs for devices.
+ */
+#define KRUN_VHOST_USER_DEVICE_RNG 4
+#define KRUN_VHOST_USER_DEVICE_SND 25
+#define KRUN_VHOST_USER_DEVICE_CAN 36
```
mtjhrc: Perhaps make it more clear in the name that the ID is specified by virtio spec not vhost spec?

```suggestion
#define KRUN_VIRTIO_DEVICE_ID_RNG 4
#define KRUN_VIRTIO_DEVICE_ID_SND 25
#define KRUN_VIRTIO_DEVICE_ID_CAN 36
```

Or even just:
```suggestion
#define KRUN_VIRTIO_DEVICE_RNG 4
#define KRUN_VIRTIO_DEVICE_SND 25
#define KRUN_VIRTIO_DEVICE_CAN 36
```

src/devices/src/virtio/vhost_user/device.rs
```diff
@@ -0,0 +1,475 @@
...
+        // For now, configuration space reads are not supported
+        // This can be extended using VHOST_USER_GET_CONFIG
+        debug!(
+            "{}: config read at offset {} (not yet implemented)",
+            self.device_name, offset
+        );
+        // Fill with zeros
+        for byte in data.iter_mut() {
+            *byte = 0;
+        }
```
mtjhrc:
```suggestion
        data.fill(0);
```

src/devices/src/virtio/vhost_user/device.rs
```diff
@@ -0,0 +1,475 @@
...
+
+        // Combine guest-acked features with backend-only features (QEMU approach)
+        let backend_feature_bits = self.acked_features | self.backend_features;
+
+        frontend
+            .set_features(backend_feature_bits)
+            .map_err(|e| io::Error::new(ErrorKind::Other, e))?;
+
+        // If protocol features are supported and enabled, negotiate them now
+        if self.backend_features & (1 << 30) != 0 {
```
mtjhrc: Both functions `new` and `activate` negotiate the protocol features. I think you can negotiate the protocol features only in `new`. You can also ask the backend for the virtio (the non-protocol) features in `new`, but you should confirm/set them in `activate`.

src/devices/src/virtio/vhost_user/device.rs
```diff
@@ -0,0 +1,475 @@
...
+        let mut frontend = self.frontend.lock().unwrap();
+
+        debug!("{}: activating vhost-user device", self.device_name);
+
+        // Combine guest-acked features with backend-only features (QEMU approach)
+        let backend_feature_bits = self.acked_features | self.backend_features;
+
+        frontend
+            .set_features(backend_feature_bits)
+            .map_err(|e| io::Error::new(ErrorKind::Other, e))?;
```
mtjhrc: This seems incorrect according to the spec - if the backend doesn't support `VHOST_USER_F_PROTOCOL_FEATURES` then this will implicitly enable all vrings (but they are not configured yet).  

Acording to the spec:
> In addition, upon receiving a VHOST_USER_SET_FEATURES message from the front-end without VHOST_USER_F_PROTOCOL_FEATURES set, the back-end must enable all rings immediately.

I guess the correct order is to call `set_features` after configuring the vrings (but before actually enabling the vrings), but I am not 100% sure.

mtjhrc (changes requested):

src/vmm/src/builder.rs
```diff
@@ -972,7 +976,28 @@ pub fn build_microvm(
...
+                    attach_vhost_user_device(&mut vmm, intc.clone(), device_config)?;
+                }
+            } else {
+                attach_rng_device(&mut vmm, event_manager, intc.clone())?;
+            }
+        }
+
+        #[cfg(not(feature = "vhost-user"))]
+        attach_rng_device(&mut vmm, event_manager, intc.clone())?;
+    }
```
mtjhrc, line 1012: This still assumes the user would only attach a virtio-rng over vhost, nothing else (even the public API function allows more device types). The logic is weird here - if you have a vhost-user-rng (all other vhost-user devices get attached too, if you don't have a vhost-user-rng none of the vhost-user devices get attached), yet the krun_add_vhost_user_device explicitly allows more device types (and it should IMO)

Maybe consider adding a krun_disable_implicit_rng instead - mirroring the existing krun_disable_implicit_console/krun_disable_implicit_vsock.

src/libkrun/src/lib.rs
dorindabassey: maybeI'll go for option1, it solves the problem without breaking the API

src/devices/src/virtio/vhost_user/device.rs
dorindabassey: right! fixed it, now it follows the order `SET_OWNER`, `SET_MEM_TABLE`,  `SET_FEATURES`, configure vrings, `SET_VRING_ENABLE` (if protocol features negotiated)

src/vmm/src/builder.rs
```diff
@@ -976,7 +976,31 @@ pub fn build_microvm(
...
+            for device_config in &vm_resources.vhost_user_devices {
+                attach_vhost_user_device(&mut vmm, intc.clone(), device_config)?;
+            }
+
+            let has_vhost_user_rng = vm_resources
+                .vhost_user_devices
+                .iter()
+                .any(|dev| dev.device_type == VIRTIO_ID_RNG);
+
+            if !vm_resources.disable_implicit_rng && !has_vhost_user_rng {
```
mtjhrc: The idea I mentioned with `krun_disable_implicit_rng` was that it would replace this automatic disablement of the internal device. I'm not saying that the automatically disabling of the built-in device is necessarily bad though - in fact it will often be more user friendly.

However, having the user explicitly call  `krun_disable_implicit_rng` has one big advantage though, that this will definitely be consistent with future devices - imagine in the future we could have vhost alternatives for basically every device, but adding more auto detection rules here will not be backwards compatible, potentially causing great confusion with regards to what is in fact enabled.

To be clear, I am not saying this has to necessarily be changed, It's just an observation.

dorindabassey: Yeah, good point about future compatibility. For now I think the auto-detection provides a better user experience, but I'll keep this in mind if we add vhost-user support for other device types in the future.

src/libkrun/src/lib.rs
```diff
@@ -1776,6 +1776,101 @@ pub unsafe extern "C" fn krun_set_snd_device(ctx_id: u32, enable: bool) -> i32 {
...
+    use vmm::resources::VhostUserDeviceConfig;
+
+    const VIRTIO_ID_RNG: u32 = 4;
+    const VIRTIO_ID_SND: u32 = 25;
+    const VIRTIO_ID_CAN: u32 = 36;
+
+    if device_type != VIRTIO_ID_RNG && device_type != VIRTIO_ID_SND && device_type != VIRTIO_ID_CAN
+    {
+        return -libc::EINVAL;
+    }
```
mtjhrc: I think we should remove this check. 

Like I mentioned in other comment somewhere, I don't think there is a reason for limiting the device types user can attach. In fact the implementation is already generic enough for many device types and even future devices that don't exist in the spec yet. It is the job of the guest to understand the virtio device(s), if the guest OS doesn't understand the device type it can just ignore the device or crash if there is some mismatch, from the libkrun side we don't really care.

src/devices/src/virtio/vhost_user/device.rs
```diff
@@ -0,0 +1,450 @@
...
+
+        // Separate backend-only features from virtio features
+        let backend_features = avail_features & VHOST_USER_F_PROTOCOL_FEATURES;
+        let our_avail_features = avail_features & !VHOST_USER_F_PROTOCOL_FEATURES;
+
+        // Determine actual queue count - may require protocol feature negotiation
+        if backend_features & VHOST_USER_F_PROTOCOL_FEATURES != 0 {
+            frontend
+                .set_features(backend_features)
+                .map_err(|e| io::Error::new(ErrorKind::Other, e))?;
```
mtjhrc: Please remove this call here, this directly breaks passt and in general is incorrect.

Configuring the features here is redundant since the features are also configured in `activate()`. Apart from being redundant the  value of `backend_features` is also incorrect here (if you trace the logic `backend_features == VHOST_USER_F_PROTOCOL_FEATURES` at this point - the virtio features are stripped).

src/devices/src/virtio/vhost_user/device.rs
```diff
@@ -0,0 +1,450 @@
...
+        // that enables vhost-user protocol extensions. It's not a virtio feature,
+        // so we don't expose it to the guest, but we always use it with the backend.
+        const VHOST_USER_F_PROTOCOL_FEATURES: u64 = 1 << 30;
+
+        // Separate backend-only features from virtio features
+        let backend_features = avail_features & VHOST_USER_F_PROTOCOL_FEATURES;
+        let our_avail_features = avail_features & !VHOST_USER_F_PROTOCOL_FEATURES;
+
+        // Determine actual queue count - may require protocol feature negotiation
+        if backend_features & VHOST_USER_F_PROTOCOL_FEATURES != 0 {
```
mtjhrc: `VHOST_USER_F_PROTOCOL_FEATURES` is not a mask, but a single bit. To illustrate you are actually doing:
 ```Rust
if avail_features & VHOST_USER_F_PROTOCOL_FEATURES & VHOST_USER_F_PROTOCOL_FEATURES != 0) {
   ...
```


 
Personally I would just alias the variable or make the `avail_features` mutable and do something like this (but feel free to write it a bit differently, just remove the confusing and redundant bit operation):
```suggestion
        // Strip the vhost specific bit to leave only standard virtio features
        let has_backend_features = avail_features & VHOST_USER_F_PROTOCOL_FEATURES != 0;
        let avail_features = avail_features & !VHOST_USER_F_PROTOCOL_FEATURES;

        if has_backend_features != {
```
Then also change the field name of the struct into `has_backend_features: bool` (you are only ever storing the single bit there anyway).

src/devices/src/virtio/vhost_user/device.rs
```diff
@@ -0,0 +1,450 @@
...
+        // VHOST_USER_F_PROTOCOL_FEATURES (bit 30) is a backend-only feature
+        // that enables vhost-user protocol extensions. It's not a virtio feature,
+        // so we don't expose it to the guest, but we always use it with the backend.
+        const VHOST_USER_F_PROTOCOL_FEATURES: u64 = 1 << 30;
+
+        // Separate backend-only features from virtio features
+        let backend_features = avail_features & VHOST_USER_F_PROTOCOL_FEATURES;
+        let our_avail_features = avail_features & !VHOST_USER_F_PROTOCOL_FEATURES;
+
+        // Determine actual queue count - may require protocol feature negotiation
```
mtjhrc: This comment seems misplaced, maybe it was meant to be above line 140:
``let actual_num_queues = if num_queues == 0 {`

src/devices/src/virtio/vhost_user/device.rs
```diff
@@ -0,0 +1,450 @@
...
+        // If protocol features not negotiated, this triggers automatic ring enabling
+        frontend
+            .set_features(backend_feature_bits)
+            .map_err(|e| io::Error::new(ErrorKind::Other, e))?;
+
+        // Create single vring call event file descriptor (backend->guest interrupt)
+        // NOTE: Do NOT use EFD_NONBLOCK here - the monitoring thread needs to block
+        let vring_call_event = EventFd::new(0)?; // Blocking eventfd
+
+        let has_protocol_features = backend_feature_bits & (1 << 30) != 0;
```
mtjhrc: the magic hardcoded `1 << 30` constant here is in fact `VHOST_USER_F_PROTOCOL_FEATURES`.

src/devices/src/virtio/vhost_user/device.rs
mtjhrc: I meant: 
SET_OWNER, SET_MEM_TABLE, configure vrings, SET_FEATURES, SET_VRING_ENABLE (if protocol features negotiated)

Not sure if this a a problem in practice though, If we want to follow what the spec, it kind of implies this might be the correct order if  protocol features are NOT negotiated. (If protocol features are negotiated this doesn't matter and either order should be correct). But maybe I am wrong, the spec is not very precise.

src/devices/src/virtio/vhost_user/device.rs
```diff
@@ -0,0 +1,450 @@
...
+        device_type: u32,
+        device_name: String,
+        num_queues: u16,
+        queue_sizes: &[u16],
+    ) -> IoResult<Self> {
+        debug!("Connecting to vhost-user backend at {}", socket_path);
+
+        // Connect to the vhost-user backend
+        let stream = UnixStream::connect(socket_path)?;
+        let mut frontend = Frontend::from_stream(stream, 1);
```
mtjhrc: You have hardcoded only 1 queue here. If you attempt to use attempt to configure a device with more than 1 queue (and without MQ feature!), the configuration methods would fail.

You can simply do this:
```suggestion 
        // NOTE: `num_queues` could be 0 here, but this is actually fine 
        // because if `VhostUserProtocolFeatures::MQ` is supported the negotioted
        // value will be used automatically by Frontend
        let mut frontend = Frontend::from_stream(stream, num_queues as u64);
```

mtjhrc (changes requested):
I left some further review comments. 

After some simple fixes, this actually works great for for virtio-net over vhost using passt.

Here are my the necessary minimum changes to fix this for passt + example:
https://github.com/dorindabassey/libkrun/compare/vhost-user-support...mtjhrc:libkrun:vhost-user-net

The fixes are included in the changes requested to this PR. I'll create a separate PR for the example (and maybe an integration test too).

sbrivio-rh:
> After some simple fixes, this actually works great for for virtio-net over vhost using passt.

Crazy stuff, thanks to both of you! @vivier have you seen this? We should iperf3 it soon.

I'll try to find some time to play with this within a couple of weeks / months (surely too late for any concrete review / practical support here though).

src/devices/src/virtio/vhost_user/device.rs
mtjhrc: I meant:
SET_OWNER, SET_MEM_TABLE, configure vrings, SET_FEATURES, SET_VRING_ENABLE (if protocol features negotiated)

Not sure if this a a problem in practice though (and haven't investigated the order in other VMMs). The spec kind of implies this might be the correct order if protocol features are NOT negotiated. (If protocol features are negotiated this doesn't matter and either order should be correct). But maybe I am wrong, the spec is not very precise.

dorindabassey: That order causes the frontend to fail because the backend needs the feature set via SET_FEATURES before it can properly handle vring configuration commands. I checked this already and that order causes VM to hang during device activation and never reaches the interrupt monitor stage, whereas SET_FEATURES before vring configuration works correctly.
BTW, the spec immediate enable rule doesn't apply here. In our case, VHOST_USER_F_PROTOCOL_FEATURES IS negotiated (bit 30 is set in backend_feature_bits), so the immediate enable behavior shouldn't be expected. Vrings are explicitly enabled later via SET_VRING_ENABLE. Also the spec doesn't provide explicit ordering requirements for when SET_FEATURES must be called relative to vring configuration commands.

mtjhrc: Thanks for looking into it. For what it’s worth, passt works with both orders.

> Also the spec doesn't provide explicit ordering requirements for when SET_FEATURES must be called relative to vring configuration commands

Yeah, that's why I was questioning this, it's not clear at all when it should come.

Since this currently works, let's leave it like it is.

src/devices/src/virtio/vhost_user/device.rs
dorindabassey: removed! thanks!

src/devices/src/virtio/vhost_user/device.rs
dorindabassey: added your suggestions and co-authorship, Thanks!

mtjhrc (changes requested):
Couple of small things and then we should be good to go:

Please fix the commit messages a bit. We generally follow the structure:
> subject(s): what has been done

where subject can be e.g. `devices/vhost:` (but it is quite loose on what exactly it has to be).

---

This commit has quite a misleading title and the first sentence:
> Add VhostUserRng device implementation

It sounds like it is implementing the backend. I would expect something along the lines of "vmm: Implement attaching vhost-user devices"

---

Another minor thing, this PR also introduces `krun_disable_implicit_rng`. That one is actually not necessary since we are going for the automatic configuration in the end, so it would be best to remove it from this PR (I meant it as an alternative to the automatic detection), or or at least have it in a separate commit.

dorindabassey:
Fixed! Thank you for the reviews @mtjhrc!

src/devices/src/virtio/vhost_user/device.rs
```diff
@@ -0,0 +1,450 @@
...
+    ActivateError, ActivateResult, DeviceQueue, DeviceState, InterruptTransport, QueueConfig,
+    VirtioDevice,
+};
+
+/// VHOST_USER_F_PROTOCOL_FEATURES (bit 30) is a backend-only feature
+/// that enables vhost-user protocol extensions. It's not a virtio feature.
+const VHOST_USER_F_PROTOCOL_FEATURES: u64 = 1 << 30;
+
+/// Translate a guest physical address to a VMM virtual address.
+fn gpa_to_vmm_va(mem: &GuestMemoryMmap, gpa: u64) -> IoResult<u64> {
```
mtjhrc: Actually you don't need to do this manually `mem.get_host_address` already implements this. I'd say you can just remove the function entirely and use `mem.get_host_address` at the callsites.

dorindabassey: Fixed, Thanks.

dorindabassey:
Hi @mtjhrc and @tylerfanelli PTAL

mtjhrc (dismissed):
LGTM now, thanks!

src/devices/Cargo.toml
```diff
@@ -18,6 +18,7 @@ input = ["zerocopy", "krun_input"]
 virgl_resource_map2 = []
 aws-nitro = []
 test_utils = []
+vhost-user = ["vhost"]
```
tylerfanelli, line 21: Sorry if this was already discussed and I missed it, but why is this an optional feature?

src/libkrun/src/lib.rs
```diff
@@ -1778,6 +1778,92 @@ pub unsafe extern "C" fn krun_set_snd_device(ctx_id: u32, enable: bool) -> i32 {
...
+            let size = *queue_sizes.add(i);
+            if size == 0 {
+                break;
+            }
+            sizes.push(size);
+            i += 1;
+
+            // Safety: prevent infinite loop if user forgets sentinel terminator
+            // Virtio spec allows up to 65536 queues (16-bit queue index: 0-65535)
+            if i >= 65536 {
```
tylerfanelli: Perhaps this should be a constant:

```
VHOST_USER_MAX_QUEUES = 65536
```

src/vmm/src/builder.rs
```diff
@@ -1330,9 +1334,81 @@ fn load_payload(
...
-            let kernel_region = unsafe {
-                MmapRegion::build_raw(kernel_host_addr as *mut u8, kernel_size, 0, 0)
-                    .map_err(StartMicrovmError::InvalidKernelBundle)?
+            #[cfg(feature = "vhost-user")]
+            let use_vhost_user = !_vm_resources.vhost_user_devices.is_empty();
+            #[cfg(not(feature = "vhost-user"))]
+            let use_vhost_user = false;
+
+            let kernel_region = if use_vhost_user {
+                #[cfg(feature = "vhost-user")]
```
tylerfanelli: So it seems the vhost-user feature is optional because it changes how the VM memory is created (memfd in the case of vhost-user). Is this correct? If so, what is stopping us from creating memfd-backed memory in all use-cases?

src/vmm/src/builder.rs
```diff
@@ -1497,10 +1573,76 @@ pub fn create_guest_memory(
...
+    let use_vhost_user = !vm_resources.vhost_user_devices.is_empty();
+    #[cfg(not(feature = "vhost-user"))]
+    let use_vhost_user = false;
+
+    // Add SHM regions before creating guest memory
     arch_mem_regions.extend(shm_manager.regions());
 
-    let guest_mem = GuestMemoryMmap::from_ranges(&arch_mem_regions)
-        .map_err(StartMicrovmError::GuestMemoryMmap)?;
+    let guest_mem = if use_vhost_user {
```
tylerfanelli, line 1620: `use_vhost_user` should probably be overridden to `false` if the feature isn't enabled. This would remove the need for the `unreachable` statements within the `if` statements.

tylerfanelli (dismissed):
LGTM. A few comments that are non-blockers. Great work!

src/vmm/src/builder.rs
dorindabassey: not really, the main reason is to avoid adding the vhost crate dependency for users who don't need vhost-user functionality. Most libkrun users rely on the built-in virtio implementations, so keeping vhost-user optional (like our other features: net, blk, gpu, snd, input) lets users build with only what they need.

 The memfd vs anonymous mmap is more of an implementation detail. You're right that technically nothing stops us from using memfd everywhere but:
  - It consumes file descriptors for each memory region
  - It changes the already existing tested code path for users who don't need vhost-user.
so better to just  keep the simpler anonymous mmap path for users who don't need vhost-user.

src/vmm/src/builder.rs
dorindabassey: Right! but the reason for the current structure is that the memfd code uses FileOffset and other types that are only available with the vhost-user feature, so it won't compile without the inner #[cfg] guards. The `unreachable!()` is just dead code that ensures the if-block has a valid value for both cfg branches.

src/vmm/src/builder.rs
tylerfanelli: Sounds good to me, thanks for the explanation.

I'd venture to say we should remove those optional features and just bring everything under the default build, but that has nothing to do with this PR. Will bring this up in v2 planning.

tylerfanelli (dismissed):
LGTM!

src/devices/src/virtio/vhost_user/device.rs
```diff
@@ -0,0 +1,452 @@
...
+        })?;
+
+        // If protocol features not negotiated, this triggers automatic ring enabling
+        frontend
+            .set_features(backend_feature_bits)
+            .map_err(|e| io::Error::new(ErrorKind::Other, e))?;
+
+        // Create single vring call event file descriptor (backend->guest interrupt)
+        // NOTE: Do NOT use EFD_NONBLOCK here - the monitoring thread needs to block
+        let vring_call_event = EventFd::new(0)?; // Blocking eventfd
```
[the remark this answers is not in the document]
mtjhrc: Intended, there is even comment explicitly explaining this.

mtjhrc:
Merged as part of #642

PR #528, opened 2026-02-04
libkrun: reject HIJACK_UNIX on macOS

slp:
It's not supported on macOS and allowing it just generates confusion.

### DIFF
--- a/src/libkrun/src/lib.rs
+++ b/src/libkrun/src/lib.rs
@@ -2299,6 +2299,11 @@ pub extern "C" fn krun_add_vsock(ctx_id: u32, tsi_features: u32) -> i32 {
         None => return -libc::EINVAL,
     };
 
+    if cfg!(target_os = "macos") && tsi_flags.contains(TsiFlags::HIJACK_UNIX) {
+        error!("TSI hijacking of UNIX sockets is not yet supported on macOS");
+        return -libc::EINVAL;
+    }
+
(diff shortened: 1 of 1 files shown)


mtjhrc:
Could we make this a feature of the library (automatically enabled on Linux in Makefile, disabled on macOS) and then it can be made queryable if this is available by the means of https://github.com/containers/libkrun/pull/525 ?
It also might be nicerer to enforce at a high level what is available on which platform...

slp:
> Could we make this a feature of the library (automatically enabled on Linux in Makefile, disabled on macOS) and then it can be made queryable if this is available by the means of #525 ?

I think turning this into a cargo feature is a little bit overkill. In the context of #525, we'll need to check more than features (from the to of my mind, I can think of kernel formats, which differ between arches), but that's since as long as we have a comment about why something is supported (or not).

mtjhrc:
> I think turning this into a cargo feature is a little bit overkill.

Fair enough, though I don't really see it as a big deal. 

> In the context of #525, we'll need to check more than features (from the to of my mind, I can think of kernel formats, which differ between arches)

Right, I guess... But what different guest/host kernel arches do we even support? And on the same arch, the struct should be stable ABI ...

Anyways I think the correct solution is to eventually use our own custom custom structs for this in the guest (like a TSIv4 guest ABI), which could also allow for longer absolute paths for sockets (resolved in the guest from relative paths).

Issue #526, opened 2026-02-04
TSI network backend appears to be broken on macOS


ihexon:
I configured the TSI network using the following code snippet:
```go
// VSock: disable implicit and add explicit
ret = C.krun_disable_implicit_vsock(C.uint32_t(vm.ctxID))
if ret != 0 {
return fmt.Errorf("krun_disable_implicit_vsock failed with code %d", ret)
}

var vsockFeat = C.uint32_t(0)
if vm.vmc.TSI {
vsockFeat = C.KRUN_TSI_HIJACK_INET | C.KRUN_TSI_HIJACK_UNIX
}

ret = C.krun_add_vsock(C.uint32_t(vm.ctxID), vsockFeat)
if ret != 0 {
return fmt.Errorf("krun_add_vsock failed with code %d", ret)
}
```

When I enter the virtual machine and execute `wget 192.168.1.1` (a router on LAN), libkrun triggers numerous errors:
```
[2026-02-04T06:54:56Z ERROR devices::virtio::vsock::tsi_stream] EventSet::OUT while not connecting
[2026-02-04T06:54:56Z ERROR devices::virtio::vsock::tsi_stream] EventSet::OUT while not connecting
[2026-02-04T06:54:56Z ERROR devices::virtio::vsock::tsi_stream] EventSet::OUT while not connecting
[2026-02-04T06:54:56Z ERROR devices::virtio::vsock::tsi_stream] EventSet::OUT while not connecting
```

macOS version
```
$ sw_vers 
ProductName:            macOS
ProductVersion:         26.2
BuildVersion:           25C56
```

Related issues https://github.com/containers/libkrunfw/issues/105

slp:
`KRUN_TSI_HIJACK_UNIX` is not supported on macOS, so I suspect what's happening on `wget` is that it's trying to access a UNIX socket and getting stuck in there, with the `EventSet::OUT` being unrelated to the actual issue.

In any case, I'm going to add a check to reject `KRUN_TSI_HIJACK_UNIX` on macOS, and fix those `EventSet::OUT` messages, which are pretty ugly.

PR #531, opened 2026-02-04
vsock/tsi: use EDGE_TRIGGERED on EventSet::OUT

slp:
When we're waiting on ProxyStatus::Connecting, we set EventSet::OUT to be informed when the socket has connected. Use EDGE_TRIGGERED along with it so the event is generated just once.

Fixes bogus errors on #526


ihexon:
Hopefully this PR can be merged soon :)

PR #532, opened 2026-02-06
Replace KILLPRIV with KILLPRIV_V2

slp:
While we were announcing it, we never really supported KILLPRIV, so this was breaking some fs tests. This PR stops announcing KILLPRIV support, and implements support for KILLPRIV_V2.

### DIFF
--- a/src/devices/src/virtio/fs/macos/passthrough.rs
+++ b/src/devices/src/virtio/fs/macos/passthrough.rs
@@ -92,35 +92,29 @@ fn item_to_value(item: &[u8], radix: u32) -> Option<u32> {
     }
 }
 
-fn get_xattr_common(buf: &[u8]) -> io::Result<Option<(u32, u32, u32)>> {
+fn get_xattr_common(buf: &[u8]) -> io::Result<(Option<u32>, Option<u32>, Option<u32>)> {
     let mut items = buf.split(|c| *c == b':');
 
     let uid = match items.next() {
-        Some(item) => match item_to_value(item, 10) {
-            Some(item) => item,
-            None => return Ok(None),
-        },
-        None => return Ok(None),
+        Some(item) => item_to_value(item, 10),
+        None => None,
     };
     let gid = match items.next() {
-        Some(item) => match item_to_value(item, 10) {
-            Some(item) => item,
-            None => return Ok(None),
-        },
-        None => return Ok(None),
+        Some(item) => item_to_value(item, 10),
+        None => None,
     };
     let mode = match items.next() {
-        Some(item) => match item_to_value(item, 8) {
-            Some(item) => item,
-            None => return Ok(None),
-        },
-        None => return Ok(None),
+        Some(item) => item_to_value(item, 8),
+        None => None,
     };
 
-    Ok(Some((uid, gid, mode)))
+    Ok((uid, gid, mode))
 }
 
-fn get_xattr_fstat(fd: RawFd, st: bindings::stat64) -> io::Result<Option<(u32, u32, u32)>> {
+fn get_xattr_fstat(
+    fd: RawFd,
+    st: bindings::stat64,
+) -> io::Result<(Option<u32>, Option<u32>, Option<u32>)> {
     let mut buf: Vec<u8> = vec![0; 32];
     let options = if (st.st_mode & libc::S_IFMT) == libc::S_IFLNK {
         libc::XATTR_NOFOLLOW
@@ -139,7 +133,7 @@ fn get_xattr_fstat(fd: RawFd, st: bindings::stat64) -> io::Result<Option<(u32, u
     };
     if res < 0 {
         debug!("fget_xattr error: {res}");
-        return Ok(None);
+        return Ok((None, None, None));
     }
 
     buf.resize(res as usize, 0);
@@ -147,7 +141,10 @@ fn get_xattr_fstat(fd: RawFd, st: bindings::stat64) -> io::Result<Option<(u32, u
     get_xattr_common(&buf)
 }
 
-fn get_xattr_lstat(path: &CString, st: bindings::stat64) -> io::Result<Option<(u32, u32, u32)>> {
+fn get_xattr_lstat(
+    path: &CString,
+    st: bindings::stat64,
+) -> io::Result<(Option<u32>, Option<u32>, Option<u32>)> {
     let mut buf: Vec<u8> = vec![0; 32];
     let options = if (st.st_mode & libc::S_IFMT) == libc::S_IFLNK {
         libc::XATTR_NOFOLLOW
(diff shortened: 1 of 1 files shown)


slp:
Tested with [pjdfstest](https://github.com/slp/pjdfstest) on both Linux and macOS.

src/devices/src/virtio/fs/linux/passthrough.rs
```diff
@@ -118,10 +118,75 @@ macro_rules! scoped_cred {
 scoped_cred!(ScopedUid, libc::uid_t, libc::SYS_setresuid);
 scoped_cred!(ScopedGid, libc::gid_t, libc::SYS_setresgid);
 
+pub struct ScopedCaps {
```
mtjhrc, line 122:
```suggestion

#[must_use]
pub struct ScopedCaps {
```
(This makes it so you get a warning if you don't explicitly use the value - like functions returning a  Result).

src/devices/src/virtio/fs/macos/passthrough.rs
```diff
@@ -732,11 +768,31 @@ impl PassthroughFs {
...
+        if (flags & libc::O_TRUNC) != 0 && kill_priv {
+            let fd = file.read().unwrap().as_raw_fd();
+            let ihandle = InodeHandle::Fd(fd);
+
+            remove_security_capability(&ihandle);
+
+            if let Ok(st) = fstat(fd, false) {
+                let new_mode = clear_suid_sgid(st.st_mode as u32);
+                if new_mode != st.st_mode as u32 {
+                    let _ = set_xattr_stat(&ihandle, Some(st), None, Some(new_mode));
```
dorindabassey: is ignoring the error here intentional? probably should propagate the error.

src/devices/src/virtio/fs/macos/passthrough.rs
```diff
@@ -1377,7 +1479,25 @@ impl FileSystem for PassthroughFs {
...
+            let fd = f.as_raw_fd();
+            let ihandle = InodeHandle::Fd(fd);
+
+            remove_security_capability(&ihandle);
+
+            if let Ok(st) = fstat(fd, false) {
+                let new_mode = clear_suid_sgid(st.st_mode as u32);
+                if new_mode != st.st_mode as u32 {
+                    // Update mode in xattr
+                    let _ = set_xattr_stat(&ihandle, Some(st), None, Some(new_mode));
```
dorindabassey: ditto

src/devices/src/virtio/fs/macos/passthrough.rs
```diff
@@ -176,42 +180,63 @@ fn is_valid_owner(owner: Option<(u32, u32)>) -> bool {
...
                 (uid, gid)
             }
-            None => orig_owner,
+            None => (orig_uid, orig_gid),
         };
 
-        (new_owner, mode.unwrap_or(orig_mode))
+        let mut buf = String::new();
+        if let Some(uid) = uid {
+            buf.push_str(&format!("{uid}"));
```
mtjhrc, line 219: This is a bit nicer + less memory allocation probably:
```suggestion
            write!(&mut buf, "{uid}");
```

src/devices/src/virtio/fs/macos/passthrough.rs
```diff
@@ -176,42 +180,63 @@ fn is_valid_owner(owner: Option<(u32, u32)>) -> bool {
...
+        } else {
+            buf.push('x');
+        }
+        if let Some(gid) = gid {
+            buf.push_str(&format!(":{gid}:"));
```
mtjhrc, line 224:
```suggestion
            write!(&mut buf, ":{gid}:");
```

src/devices/src/virtio/fs/macos/passthrough.rs
```diff
@@ -176,42 +180,63 @@ fn is_valid_owner(owner: Option<(u32, u32)>) -> bool {
...
+        } else {
+            buf.push_str(":x:");
+        }
+        if let Some(mode) = mode {
+            buf.push_str(&format!("0{:o}", mode));
```
mtjhrc, line 229:
```suggestion
            write!(&mut buf, "0{mode:o}");
```

src/devices/src/virtio/fs/macos/passthrough.rs
```diff
@@ -176,42 +180,63 @@ fn is_valid_owner(owner: Option<(u32, u32)>) -> bool {
...
+        } else if let Some(orig_mode) = orig_mode {
+            buf.push_str(&format!("0{:o}", orig_mode));
```
mtjhrc, line 231:
```suggestion
            write!(&mut buf, "0{orig_mode:o}");
```

src/devices/src/virtio/fs/macos/passthrough.rs
slp: This is a very good question. Strictly speaking, not being able to clear suid/sgid isn't fatal. On the other hand, the guest isn't even aware this is happening, and as such it can't deal properly with an error.

I think the most reasonable option is adding an "error!" here, so we don't bother the guest but we can still warn the user about something misbehaving.

slp: I've just updated the PR to issue an error message.

src/devices/src/virtio/fs/linux/passthrough.rs
slp: Applied, thanks!

src/devices/src/virtio/fs/macos/passthrough.rs
```diff
@@ -97,74 +92,83 @@ fn item_to_value(item: &[u8], radix: u32) -> Option<u32> {
...
+        libc::XATTR_NOFOLLOW
+    } else {
+        0
+    };
+    let res = unsafe {
+        libc::fgetxattr(
+            fd,
+            XATTR_KEY.as_ptr() as *const i8,
+            buf.as_mut_ptr() as *mut libc::c_void,
+            64,
```
pftbest, line 129: Buffer size passed here is 64, but the actual Vec<u8> is only 32 bytes if I read this correctly.

pftbest: Sorry, I missed this was already merged

slp: @pftbest Good catch! We should change those hard coded numbers with a constant = 64. Since you found it, do you want to create a PR fixing it or should I do it myself?

pftbest: @slp Please go ahead, I won't have time to fix this today.

slp: Ack, thanks anyways!

Issue #539, opened 2026-02-09
macOS virtio-fs fatal runtime error: IO Safety violation: owned file descriptor already closed, aborting


pftbest:
I was debugging the issue I observed on krunvm 0.2.6 where sometimes files would become invisible inside some directories. (https://github.com/containers/libkrunfw/issues/111#issuecomment-3873068634)

I compiled libkrun with additional runtime checks
```toml
[profile.release]
debug-assertions = true
overflow-checks = true
```
And got this error when I run `apt update` in debian:
```
libkrun$ ./examples/chroot_vm /Volumes/krunvm/root/vfs/dir/46be82fe112f11c881bebd11207da6ee88c5c9a2d067cad4231b6cfa3b00f1bb /bin/bash
root@localhost:/#
root@localhost:/#
root@localhost:/# apt update
fatal runtime error: IO Safety violation: owned file descriptor already closed, aborting
zsh: abort
```

And this stacktrace
```
std[7af367ec0ba611a7]::process::abort
<alloc[e101761a117f5a99]::sync::Arc<devices[78dfc7782f1bbf32]::virtio::fs::macos::passthrough::HandleData>>::drop_slow
<devices[78dfc7782f1bbf32]::virtio::fs::macos::passthrough::PassthroughFs as devices[78dfc7782f1bbf32]::virtio::fs::filesystem::FileSystem>::releasedir
<devices[78dfc7782f1bbf32]::virtio::fs::server::Server<devices[78dfc7782f1bbf32]::virtio::fs::macos::passthrough::PassthroughFs>>::releasedir
<devices[78dfc7782f1bbf32]::virtio::fs::server::Server<devices[78dfc7782f1bbf32]::virtio::fs::macos::passthrough::PassthroughFs>>::handle_message
```

slp:
Thanks for the report. If possible, please give #540 a try.

PR #540, opened 2026-02-10
virtio/fs/macos: duplicate fd for fdopendir

slp:
fdopendir() takes ownership of the fd passed as argument to it, so passing the fd of the HandleData->File will cause the fd to be closed twice, first when closedir() is called and second when Arc<HandleData> is dropped.

If the library is built with debug-assertions enabled, this issue will trigger an error like:

fatal runtime error: IO Safety violation: owned file descriptor already closed, aborting

To avoid this issue, let's use dup() to obtain a new descriptor from HandleData->File to be donated to fdopendir().

Fixes: #539

### DIFF
--- a/src/devices/src/virtio/fs/macos/passthrough.rs
+++ b/src/devices/src/virtio/fs/macos/passthrough.rs
@@ -657,9 +657,17 @@ impl PassthroughFs {
         let mut ds = data.dirstream.lock().unwrap();
 
         let dir_stream = if ds.stream == 0 {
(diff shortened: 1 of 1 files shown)


pftbest:
Checked this PR on my system, the assertion no longer happens, so the change is good. But the ghost files still appear so it is probably a separate issue.

slp:
Let's put this one on hold, because I'm preparing another one that fixes both this and the ghost files.

slp:
Superseded by #542

PR #544, opened 2026-02-11
fs/macos: fix ghost files are double fd closing

slp:
Ghost files were caused by a bug introduced in #513, while the double fd closing came from using the HandleData->File fd with fdopendir, which takes ownership of the fd.

### DIFF
--- a/src/devices/src/virtio/fs/macos/passthrough.rs
+++ b/src/devices/src/virtio/fs/macos/passthrough.rs
@@ -657,9 +657,17 @@ impl PassthroughFs {
         let mut ds = data.dirstream.lock().unwrap();
 
         let dir_stream = if ds.stream == 0 {
-            let dir = unsafe { libc::fdopendir(data.file.write().unwrap().as_raw_fd()) };
-            if dir.is_null() {
+            // fdopendir() takes ownership of the fd, so we need to obtain a new one
+            // to be donated.
+            let newfd = unsafe { libc::dup(data.file.write().unwrap().as_raw_fd()) };
+            if newfd < 0 {
                 return Err(linux_error(io::Error::last_os_error()));
+            }
+            let dir = unsafe { libc::fdopendir(newfd) };
+            if dir.is_null() {
+                let err = io::Error::last_os_error();
+                let _ = unsafe { libc::close(newfd) };
+                return Err(linux_error(err));
             }
             ds.stream = dir as u64;
             dir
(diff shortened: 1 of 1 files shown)


pftbest:
Works on my side. Performance only 6% lower than `opendir` version in my tests, so that is ok too.

src/devices/src/virtio/fs/macos/passthrough.rs
```diff
@@ -825,12 +838,16 @@ impl PassthroughFs {
...
+                warn!(
+                    "Couldn't grab a file descriptor for file \"{}\": {err}",
+                    name.to_string_lossy()
+                );
+                None
+            }
+        };
 
         // Safe because this doesn't modify any memory and we check the return value.
         let res = unsafe { libc::unlinkat(fd, name.as_ptr(), flags) };
```
dorindabassey, line 853: Hmm, If `grab_unlinked_fd` succeeded but `unlinkat` fails here, there could be a potential fd leak from the already opened fd.

src/devices/src/virtio/fs/macos/passthrough.rs
```diff
@@ -841,6 +858,12 @@ impl PassthroughFs {
...
         if res == 0 {
+            if let Some(unlinked_fd) = unlinked_fd {
+                if let Err(err) = self.store_unlinked_fd(unlinked_fd) {
+                    unsafe { libc::close(unlinked_fd) };
+                    warn!("Couldn't store unlinked fd \"{}\": {err}", unlinked_fd);
+                }
+            }
             Ok(())
         } else {
             Err(linux_error(err))
```
dorindabassey, line 872: as per the previous issue
```suggestion
            if let Some(unlinked_fd) = unlinked_fd {
                if let Err(err) = self.store_unlinked_fd(unlinked_fd) {
                    unsafe { libc::close(unlinked_fd) };
                    warn!("Couldn't store unlinked fd \"{}\": {err}", unlinked_fd);
                }
            }
            Ok(())
        } else {
            if let Some(unlinked_fd) = unlinked_fd {                                                                        
                unsafe { libc::close(unlinked_fd) };                                                                        
            } 
            Err(linux_error(err))
```

slp: Nice catch, thanks!

slp:
Applied @dorindabassey's suggestion (thanks!) squashing it in the previous commit (adding her as Co-author) and rebased the series on HEAD.

mtjhrc (approved):
LGTM, thanks!

PR #546, opened 2026-02-12
vsock/tsi: fix listen() with backlog > 128 fails with EINVAL

ihexon:
When a guest program calls `listen(fd, 256)` (or any backlog > 128), it gets EINVAL immediately.

A python POC:：
```python
>>> import socket; s=socket.socket(); s.bind(("0.0.0.0",8080)); s.listen(256); input()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
OSError: [Errno 22] Invalid argument
```

The real problem is that the value of backlog is not being properly constrained and passed:
- func [`try_listen()`](https://github.com/containers/libkrun/blob/d4c0e9e0f551229cfe2244a27df150d190a2607a/src/devices/src/virtio/vsock/tsi_stream.rs#L197) passing the `req.backlog` to `Backlog::new(req.backlog)` https://github.com/containers/libkrun/blob/d4c0e9e0f551229cfe2244a27df150d190a2607a/src/devices/src/virtio/vsock/tsi_stream.rs#L244

- `Backlog::new`(from nix 0.30.1) performs some mandatory checks.:
```rust
if !(MIN..=Self::MAXCONN.0).contains(&val) {
            return Err(Errno::EINVAL);
        }
```

`MAXCONN` defined as:

```rust
pub const MAXCONN: Self = Self(libc::SOMAXCONN); // In MacOS, libc::SOMAXCONN == 128
```

Any program with a listen backlog over `128` (like Nginx 512, Apache 511, or Go using `syscall.SOMAXCONN`) will `return Err(Errno::EINVAL)` under TSI network.

This PR fixes the issue by referencing the kernel's [`__sys_listen_socket`](https://github.com/torvalds/linux/blob/37a93dd5c49b5fda807fd204edf2547c3493319c/net/socket.c#L1918), clamp the backlog to `[0, SOMAXCONN]` before passing it `Backlog::new()`, which is acceptable.

### DIFF
--- a/src/devices/src/virtio/vsock/tsi_stream.rs
+++ b/src/devices/src/virtio/vsock/tsi_stream.rs
@@ -241,7 +241,10 @@ impl TsiStreamProxy {
                 // it's possible the userspace application can't do it itself.
                 self.unixsock_path = unixsock_path;
 
-                match Backlog::new(req.backlog) {
+                // Clamp backlog to SOMAXCONN, mirroring Linux kernel's __sys_listen behavior.
+                // The nix crate's Backlog::new() rejects values above SOMAXCONN with EINVAL.
+                let clamped_backlog = req.backlog.clamp(0, libc::SOMAXCONN);
+                match Backlog::new(clamped_backlog) {
                     Ok(backlog) => match listen(&self.fd, backlog) {
                         Ok(_) => {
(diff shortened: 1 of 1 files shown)


slp (approved):
Nice catch, thanks!

PR #548, opened 2026-02-13
virtio/fs/macos: preload directory entries to avoid calling telldir()

pftbest:
This PR depends on https://github.com/containers/libkrun/pull/544 and includes its commits. I can rebase when 544 is merged.

I was doing some performance measurements on filesystem operations and noticed that `telldir()` call is very heavy on macOS.

In some of my synthetic tests it could take up to 40% of total CPU cycles on the host:
Screenshot 2026-02-12 at 22 02 57

The proposed fix is to avoid calling `telldir()` at all and preload all dir entries ahead of time on the first call to `do_readdir()`.

The memory impact: In the worst-case scenario each entry would be max 1K in size. A plain directory with 100k files would require 100M in host memory.

However each call to telldir() also allocates memory internally via malloc() and stores it inside DIR stream. We call it on each dir entry so the actual difference should be smaller than that in practice.

Unfortunately after I made this change I did not see a significant time savings on the guest side. Yes the host cycles were reduced but the most of the time is spent waiting on communication between host and guest, so I only see 5-10% improvement on the guest side.

For example running `time make clean` on the linux kernel source code went from around 12s to 11s.

<details><summary>Trace for make clean without this patch. do_readdir is at 24% cycles</summary>
<p>
Screenshot 2026-02-13 at 12 44 08
</p>
</details> 

<details><summary>Trace for make clean with this patch do_readdir is at 9% cycles</summary>
<p>
Screenshot 2026-02-13 at 12 39 36

</p>
</details>

### DIFF
--- a/src/devices/src/virtio/fs/macos/passthrough.rs
+++ b/src/devices/src/virtio/fs/macos/passthrough.rs
@@ -17,6 +17,7 @@ use std::sync::{Arc, Mutex, RwLock};
 use std::time::Duration;
 
 use crossbeam_channel::{unbounded, Sender};
+use nix::errno::Errno;
 use utils::worker_message::WorkerMessage;
 
 use crate::virtio::fs::filesystem::SecContext;
@@ -59,9 +60,92 @@ enum InodeHandle {
     VolPath(CString),
 }
 
+struct CachedDirEntry {
+    ino: bindings::ino64_t,
+    name: Box<[u8]>,
+    type_: u8,
+}
+
 struct DirStream {
-    stream: u64,
-    offset: i64,
+    entries: Vec<CachedDirEntry>,
+    ready: bool,
+}
+
+impl DirStream {
+    fn new() -> Self {
+        Self {
+            entries: Vec::new(),
+            ready: false,
+        }
+    }
+
+    fn get_entry<'a>(&'a self, offset: u64) -> Option<DirEntry<'a>> {
+        self.entries.get(offset as usize).map(|e| DirEntry {
+            ino: e.ino,
+            // offset points to the next entry, not the current one
+            offset: offset + 1,
+            type_: u32::from(e.type_),
+            name: &e.name,
+        })
+    }
+
+    fn fill_from_fd(&mut self, fd: RawFd) -> io::Result<()> {
+        // fdopendir() takes ownership of the fd, so we need to obtain a new one
+        // to be donated.
+        let newfd = unsafe { libc::dup(fd) };
+        if newfd < 0 {
+            return Err(linux_error(io::Error::last_os_error()));
+        }
+        let dir = unsafe { libc::fdopendir(newfd) };
+        if dir.is_null() {
+            let err = io::Error::last_os_error();
+            let _ = unsafe { libc::close(newfd) };
+            return Err(linux_error(err));
+        }
+
+        loop {
+            // To detect if error happened in readdir we should clear errno
+            // before the call and then verify it after
+            Errno::clear();
+            let dentry = unsafe { libc::readdir(dir) };
+            if dentry.is_null() {
+                let errno = Errno::last_raw();
+                if errno != 0 {
+                    let err = io::Error::from_raw_os_error(errno);
+                    let _ = unsafe { libc::closedir(dir) };
+                    // Error happened in readdir, but we keep the entries we
+                    // already read to handle the partial read.
+                    return Err(linux_error(err));
+                }
+                break;
+            }
+            // SAFETY: dentry is not null.
+            // We trust macOS to return correct number of bytes for the name
+            // length. The lifetime of a slice does not escape the unsafe block
+            // as we copy the data into box right away.
+            let name = unsafe {
+                let name_len = usize::from((*dentry).d_namlen);
+                let name_ptr = (*dentry).d_name.as_ptr().cast();
+                let name = std::slice::from_raw_parts(name_ptr, name_len);
+
+                if name == b"." || name == b".." {
+                    continue;
+                }
+                Box::<[u8]>::from(name)
+            };
+
+            // SAFETY: dentry is not null.
+            let ino = unsafe { (*dentry).d_ino };
+            // SAFETY: dentry is not null. The entry types use the same
+            // exact constants (`libc::DT_*`) on macOS, Linux, and FUSE.
+            let type_ = unsafe { (*dentry).d_type };
+
+            self.entries.push(CachedDirEntry { ino, name, type_ });
+        }
+
+        unsafe { libc::closedir(dir) };
+        Ok(())
+    }
 }
 
 struct HandleData {
@@ -635,7 +719,7 @@ impl PassthroughFs {
         inode: Inode,
         handle: Handle,
         size: u32,
-        offset: u64,
+        mut offset: u64,
         mut add_entry: F,
     ) -> io::Result<()>
     where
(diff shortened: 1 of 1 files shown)


src/devices/src/virtio/fs/macos/passthrough.rs
```diff
@@ -59,9 +59,76 @@ enum InodeHandle {
...
+            let err = io::Error::last_os_error();
+            let _ = unsafe { libc::close(newfd) };
+            return Err(linux_error(err));
+        }
+
+        loop {
+            let dentry = unsafe { libc::readdir(dir) };
+            if dentry.is_null() {
+                break;
+            }
```
dorindabassey, line 121: here, `readdir()` can return NULL for both EOF and errors, an error could occur if the directory is deleted while being read. So maybe check errno to distinguish errors from EOF, then handle the error.

src/devices/src/virtio/fs/macos/passthrough.rs
```diff
@@ -59,9 +59,76 @@ enum InodeHandle {
...
+            offset: offset + 1,
+            type_: u32::from(e.type_),
+            name: &e.name,
+        })
+    }
+
+    fn fill_from_fd(&mut self, fd: RawFd) -> io::Result<()> {
+        // fdopendir() takes ownership of the fd, so we need to obtain a new one
+        // to be donated.
+        let newfd = unsafe { libc::dup(fd) };
```
dorindabassey, line 95: humm, they're a lot of unsafe blocks without safety comments.

src/devices/src/virtio/fs/macos/passthrough.rs
```diff
@@ -59,9 +59,76 @@ enum InodeHandle {
...
+        Self {
+            entries: Vec::new(),
+            ready: false,
+        }
+    }
+
+    fn get_entry<'a>(&'a self, offset: u64) -> Option<DirEntry<'a>> {
+        self.entries.get(offset as usize).map(|e| DirEntry {
+            ino: e.ino,
+            offset: offset + 1,
```
dorindabassey, line 86: There's a risk of overflow here, although very unlikely (as it would require 2^64 directory entries), but this could cause unexpected behaviour.
```suggestion
            offset: offset.saturating_add(1),
```

mtjhrc: I'm confused why this even does `offset+1` in the first place. I feel like it  should just be `offset` directly, no?

(The original code was doing `ds.offset+1`, because `ds.offset` was saved before calling the `readdir` for the entry)

pftbest: @mtjhrc The offset we return here is opaque to guest, it's basically a cookie that can be used to start again from that point. So the guest will not do `+1` itself when calling `readdir` again, it will pass the value unchanged from the entry they got, or `0` if they want to read from the start.
If we return `offset` directly we will run into conflict on the first entry, because 0 would mean both start from scratch or start after the first entry.

Also I don't think this addition could ever overflow, as we only do +1 in case we successfully got an entry from Vec, and all modern systems only use 48 bits for ram sizes.

src/devices/src/virtio/fs/macos/passthrough.rs
pftbest: @dorindabassey The original code did not check for errors, so I didn't add it too, but maybe we should. The only problem here is what to do with partial reads when we get an error here but we already read some of the entries. We can either drop them all or somehow still continue and pass them to add_entry. Both options have downsides, and the latter is also harder to implement because it requires we store the error until a next call to `do_readdir` from the guest.

src/devices/src/virtio/fs/macos/passthrough.rs
mtjhrc: Ahh, that the makes sense now, so the offset kind of points at the next entry, not the current one, thanks for the explanation.

> we only do +1 in case we successfully got an entry from Vec, and all modern systems only use 48 bits for ram sizes

makes sense, I wouldn't worry about it

src/devices/src/virtio/fs/macos/passthrough.rs
mtjhrc: > The original code did not check for errors, so I didn't add it too, but maybe we should. 

The manpage says this returns the same errors as `getdirentries` which would mean the only relevant error for us would be a general `EIO` (but I wouldn't trust the manpage fully). 
I think this is quite unlikely to fail, so for now I think it's good enough to just log the error.

src/devices/src/virtio/fs/macos/passthrough.rs
dorindabassey: > Also I don't think this addition could ever overflow, as we only do +1 in case we successfully got an entry from Vec, and all modern systems only use 48 bits for ram sizes.

Ok in that case the overflow is practically impossible.

src/devices/src/virtio/fs/macos/passthrough.rs
dorindabassey: I get your point about the partial read problem, Since this is called during `fill_from_fd()` which happens on first `do_readdir()` call, I'd suggest to just return the error immediately, we don't have to cache anything
So if error occurs, the cache stays empty and guest gets error immediately. but you'll need to set errno to 0 before calling `readdir()`
```suggestion
            unsafe {
                *libc::__error() = 0;
            }
            let dentry = unsafe { libc::readdir(dir) };
            if dentry.is_null() {
                let err = io::Error::last_os_error();
                if err.raw_os_error() != Some(0) {
                    return Err(linux_error(err));
                }
                break;
            }
```

pftbest: Ok, I implemented it like this. If we got the error right away and no valid entries, then return the error to the guest directly. If we got some number of valid entries before the error, then we just log the error and return Ok to the guest with the entries. I think this should handle most of the cases correctly.

@dorindabassey I used Errno from `nix` crate to check for the error to reduce amount of unsafe blocks.

src/devices/src/virtio/fs/macos/passthrough.rs
pftbest: @dorindabassey I added the safety comments to all non FFI related unsafe blocks, please check. Adding them on direct libc calls seems redundant, but I can add them too if really needed.

dorindabassey: ok that's fine.

slp:
I think gemini's both suggestions are quite reasonable.

pftbest:
@slp  Sorry, but I don't agree with its conclusion for the first one. I specifically rely on the write lock to be locked for the full duration of the `fill_from_fd` function because it updates internal position of the original fd. All duped fd share the same position internally so running 2 instances of `fill_from_fd` on the same fd will lead to corrupted data.

I'm not quite sure if libkrun uses single thread or multiple threads to handle fs requests, but even if it's single threaded then `read()` vs `write()` doesn't matter and if it's multiple threads then `write()` is correct here.

Second suggestion I can apply, but will need to move the safety comment too.

mtjhrc:
> @slp  I specifically rely on the write lock to be locked for the full duration of the `fill_from_fd` function because it updates internal position of the original fd. All duped fd share the same position internally so running 2 instances of `fill_from_fd` on the same fd will lead to corrupted data.

Makes sense

> Second suggestion I can apply, but will need to move the safety comment too.

Looking at it again, the way you have written it is maybe safer (it is very clear what the lifetime of the slice is). 

Feel free to ignore Gemini.

PR #549, opened 2026-02-17
aws-nitro: Separate guest kernel module loading, updates

tylerfanelli:


tylerfanelli:
cc @jakecorrenti

init/aws-nitro/main.c
```diff
@@ -447,6 +406,11 @@ int main(int argc, char *argv[])
         return -errno;
     }
 
+    // Initialize the configured kernel modules.
+    ret = mods_load();
+    if (ret < 0)
+        goto out;
+
```
mz-pdm, line 414: It would be worth to explain in the commit message why the module loading is done earlier now.

init/aws-nitro/mod.c
```diff
@@ -0,0 +1,95 @@
...
+    if (ret < 0) {
+        fprintf(stderr, "unlink module %s\n", path);
+        return -errno;
+    }
+
+    return 0;
+}
+
+/*
+ * Load the conifgured kernel modules.
```
mz-pdm: s/conifgured/configured/

init/aws-nitro/mod.c
```diff
@@ -0,0 +1,95 @@
...
+static int mod_load(const char *path)
+{
+    int fd, ret;
+
+    // Open and load the kernel module.
+    fd = open(path, O_RDONLY | O_CLOEXEC);
+    if (fd < 0 && errno == ENOENT)
+        return 0;
+    else if (fd < 0) {
+        fprintf(stderr, "open module %s\n", path);
```
mz-pdm: Switching from perror to fprintf omits printing the error corresponding to `errno`, right? Is it intentional here?

Makefile
mz-pdm, line 31: When on it: There is indentation mismatch in the file list (not visible here), some lines are indented by spaces, some by tabs -- could it be fixed?

init/aws-nitro/main.c
```diff
@@ -25,6 +24,7 @@
 #include "include/archive.h"
 #include "include/args_reader.h"
 #include "include/fs.h"
+#include "include/mod.h"
 
 #define finit_module(fd, param_values, flags)                                  \
     (int)syscall(__NR_finit_module, fd, param_values, flags)
```
mz-pdm: I think it's no longer used here.

init/aws-nitro/mod.c
```diff
@@ -0,0 +1,95 @@
...
+#include <fcntl.h>
+#include <stddef.h>
+#include <stdio.h>
+#include <string.h>
+#include <sys/syscall.h>
+#include <unistd.h>
+
+#include "include/mod.h"
+
+#define KRUN_LINUX_MODS_DIR_NAME "/krun_linux_mods"
```
mz-pdm, line 14: I'm confused about the directory, the commit message mentions /root.

init/aws-nitro/mod.c
```diff
@@ -0,0 +1,95 @@
...
+{
+    char path[MOD_FILE_NAME_BUF_SIZE + strlen(KRUN_LINUX_MODS_DIR_NAME) + 1];
+    struct dirent *entry;
+    int ret, valid;
+    DIR *dir;
+
+    ret = 0;
+
+    dir = opendir(KRUN_LINUX_MODS_DIR_NAME);
+    if (dir) {
```
mz-pdm: Would it be worth to print the error if opendir fails (maybe for a reason other than the directory doesn't exist)?

Makefile
```diff
@@ -140,7 +140,7 @@ endif
 
 AWS_NITRO_INIT_BINARY= init/aws-nitro/init
 $(AWS_NITRO_INIT_BINARY): $(AWS_NITRO_INIT_SRC)
-	$(CC) -O2 -static -Wall $(AWS_NITRO_INIT_LD_FLAGS) -o $@ $(AWS_NITRO_INIT_SRC) $(AWS_NITRO_INIT_LD_FLAGS)
+	$(CC) -O2 -static -s -Wall $(AWS_NITRO_INIT_LD_FLAGS) -o $@ $(AWS_NITRO_INIT_SRC) $(AWS_NITRO_INIT_LD_FLAGS)
```
mz-pdm, line 144: LGTM.

init/aws-nitro/device/app_stdio_output.c
```diff
@@ -40,7 +40,7 @@ int app_stdio_output(unsigned int vsock_port)
     ret = setsockopt(sock_fd, AF_VSOCK, SO_VM_SOCKETS_CONNECT_TIMEOUT,
                      (void *)&timeval, sizeof(struct timeval));
     if (ret < 0) {
-        perror("unable to connect to host socket");
+        perror("unable to set application output vsock timeout");
```
mz-pdm, line 43: LGTM.

init/aws-nitro/device/app_stdio_output.c
```diff
@@ -81,4 +81,4 @@ void app_stdio_close(void)
         close(APP_STDIO_OUTPUT_VSOCK_FD);
         APP_STDIO_OUTPUT_VSOCK_FD = -1;
     }
-}
\ No newline at end of file
+}
```
mz-pdm, line 84: Unrelated but OK for me.

src/aws_nitro/src/enclave/proxy/proxies/output.rs
```diff
@@ -48,11 +48,7 @@ impl OutputProxy {
 impl DeviceProxy for OutputProxy {
     /// Enclave argument of the proxy.
     fn arg(&self) -> Option<EnclaveArg<'_>> {
-        // The enclave only needs to be made aware that it is to be run in debug mode.
-        match self.debug {
-            true => Some(EnclaveArg::Debug),
-            false => None,
-        }
+        Some(EnclaveArg::AppOutput)
```
mz-pdm: AppOutput arg is provided whether in debug mode or not?

src/aws_nitro/src/enclave/mod.rs
```diff
@@ -239,6 +248,17 @@ impl NitroEnclave {
...
     }
+
+    /// Start the debug output thread, read from the serial console.
+    fn start_console_debug(&self, cid: u32) -> Result<(), proxy::Error> {
+        let mut output = OutputProxy::new(&self.output_path, true)?;
+        let mut vsock_rcv = output.vsock(cid)?;
+        let _: JoinHandle<Result<(), proxy::Error>> = thread::spawn(move || loop {
+            output.rcv(&mut vsock_rcv)?;
+        });
+
```
mz-pdm, line 261: Could this change of the output handling be explained in the commit message?

init/aws-nitro/mod.c
```diff
@@ -0,0 +1,95 @@
...
+        while ((entry = readdir(dir)) != NULL) {
+            /*
+             * Ignore the "." and ".." directory entries, as they are not kernel
+             * modules.
+             */
+            valid = 1;
+            valid &= !(strcmp(entry->d_name, ".") == 0);
+            valid &= !(strcmp(entry->d_name, "..") == 0);
```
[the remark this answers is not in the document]
mz-pdm: I agree with the AI here.

init/aws-nitro/mod.c
```diff
@@ -0,0 +1,95 @@
...
+
+            if (!valid)
+                continue;
+
+            // Copy the full path of the module file.
+            sprintf(path, "%s/%s", KRUN_LINUX_MODS_DIR_NAME, entry->d_name);
```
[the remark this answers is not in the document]
mz-pdm: While there may be not module names longer than 255 characters, we are dealing with non-validated input here. Any reason not to use a safe call here?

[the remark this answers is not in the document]
tylerfanelli: No, fixed in latest commit.

init/aws-nitro/mod.c
tylerfanelli: It's more that we can format the string and print the exact path that we were unable to initialize, although it would be useful to print the errno as well. Latest change adds the errno to the message as well.

src/aws_nitro/src/enclave/proxy/proxies/output.rs
tylerfanelli: The arg is not added in non-debug mode, so any time this is called, it will be in app output mode.

init/aws-nitro/main.c
tylerfanelli: I don't quite follow, it's only done earlier because that's likely the first thing the initrd should do (prepare the kernel) before building the user environment?

init/aws-nitro/main.c
tylerfanelli: The `#include`? It's used for the `mods_load` definition, which is declared in `include/mod.h`.

init/aws-nitro/mod.c
tylerfanelli: Yes, that was a typo. Fixed in latest change.

Makefile
tylerfanelli: Sure, I can address that in another PR.

init/aws-nitro/mod.c
tylerfanelli: Sure, added in latest revision.

src/aws_nitro/src/enclave/mod.rs
tylerfanelli: I've instead added an explanation within the comment describing the `start_console_debug` function.

init/aws-nitro/main.c
mz-pdm: There are questions that pop up when seeing that the module loading is moved before console_init: Is it because some module may be needed to init the console properly? And can we miss some messages from the module loading due to this move?

Makefile
mz-pdm: Thanks.

init/aws-nitro/main.c
mz-pdm: No, I mean the original ``#define finit_module`` (in main.c).

init/aws-nitro/mod.c
```diff
@@ -0,0 +1,96 @@
...
+{
+    char path[MOD_FILE_NAME_BUF_SIZE + sizeof(KRUN_LINUX_MODS_DIR_NAME) + 1];
+    struct dirent *entry;
+    int ret;
+    DIR *dir;
+
+    ret = 0;
+
+    dir = opendir(KRUN_LINUX_MODS_DIR_NAME);
+    if (dir) {
```
jakecorrenti: nitpick: everywhere else we explicitly check if a pointer != NULL, and we should do that here to maintain consistency

init/aws-nitro/mod.c
```diff
@@ -0,0 +1,96 @@
...
+                continue;
+
+            // Copy the full path of the module file.
+            snprintf(path, sizeof(path), "%s/%s", KRUN_LINUX_MODS_DIR_NAME,
+                     entry->d_name);
+            ret = mod_load(path);
+            if (ret < 0)
+                break;
+        }
+        closedir(dir);
```
jakecorrenti, line 90: should we print an error if `closedir()` fails?

init/aws-nitro/mod.c
```diff
@@ -0,0 +1,96 @@
...
+    int fd, ret;
+
+    // Open and load the kernel module.
+    fd = open(path, O_RDONLY | O_CLOEXEC);
+    if (fd < 0 && errno == ENOENT)
+        return 0;
+    else if (fd < 0) {
+        fprintf(stderr, "open module %s (errno %d)\n", path, errno);
+        return -errno;
+    }
```
jakecorrenti: readability wise I think this is a little bit better so you're not checking `fd < 0` more than once. Up to you, though.
```suggestion
    if (fd < 0) {
        if (errno == ENOENT)
          return 0;
          
        fprintf(stderr, "open module %s (errno %d)\n", path, errno);
        return -errno;
    }
```

src/aws_nitro/src/enclave/proxy/proxies/output.rs
mz-pdm: Did you mean "The arg is not added in **debug** mode"? Still, looks error-prone to me to rely on that implicit assumption; I'd add some protection here like some check that ``debug`` is indeed not set. But up to you, just for my understanding (or do I misunderstand something)?

init/aws-nitro/mod.c
tylerfanelli: Ehh, I don't think it's necessary. We soon switch the rootfs anyway.

tylerfanelli:
> I find the change from `debug` -> `app_output` a little confusing. The output from the application should always be present, and specifying `debug` was specifically to determine whether we should include the dmesg output. Conditionally, reading it as "if we're not in debug mode, do this..." made more sense to me than "if we want app output (which is functionally always true), do this". If I'm in the minority on that I'm not going to block it from going in as is.

The problem is that debug output needed to be printed earlier, and couldn't be a part of the normal proxy list, as if there's a problem we need to see debug output to find out where. We were unable to do this before because of all other proxies requiring initialization before debug out was viewable.

init/aws-nitro/main.c
tylerfanelli: Thanks, removed.

src/aws_nitro/src/enclave/proxy/proxies/output.rs
tylerfanelli: Sure, I've added that check.

init/aws-nitro/main.c
tylerfanelli: Thanks, I've added a comment to the `mods_load` call describing its use and why it's called when it is.

jakecorrenti:
> > I find the change from `debug` -> `app_output` a little confusing. The output from the application should always be present, and specifying `debug` was specifically to determine whether we should include the dmesg output. Conditionally, reading it as "if we're not in debug mode, do this..." made more sense to me than "if we want app output (which is functionally always true), do this". If I'm in the minority on that I'm not going to block it from going in as is.
> 
> The problem is that debug output needed to be printed earlier, and couldn't be a part of the normal proxy list, as if there's a problem we need to see debug output to find out where. We were unable to do this before because of all other proxies requiring initialization before debug out was viewable.

The logic behind the change makes sense, I was just commenting on the name change being confusing to me.

PR #551, opened 2026-02-17
libkrun, fs: Restrict the usage of REMOVE_ROOT_DIR_REQ ioctl

mtjhrc:
Restrict the usage of VIRTIO_IOC_REMOVE_ROOT_DIR_REQ which deletes the root filesystem. Make it only available on the throwaway filesystem used by krun_set_root_disk_remount.

### DIFF
--- a/src/devices/src/virtio/fs/device.rs
+++ b/src/devices/src/virtio/fs/device.rs
@@ -54,7 +54,12 @@ pub struct Fs {
 }
 
 impl Fs {
-    pub fn new(fs_id: String, shared_dir: String, exit_code: Arc<AtomicI32>) -> super::Result<Fs> {
+    pub fn new(
+        fs_id: String,
+        shared_dir: String,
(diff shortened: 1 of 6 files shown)


slp:
Good one. I think we should go even further and completely get rid of this. Instead, we should allow `Passthrough` to operate without a root directory, just for opening `/init.krun`. Should be fairly simple, we just need to fabricate some bogus `stat` values for `/`.

slp:
On a second thought, this approach is much simpler and fixes the issue just fine. Let's get it merged, thanks!

mtjhrc:
> I think we should go even further and completely get rid of this. Instead, we should allow `Passthrough` to operate without a root directory, just for opening `/init.krun`.

> On a second thought, this approach is much simpler and fixes the issue just fine

Yeah pretty much this, while there is definitely a better way to implement this functionality, this was just a simple improvement for the current status, suitable for 1.7.x.

PR #553, opened 2026-02-18
feat: allow overriding embedded binary paths via env variables

rrauch:
This PR allows overriding the currently hard-coded paths of `INIT_BINARY` (and also `EDK2_BINARY`) at build time.

I've had an issue when using the `devices` crate from outside the `libkrun` tree:

Currently, it's perfectly possible to include the crate in my own project (using the `git` crate location type in my `Cargo.toml`), but the build will fail because the `init` binary included in `virtio::fs::<os>::passthrough` cannot be found.

To fix this, I've replaced

```rust
static INIT_BINARY: &[u8] = include_bytes!("../../../../../../init/init");
```

with

```rust
static INIT_BINARY: &[u8] = include_bytes!(env!("INIT_BINARY_PATH"));
```

Now the `devices` crate can be built even when used out-of-tree. A nice side-effect is that this also allows the use of a custom `init` when desired.

If no `INIT_BINARY_PATH` env variable is set, it falls back to the currently used path, so this PR should remain backwards compatible.

I've noticed that `EDK2_BINARY` (in `vmm`) follows a similar approach, so I have updated that too - even though I am not currently using `aarch64`. I have not tested this myself, but believe it should be fine.

### DIFF
--- a/src/devices/build.rs
+++ b/src/devices/build.rs
@@ -1,0 +1,10 @@
+fn main() {
+    let init_binary_path = std::env::var("KRUN_INIT_BINARY_PATH").unwrap_or_else(|_| {
+        format!(
+            "{}/../../init/init",
+            std::env::var("CARGO_MANIFEST_DIR").unwrap()
+        )
+    });
+    println!("cargo:rustc-env=KRUN_INIT_BINARY_PATH={init_binary_path}");
+    println!("cargo:rerun-if-env-changed=KRUN_INIT_BINARY_PATH");
+}
--- a/src/devices/src/virtio/fs/linux/passthrough.rs
+++ b/src/devices/src/virtio/fs/linux/passthrough.rs
@@ -33,7 +33,7 @@ const EMPTY_CSTR: &[u8] = b"\0";
 const PROC_CSTR: &[u8] = b"/proc/self/fd\0";
 const INIT_CSTR: &[u8] = b"init.krun\0";
 
-static INIT_BINARY: &[u8] = include_bytes!("../../../../../../init/init");
+static INIT_BINARY: &[u8] = include_bytes!(env!("KRUN_INIT_BINARY_PATH"));
 
 type Inode = u64;
 type Handle = u64;
--- a/src/devices/src/virtio/fs/macos/passthrough.rs
+++ b/src/devices/src/virtio/fs/macos/passthrough.rs
@@ -37,7 +37,7 @@ const SECURITY_CAPABILITY: &[u8] = b"security.capability\0";
 
 const UID_MAX: u32 = u32::MAX - 1;
 
-static INIT_BINARY: &[u8] = include_bytes!("../../../../../../init/init");
+static INIT_BINARY: &[u8] = include_bytes!(env!("KRUN_INIT_BINARY_PATH"));
 
 type Inode = u64;
 type Handle = u64;
--- a/src/devices/Cargo.toml
+++ b/src/devices/Cargo.toml
@@ -3,6 +3,7 @@ name = "devices"
 version = "0.1.0"
 authors = ["The Chromium OS Authors"]
 edition = "2021"
+build = "build.rs"
 
 [features]
 tee = []
--- a/src/vmm/Cargo.toml
+++ b/src/vmm/Cargo.toml
@@ -3,6 +3,7 @@ name = "vmm"
 version = "0.1.0"
 authors = ["Amazon Firecracker team <firecracker-devel@amazon.com>"]
 edition = "2021"
+build = "build.rs"
 
 [features]
 tee = []
--- a/src/vmm/build.rs
+++ b/src/vmm/build.rs
@@ -1,0 +1,13 @@
+fn main() {
+    #[cfg(target_arch = "aarch64")]
+    {
+        let edk2_binary_path = std::env::var("KRUN_EDK2_BINARY_PATH").unwrap_or_else(|_| {
+            format!(
+                "{}/../../edk2/KRUN_EFI.silent.fd",
+                std::env::var("CARGO_MANIFEST_DIR").unwrap()
+            )
+        });
+        println!("cargo:rustc-env=KRUN_EDK2_BINARY_PATH={edk2_binary_path}");
+        println!("cargo:rerun-if-env-changed=KRUN_EDK2_BINARY_PATH");
+    }
+}
--- a/src/vmm/src/builder.rs
+++ b/src/vmm/src/builder.rs
@@ -101,7 +101,7 @@ use vm_memory::{GuestAddress, GuestMemoryMmap};
 
 #[cfg(target_arch = "aarch64")]
 #[allow(dead_code)]
-static EDK2_BINARY: &[u8] = include_bytes!("../../../edk2/KRUN_EFI.silent.fd");
+static EDK2_BINARY: &[u8] = include_bytes!(env!("KRUN_EDK2_BINARY_PATH"));
 
 /// Errors associated with starting the instance.
 #[derive(Debug)]


src/devices/src/virtio/fs/linux/passthrough.rs
```diff
@@ -33,7 +33,7 @@ const EMPTY_CSTR: &[u8] = b"\0";
 const PROC_CSTR: &[u8] = b"/proc/self/fd\0";
 const INIT_CSTR: &[u8] = b"init.krun\0";
 
-static INIT_BINARY: &[u8] = include_bytes!("../../../../../../init/init");
+static INIT_BINARY: &[u8] = include_bytes!(env!("INIT_BINARY_PATH"));
```
mtjhrc: I would prefix the environment variable names with `KRUN_`, it make sure we don't accidentally pick up something from the environment we didn't want.

src/devices/build.rs
```diff
@@ -0,0 +1,10 @@
+fn main() {
+    let init_binary_path = std::env::var("INIT_BINARY_PATH").unwrap_or_else(|_| {
+        format!(
+            "{}/../../init/init",
+            std::env::var("CARGO_MANIFEST_DIR").unwrap()
+        )
+    });
```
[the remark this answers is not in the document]
mtjhrc: There is no point to handling this IMO

src/devices/src/virtio/fs/linux/passthrough.rs
rrauch: Makes sense, I'll prefix the env variable names with `KRUN_`.

src/devices/build.rs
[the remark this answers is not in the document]
rrauch: Yes, the `expect` here is just LLM noise

src/devices/build.rs
```diff
@@ -0,0 +1,10 @@
...
+    println!("cargo:rustc-env=INIT_BINARY_PATH={init_binary_path}");
+    println!("cargo:rerun-if-changed={init_binary_path}");
```
mz-pdm: Shouldn't it be ``println!("cargo:rerun-if-env-changed=INIT_BINARY_PATH")``?

rrauch: According to https://doc.rust-lang.org/cargo/reference/build-scripts.html#rerun-if-env-changed:
 > As of 1.46, using [env!](https://doc.rust-lang.org/std/macro.env.html) and [option_env!](https://doc.rust-lang.org/std/macro.option_env.html) in source code will automatically detect changes and trigger rebuilds. rerun-if-env-changed is no longer needed for variables already referenced by these macros.
 
So that case is already covered by the `env!` macro.

`println!("cargo:rerun-if-changed={init_binary_path}")` handles the case where the `init` binary itself changed (e.g. recompiled).

mtjhrc: Actually we should have both. In case the value of the env variable had changed, and in case someone updated the binary file of init.

rrauch: Yes, I agree. And I think both cases are already covered. See my previous comment.

mtjhrc: Actually, It seems like `include_bytes!` seem to already handle that too (but I can't find it documented now)

src/devices/src/virtio/fs/macos/passthrough.rs
```diff
@@ -36,7 +36,7 @@ const SECURITY_CAPABILITY: &[u8] = b"security.capability\0";
 
 const UID_MAX: u32 = u32::MAX - 1;
 
-static INIT_BINARY: &[u8] = include_bytes!("../../../../../../init/init");
+static INIT_BINARY: &[u8] = include_bytes!(env!("INIT_BINARY_PATH"));
```
mtjhrc: So if `build.rs` is not necessary for invalidation, I would just put the code here and get rid of the `build.rs`

```suggestion
static INIT_BINARY: &[u8] = include_bytes!(match option_env!("KRUN_INIT_BINARY_PATH") {
    Some(path) => path,
    None => "../../../../../../init/init"
});
```

rrauch: That won't work, unfortunately. `include_bytes!` only accepts string literals, not arbitrary expressions:

```
error: argument must be a string literal
  --> src/devices/src/virtio/fs/linux/passthrough.rs:36:44
   |
36 |   static INIT_BINARY: &[u8] = include_bytes!(match option_env!("KRUN_INIT_BINARY_PATH") {
   |  ____________________________________________^
37 | |     Some(path) => path,
38 | |     None => "../../../../../../init/init"
39 | | });
   | |_^
```
I don't think we can get rid of the `build.rs`.

src/devices/build.rs
rrauch: I can't find anything in the docs either. I found https://github.com/rust-lang/rust/pull/24423 which states that the included file is tracked, but in the commit itself it says only the file is tracked and not its content - whatever that means in this context.

I would suggest to keep the `cargo:rerun-if-changed` - I can't see any harm here and it makes it explicit.
However, if you'd prefer to have it removed, I don't mind either.

src/devices/src/virtio/fs/macos/passthrough.rs
mtjhrc: Oh, I didn't realize that.

src/devices/build.rs
mz-pdm: If I'm not mistaken, the ``rerun-if-*`` stuff specifies when the *build script* should be rebuilt. It's necessary to include the check for the environment variable (there is no ``env!`` in the script) so that the changed path is noticed. `rerun-if-changed` is superfluous because the contents of the file has no influence on the build script itself; having it included is harmless (at worst the build script is rebuilt unnecessary) but we should understand what's happening exactly.

ggoodman:
I had experimented with requiring the embedding project to supply the init binary as a memory range. This was quite nice as in my use case the coupling was between the embedding program and the init. So in that design, libkrun was only responsible for the hand-off from the host to the guest init instead of also owning init. I think the coupling and complexity of the built-in init suffers because it has to handle every use-case.

mtjhrc:
> I had experimented with requiring the embedding project to supply the init binary as a memory range. This was quite nice as in my use case the coupling was between the embedding program and the init. So in that design, libkrun was only responsible for the hand-off from the host to the guest init instead of also owning init.

We will likely end up doing something like this for libkrun 2.0.
IMO we should eventually have an API for specifying any path and a memory range to inject into the fs. Then we could ship the init as some kind of extension for container runtime (or maybe as part of libkrunfw?).

src/devices/build.rs
rrauch: @mz-pdm Your comment is spot on! `rerun-if-*` is really *only* about when to re-run the build script, nothing else.
I believe your original suggestion is correct then; it should indeed be `cargo:rerun-if-env-changed`. I stand corrected.

I've tried it & it works reliably. Also the scenario where `KRUN_INIT_BINARY_PATH` was set to some value previously but has since become unset works correctly. I believe it might have been a bit "flaky" before.

I'll change it to `cargo:rerun-if-env-changed` then.

rrauch:
@ggoodman I am also using my own `init` binary currently. I am actually supplying it via a `initrd` as I am not too fond of the "virtual init-inode" approach libkrun is currently taking to inject the init binary via virtio-fs.

I had to make some small changes to `vmm` to allow the general `initrd` use as it is currently limited to the `tee` feature. I wasn't sure at first if it was going to work because of this limit, but in the end it was only a couple of lines and it seems to work just fine.

However, the `devices` crate still insists on an `init` binary. Hence this PR. In my use-case I am just feeding it a dummy binary to satisfy the compiler.

mz-pdm:
The change itself looks fine to me now. Should the commits be squashed?

mtjhrc:
> The change itself looks fine to me now. Should the commits be squashed?

Yes, please squash the commits. 

Also note for the commit message title we tend to use the form like `subject(s): the actual change text` (NOT the so called "conventional commits" e.g. "feat: "). The `subject` is typically a specific device, subsystem or something like "Makefile: " (we don't have a strict or specific guideline, just look at the existing commit messages). 

e.g. :
1. `devices/fs: allow overriding embedded init path via env variable`
2. `vmm: allow overriding embedded EDK2 firmware path via env variable`

(or you can put it in 1 commit, doesn't really matter here, the change is small...)

rrauch:
I've squashed the commits and changed the commit message to a more descriptive one that includes the crates/subsystems affected.

slp:
@mtjhrc Since you've requested changes, PTAL.

PR #248, opened 2025-01-12
Add pipewire cross domain channel type.

WhatAmISupposedToPutHere:
This extends the already existing wayland/x11 cross domain channel to also work for sharing pipewire between host and guest. The necessary changes are eventfd support (inspired by how wayland pipes are handled), a few fixes to sharing shm buffers, and a new protocol version that supports more fds per message (pw can sometimes send more than 4).

PR #558, opened 2026-02-21
rutabaga: do not rely on seals to detect read-only shm fds

valpackett:
Some compositors send read-only but not sealed memfds (at least wlroots does so for dmabuf feedback), so seal-based permission detection would result in mapping RO memory as RW and the guest application crashing.  Fix by properly checking the access mode. Keep seal detection because e.g. Smithay based compositors do send O_RDWR but write-sealed memfds.

[Upstream rutabaga does not do any detection yet](https://github.com/magma-gpu/rutabaga_gfx/blob/a7d5a891cac54b1d8f4eadc29ecec1b19b9e3ba8/src/cross_domain/mod.rs#L879-L886), but defaults to RO for all shm/memfd (!!) and RW for dmabuf.

---

~~I completely removed the seal part because it seems redundant (since seals are an extra facility that allows to make the RO-ness irreversible, so you shouldn't be able to have an RO seal on an RW descriptor) *and* upstream defaults to RO, but I wonder what the lore is behind checking for the seal and not for the flag…~~

I ~~will propose~~ [proposed](https://github.com/magma-gpu/rutabaga_gfx/pull/42) the checking for upstream rutabaga now too btw.

P.S. Please don't LLM-review this.

### DIFF
--- a/src/rutabaga_gfx/src/cross_domain/mod.rs
+++ b/src/rutabaga_gfx/src/cross_domain/mod.rs
@@ -1122,15 +1122,27 @@ impl RutabagaContext for CrossDomainContext {
             .remove(&item_id)
             .ok_or(RutabagaError::InvalidCrossDomainItemId)?;
 
+        fn access_mode(descriptor: &SafeDescriptor) -> RutabagaResult<u32> {
+            Ok(
+                match fcntl(descriptor.as_fd(), FcntlArg::F_GETFL)? & libc::O_ACCMODE {
+                    libc::O_RDWR => RUTABAGA_MAP_ACCESS_RW,
+                    libc::O_WRONLY => RUTABAGA_MAP_ACCESS_WRITE,
+                    _ => RUTABAGA_MAP_ACCESS_READ,
+                    // (there is a "secret fourth option" O_WRONLY|O_RDWR meaning "no access", very unlikely we'd see it)
+                },
+            )
+        }
+
         // Items that are removed from the table after one usage.
         match item {
             CrossDomainItem::ShmBlob(descriptor) => {
                 #[allow(unused_mut)]
-                let mut access = RUTABAGA_MAP_ACCESS_RW;
+                let mut access = access_mode(&descriptor)?;
+                // Some compositors actually do send descriptors that are O_RDWR but write-sealed (!)
                 #[cfg(target_os = "linux")]
                 if fcntl(&descriptor, FcntlArg::F_GET_SEALS)? & libc::F_SEAL_WRITE != 0 {
                     access &= !RUTABAGA_MAP_ACCESS_WRITE;
-                };
+                }
 
                 let hnd = RutabagaHandle {
                     os_handle: descriptor,
@@ -1156,10 +1168,7 @@ impl RutabagaContext for CrossDomainContext {
                 })
             }
             CrossDomainItem::DmaBuf(descriptor) => {
-                let mut access = RUTABAGA_MAP_ACCESS_READ;
-                if fcntl(descriptor.as_fd(), FcntlArg::F_GETFL)? & libc::O_WRONLY != 0 {
-                    access |= RUTABAGA_MAP_ACCESS_WRITE;
-                }
+                let access = access_mode(&descriptor)?;
 
                 let hnd = RutabagaHandle {
                     os_handle: descriptor,


valpackett:
Whoops… the "correct" check makes it always crash, the "lore" is that I'm the one who added the seal check :woman_facepalming: aaaand before that, the `READ` was changed to unconditional `RW` in the PR that implemented PipeWire support.

But it's not included in the PR currently upstreaming PipeWire support to rutabaga, so looks like that change was bogus all along (??)

---

UPD: now the *actually* correct check seems to work in all cases :rocket:

alyssais:
Hmm, I fixed something that was at least very similar to this upstream in 2023. https://chromium.googlesource.com/chromiumos/platform/crosvm/+/998597a1bd29432bdee28d298511549edff1434a%5E%21/

valpackett:
@alyssais oh hi :) So it looks like there you actually wrote the "downstream" (from the decision-making point) code in the `map_resource` handlers that does `match map_info & RUTABAGA_MAP_ACCESS_MASK {…}`…

And in the same commit you hardcoded the `map_info` decisions per type:

```diff
--- a/rutabaga_gfx/src/cross_domain/mod.rs
+++ b/rutabaga_gfx/src/cross_domain/mod.rs

@@ -716,7 +716,7 @@
                         blob: true,
                         blob_mem: resource_create_blob.blob_mem,
                         blob_flags: resource_create_blob.blob_flags,
-                        map_info: Some(reqs.map_info),
+                        map_info: Some(reqs.map_info | RUTABAGA_MAP_ACCESS_RW),
                         info_2d: None,
                         info_3d: Some(info_3d),
                         vulkan_info: reqs.vulkan_info,
@@ -750,7 +750,7 @@
                         blob: true,
                         blob_mem: resource_create_blob.blob_mem,
                         blob_flags: resource_create_blob.blob_flags,
-                        map_info: Some(RUTABAGA_MAP_CACHE_CACHED),
+                        map_info: Some(RUTABAGA_MAP_CACHE_CACHED | RUTABAGA_MAP_ACCESS_READ),
                         info_2d: None,
                         info_3d: None,
                         vulkan_info: None,
```

So that's where the "SHM is READ, DMABUF is RW" logic comes from! (BTW I was somewhat surprised at first that "woah, host SHM is READ?? but we write to host SHM constantly" but right— this is only for incoming FDs sent by the compositor, while the ones we write to are allocated on demand when the proxy requests them.)

It is indeed "good enough" for just Wayland, but here we started to transport other protocols over cross-domain, like PipeWire and X11 (and now I'm doing D-Bus).

As I mentioned in https://github.com/magma-gpu/rutabaga_gfx/pull/42 **the PR that added PipeWire here (#248) was what changed the hardcoded `READ` to hardcoded `RW` for these SHM descriptors** and added a check for RDONLY dmabufs.

That's where all the trouble began, first with sealed-but-RDWR fds (so I've added the seal check) and now with RDONLY-but-not-sealed ones. yeah, now it's time to make the decision properly based on what was sent to us.

dorindabassey:
Hi @valpackett, thanks for the fix. I see you've already opened https://github.com/magma-gpu/rutabaga_gfx/pull/42 for this upstream, Since we're actively working to remove the local rutabaga fork in favor of the upstream version from https://github.com/magma-gpu/rutabaga_gfx/pull/36 , could we hold off on merging changes to the local copy? Once your upstream PR lands and and a new rutabaga release is cut, we can pick up this fix when we switch to the upstream version. I just think it's better to avoid further divergence from upstream if possible. Wdyt?

btw we have a new release cut for libkrun: https://github.com/magma-gpu/rutabaga_gfx/pull/36#issuecomment-3923352509

slp:
> Hi @valpackett, thanks for the fix. I see you've already opened [magma-gpu/rutabaga_gfx#42](https://github.com/magma-gpu/rutabaga_gfx/pull/42) for this upstream, Since we're actively working to remove the local rutabaga fork in favor of the upstream version from [magma-gpu/rutabaga_gfx#36](https://github.com/magma-gpu/rutabaga_gfx/pull/36) , could we hold off on merging changes to the local copy? Once your upstream PR lands and and a new rutabaga release is cut, we can pick up this fix when we switch to the upstream version. I just think it's better to avoid further divergence from upstream if possible. Wdyt?

I disagree. While planned, we still don't know exactly when we'll be able to switch to upstream rutabaga_gfx. Quite possibly it won't happen until 2.0, as it's a *very* risky change which can have a huge impact on both krunkit (used by Podman Desktop) and muvm. This means we need to maintain our vendored rutabaga for the remaining 1.x lifetime (which can easily expand to many months after the 2.0 release).

Fixes and improvements like the one in this PR are still very appreciated.

dorindabassey:
> I disagree. While planned, we still don't know exactly when we'll be able to switch to upstream rutabaga_gfx. Quite possibly it won't happen until 2.0, as it's a _very_ risky change which can have a huge impact on both krunkit (used by Podman Desktop) and muvm. This means we need to maintain our vendored rutabaga for the remaining 1.x lifetime (which can easily expand to many months after the 2.0 release).
> 
> Fixes and improvements like the one in this PR are still very appreciated.

@slp I've opened #560 as a draft PR to test the upstream rutabaga migration for 2.0. with this it can help us figure out the timeline and any issues. For 1.x, understood that we'll continue maintaining the vendored copy.

valpackett:
Rebased. Should be ready to go.

slp (approved):
LGTM, thanks!

PR #559, opened 2026-02-23
misc: configure gemini for manual reviews

slp:
Configure Gemini Code Assist to only act on a PR when explicitly requested (i.e. with "/gemini review").


slp:
cc/ @valpackett

PR #560, opened 2026-02-23
Migrate to rutabaga_gfx v0.1.85 from crates.io

dorindabassey:
As part of the efforts to remove vendored rutabaga_gfx and use upstream Rutabaga_gfx, we are adopting the new RutabagaBuilder API with capset_mask, Add temporary workaround for VirtioFsKey not being publicly exported(marking as draft pending https://github.com/magma-gpu/rutabaga_gfx/pull/43 merge).


mtjhrc:
To continue on the discussion from https://github.com/containers/libkrun/pull/558

Personally I also think we need to continue to accept important changes to rutabaga until we use the upstream. Actually this will help us close the gap (but the things we merge should also go upstream!). 
Also, I think it's entirely possible to switch to upstream on libkrun 1.x too, but we need feature parity, packaging and test that everything works. If that is something  we want to work on depends on the timeline of libkrun 2.0 (we can also discuss this on the next libkrun sync).

src/devices/src/virtio/gpu/virtio_gpu.rs
```diff
@@ -282,24 +290,23 @@ impl VirtioGpu {
...
+            .build()
+            .ok()
     }
 
     #[allow(clippy::too_many_arguments)]
     pub fn new(
         mem: GuestMemoryMmap,
         queue_ctl: Arc<Mutex<VirtQueue>>,
         interrupt: InterruptTransport,
-        virgl_flags: u32,
```
valpackett, line 307: We cannot do this…

We can't just "allow everything" because some flags are mutually exclusive. Venus only supports out-of-process mode, vDRM only supports in-process mode (at least as of the last time I looked into all that). And legacy GL can ruin things for $reasons.

This is how muvm currently picks virgl_flags:

```rust
        let virgl_mode = match options.gpu_mode.unwrap_or_default() {
            muvm::cli_options::GpuMode::Drm => VIRGLRENDERER_DRM,
            muvm::cli_options::GpuMode::Venus => VIRGLRENDERER_VENUS | VIRGLRENDERER_RENDER_SERVER,
            muvm::cli_options::GpuMode::Software => 0,
        };
        let virgl_flags = VIRGLRENDERER_USE_EGL
            | VIRGLRENDERER_NO_VIRGL /* Legacy method that we don't support; interferes with software-only mode */
            | virgl_mode
            | VIRGLRENDERER_THREAD_SYNC
            | VIRGLRENDERER_USE_ASYNC_FENCE_CB;
        // SAFETY: Safe as no pointers involved.
        let err = unsafe {
            krun_set_gpu_options2(
                ctx_id,
                virgl_flags,
                (u32::from(vram_shm_mib) as u64) * 1024 * 1024,
            )
        };
```

We have to be respect the flags (at least by translating into capsets and builder calls) until we provide new FFI API with a cleaner way to communicate these decisions.

valpackett: Wow, THE strangest thing I found while debugging from here: #565

valpackett:
Finally testing this since I needed to test https://github.com/magma-gpu/rutabaga_gfx/pull/42 and also maybe avoid contributing D-Bus support into the fork when we're so close to unforking… :)

Other than the flags/capset control issue above, one thing missing compared to before is gbm (referenced in the discussion in https://github.com/magma-gpu/rutabaga_gfx/issues/40), I ported over the upstream gbm support hacks from the vendored fork but might've screwed something up since software-rendered apps crash the proxy with that enabled (investigating). But without the gbm feature they work fine. (FWIW this gbm gralloc thing seems to only be an optimization, to make sure the SHM copy-out in the guest proxy goes directly to GPU-managed memory)

src/devices/src/virtio/gpu/virtio_gpu.rs
dorindabassey: Fixed, I've put the flag back. Thanks for the reference.

dorindabassey:
> Finally testing this since I needed to test [magma-gpu/rutabaga_gfx#42](https://github.com/magma-gpu/rutabaga_gfx/pull/42) and also maybe avoid contributing D-Bus support into the fork when we're so close to unforking… :)

+1
 
> Other than the flags/capset control issue above, one thing missing compared to before is gbm (referenced in the discussion in [magma-gpu/rutabaga_gfx#40](https://github.com/magma-gpu/rutabaga_gfx/issues/40)), I ported over the upstream gbm support hacks from the vendored fork but might've screwed something up since software-rendered apps crash the proxy with that enabled (investigating). But without the gbm feature they work fine. (FWIW this gbm gralloc thing seems to only be an optimization, to make sure the SHM copy-out in the guest proxy goes directly to GPU-managed memory)

Fixed the flags/capset issue, For gbm - I see from magma-gpu/rutabaga_gfx#40 that there's a plan to move from minigbm to upstream GBM. Since the gbm feature isn't currently enabled in our Cargo.toml and you said it crashes with software rendering, maybe let's land this PR first and add gbm support as a follow-up once the upstream rutabaga_gfx gbm migration (magma-gpu/rutabaga_gfx#40) is done? I don't know the timeline for that, I just want to make sure we unblock the migration without getting stuck on an optimization feature. wdyt?
Just to clarify, after taking a second look at the vendored rutabaga_gfx, even though we had minigbm feature defined it was never actually enabled in our Cargo.toml (we only used `virgl_renderer`). So gbm wasn't being used before either.

valpackett:
> maybe let's land this PR first and add gbm support as a follow-up once the upstream rutabaga_gfx gbm migration is done?

Definitely +1

> even though we had minigbm feature defined it was never actually enabled in our Cargo.toml (we only used virgl_renderer). So gbm wasn't being used before either.

Ohhh… huh. I was so sure it was being used, but that seems to be true!

> Fixed the flags/capset issue

Thanks a lot, will take a look now

src/devices/src/virtio/gpu/virtio_gpu.rs
```diff
@@ -252,28 +279,61 @@ impl VirtioGpu {
...
+        // TODO: Consider making these public in rutabaga_gfx to avoid duplication.
+        const VIRGLRENDERER_USE_EGL: u32 = 1 << 0;
+        const VIRGLRENDERER_USE_GLES: u32 = 1 << 4;
+        const VIRGLRENDERER_VENUS: u32 = 1 << 6;
+        const VIRGLRENDERER_NO_VIRGL: u32 = 1 << 7;
+        const VIRGLRENDERER_RENDER_SERVER: u32 = 1 << 9;
+        const VIRGLRENDERER_DRM: u32 = 1 << 10;
+
+        let mut capset_mask: u64 = 0;
+
```
valpackett, line 309:
```suggestion

        capset_mask |= 1 << rutabaga_gfx::RUTABAGA_CAPSET_CROSS_DOMAIN;

```

(and inline `rutabaga_paths_opt`)

Only this + #565 remains \o/

dorindabassey: fixed, Thanks!

valpackett:
Great news: the entire libkrun branch of rutabaga has been [applied directly to main](https://github.com/magma-gpu/rutabaga_gfx/pull/55)!

BTW, my current dev branch https://github.com/valpackett/libkrun/commits/test/ is basically this PR rebased and switched to use my branch of rutabaga which.. is just upstream rutabaga main right now, since everything has landed there \o/

slp:
We're still missing the bits to support virglrenderer on macOS, but we'll be waiting a bit to see how upstream support evolves first, since we still have some time before 2.0.

dorindabassey:
Bumping this in your inbox - I have updated also the CI to use latest virglrenderer 1.3.0 since that's what rutabaga uses aswell. Unfortunately the macOS CI fails because the Homebrew tap ships virglrenderer 0.10.4e, but upstream rutabaga_gfx requires 1.3.0. @slp I suggest we update the tap  to upstream 1.3.0, or rebase your fork on 1.3.0? so that macOS GPU clippy passes, Is that going to be a problem for MacOs build?

slp:
> @slp I suggest we update the tap to upstream 1.3.0, or rebase your fork on 1.3.0? so that macOS GPU clippy passes, Is that going to be a problem for MacOs build?

I have a branch here updating our fork to 1.3.0. I'll try to update the tap later this month.

dorindabassey:
Can you share the branch? I can update the tap for you - I already have the formula changes ready, just need to point it at your rebased fork instead

slp:
> Can you share the branch? I can update the tap for you - I already have the formula changes ready, just need to point it at your rebased fork instead

We need to test it very carefully and do a coordinated upgrade with libkrun. We have people consuming directly the tap for uses that require GPU acceleration.

mtjhrc:
> We need to test it very carefully and do a coordinated upgrade with libkrun. We have people consuming directly the tap for uses that require GPU acceleration.

Sure, but that's for stable libkrun 1.x. In 2.x we could upgrade earlier and just point the CI to the current fork no?

slp:
> > We need to test it very carefully and do a coordinated upgrade with libkrun. We have people consuming directly the tap for uses that require GPU acceleration.
> 
> Sure, but that's for stable libkrun 1.x. In 2.x we could upgrade earlier and just point the CI to the current fork no?

You mean having a new tap for 2.x? Yes, we could do that.

mtjhrc:
> You mean having a new tap for 2.x? Yes, we could do that.

Yes that would work. But I meant just `make install` your virglrenderer 1.3.0 fork in the CI for now is the easiest solution currently.

dorindabassey:
> Yes that would work. But I meant just `make install` your virglrenderer 1.3.0 fork in the CI for now is the easiest solution currently.

Right, Since this is just for clippy - it only needs to compile, not run GPU.

dorindabassey:
Hi @slp  bumping this to the top of your inbox: I have rebased and updated the PR to use the latest rutabaga_gfx from crates.io and also updated the CI for MacOS build to use latest virglrenderer 1.3.0 since that's what rutabaga uses aswell.
cc @mtjhrc

slp:
This breaks GPU acceleration on macOS, but let's get it merged in 2.0-devel and we'll fix it with a new virglrenderer before the first 2.0 release.

slp (approved):

PR #563, opened 2026-02-26
vmm: prevent CR translation to LF in interactive term

### LINKED ISSUE #562: Enter (CR) translated to LF in interactive TTY sessions
`term_set_raw_mode` currently clears only local flags and  leaves ICRNL unchanged. On Linux this means that CR->LF translation is enabled  for the guest in raw mode unless `stty -ICRNL` is explicitly run on the host terminal.

This becomes a problem using [krun](https://github.com/containers/crun) with Podman since my understanding is that a fresh PTY is allocated with ICRNL enabled, regardless of the user terminal settings. As a result, any libkrun backed container receives LF on `<Enter>` in raw mode, which breaks CLI apps like fzf and codex.

Minimal working example using `chroot_vm`:

```sh
$ ./chroot_vm ./rootfs_fedora /bin/sh
sh-5.3# stty -F /dev/tty raw -echo; dd if=/dev/tty bs=1 count=1 2>/dev/null | od -An -t x1
# Press Enter
# Returns 0a unless host-side ICRNL disabled
```

This behaviour is probably what's responsible for the following bug reports with Podman and krunvm:

containers/podman/issues/28067

containers/krunvm/issues/81

jakecorrenti:
Fix a bug where Carriage Return (CR) get's translated into a Line Feed (LF) in interactive terminal use-cases.

Fixes #562

### DIFF
--- a/src/vmm/src/terminal.rs
+++ b/src/vmm/src/terminal.rs
@@ -1,4 +1,4 @@
-use nix::sys::termios::{tcgetattr, tcsetattr, LocalFlags, SetArg, Termios};
+use nix::sys::termios::{cfmakeraw, tcgetattr, tcsetattr, LocalFlags, SetArg, Termios};
 use std::os::fd::BorrowedFd;
 
 #[must_use]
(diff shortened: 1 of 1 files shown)


mtjhrc:
Probably also fixes this old bug https://github.com/containers/krunvm/issues/39

slp (approved):
LGTM, thanks!

slp:
Also thanks to @jsks for the detailed bug report!

PR #564, opened 2026-02-26
Cleanup macOS epoll shim

nirs:
Clean up the `wait()` function in the macOS epoll/kevent shim to make it
easier to read and debug.

## Changes

1. **Return early on errors** -- handle kevent() errors upfront instead
   of matching at the end, making the flow clearer.

2. **Cast kevent count to usize before the loop** -- eliminates repeated
   `i as usize` casts on every array access.

3. **Improve kevent debug log** -- log processed event info (READ/WRITE,
   fd, data) instead of the raw kevent struct, making debug output
   immediately useful.

## Example logs

```
[2026-02-26T18:28:23.856913Z DEBUG utils::macos::epoll] kevent: WRITE fd=3 data=66560
[2026-02-26T18:28:23.856923Z DEBUG utils::macos::epoll] kevent: READ fd=3 data=3539700
[2026-02-26T18:28:23.859383Z DEBUG utils::macos::epoll] kevent: READ fd=83 data=8
```

Includes the timestamp commit from #556 for better logging.

### DIFF
--- a/src/utils/src/macos/epoll.rs
+++ b/src/utils/src/macos/epoll.rs
@@ -12,6 +12,18 @@ use std::time::Duration;
 use bitflags::bitflags;
 use log::debug;
 
+fn event_name(event: u32) -> &'static str {
+    match event {
+        e if e == EventSet::IN.bits() => "READ",
+        e if e == EventSet::OUT.bits() => "WRITE",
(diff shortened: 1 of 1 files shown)


mtjhrc (approved):
LGTM, thanks!

PR #588, opened 2026-03-14
libkrun: Produce a proper Rust library

d-e-s-o:
Produce a proper Rust library to avoid the need for C library
installation as well as binding generation/definition. Basically, we can
now reference the crate via `Cargo.toml`:
```toml
[dependencies]
libkrun = ...
```

instead of having to rely on constructs like:
```rust
#[link(name = "krun")]
extern "C" {
  pub fn krun_create_ctx() -> i32;
  ...
}
```

For a brief discussion of the topic, see https://github.com/containers/libkrun/pull/494#issuecomment-4017885314

### DIFF
--- a/src/devices/src/virtio/console/process_rx.rs
+++ b/src/devices/src/virtio/console/process_rx.rs
@@ -95,6 +95,9 @@ fn read_to_desc(
     input: &mut (dyn PortInput + Send),
     eof: &mut bool,
 ) -> Result<usize, GuestMemoryError> {
+    // TODO: Switch to using `get_slices()` with the next vm-memory
+    //       bump.
+    #[allow(deprecated)]
     desc.mem
         .try_access(desc.len as usize, desc.addr, |_, len, addr, region| {
             let mut target = region.get_slice(addr, len).unwrap();
--- a/src/devices/src/virtio/console/process_tx.rs
+++ b/src/devices/src/virtio/console/process_tx.rs
@@ -83,6 +83,9 @@ fn write_desc_to_output(
     output: &mut (dyn PortOutput + Send),
     interrupt: &InterruptTransport,
 ) -> Result<usize, GuestMemoryError> {
+    // TODO: Switch to using `get_slices()` with the next vm-memory
+    //       bump.
+    #[allow(deprecated)]
     desc.mem
         .try_access(desc.len as usize, desc.addr, |_, len, addr, region| {
             let src = region.get_slice(addr, len).unwrap();
--- a/src/vmm/Cargo.toml
+++ b/src/vmm/Cargo.toml
@@ -21,11 +21,11 @@ aws-nitro = []
 crossbeam-channel = ">=0.5.15"
 flate2 = "1.0.35"
 libc = ">=0.2.39"
-linux-loader = { version = "0.13.0", features = ["bzimage", "elf", "pe"] }
+linux-loader = { version = "0.13.2", features = ["bzimage", "elf", "pe"] }
 log = "0.4.0"
 nix = { version = "0.30.1", features = ["fs", "term"] }
-vm-memory = { version = ">=0.13", features = ["backend-mmap"] }
-vmm-sys-util = ">=0.14"
+vm-memory = { version = "0.17.0", features = ["backend-mmap"] }
+vmm-sys-util = "0.14"
 krun_display = { path = "../krun_display", optional = true, features = ["bindgen_clang_runtime"] }
 krun_input = { path = "../krun_input", optional = true, features = ["bindgen_clang_runtime"] }
 
@@ -51,8 +51,8 @@ zstd = "0.13"
 
 [target.'cfg(target_os = "linux")'.dependencies]
 tdx = { version = "0.1.0", optional = true }
-kvm-bindings = { version = ">=0.11", features = ["fam-wrappers"] }
-kvm-ioctls = ">=0.21"
+kvm-bindings = { version = "0.12", features = ["fam-wrappers"] }
+kvm-ioctls = "0.22"
 
 [target.'cfg(target_os = "macos")'.dependencies]
 hvf = { path = "../hvf" }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,25 +3,16 @@
 version = 4
 
 [[package]]
-name = "addr2line"
-version = "0.24.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "dfbe277e56a376000877090da837660b4427aad530e3028d44e0bffe4f89a1c1"
-dependencies = [
- "gimli",
-]
-
-[[package]]
 name = "adler2"
-version = "2.0.0"
+version = "2.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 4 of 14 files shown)


d-e-s-o:
Still blocked on https://github.com/rust-vmm/linux-loader/pull/223 or https://github.com/rust-vmm/linux-loader/pull/229 and it being included in a new release.

ggoodman:
I had my 🤖 help me put together a PR describing something I hope we can land before introducing the rust api: #593.

The main intent is that we not couple the init into the crate and instead have it as an optional sibling crate.

d-e-s-o:
> Still blocked on [rust-vmm/linux-loader#223](https://github.com/rust-vmm/linux-loader/pull/223) or [rust-vmm/linux-loader#229](https://github.com/rust-vmm/linux-loader/pull/229) and it being included in a new release.

Doesn't seem as if one of those is going to get merged anytime soon. So we have to stick with vm-memory 0.17. Updated PR accordingly.

d-e-s-o:
Could somebody help review the change please?

src/devices/src/virtio/console/process_rx.rs
```diff
@@ -89,19 +89,18 @@ fn read_to_desc(
...
+    let mut total = 0;
+    for slice in desc.mem.get_slices(desc.addr, desc.len as usize) {
+        let mut slice = slice?;
+        match input.read_volatile(&mut slice) {
+            Ok(0) => {
+                *eof = true;
+                break;
             }
-        })
+            Ok(n) => total += n,
```
dorindabassey: Here, going from `try_access()` to `get_slices()` changes the read behavior when guest memory is fragmented. The previous code called `read()` once, and the new code may call it multiple times (once per slice) until `WouldBlock`. This is probably fine, but worth noting since it's not mentioned in the PR description and should probably be tested with fragmented guest memory.

src/vmm/Cargo.toml
```diff
@@ -21,11 +21,11 @@ aws-nitro = []
 crossbeam-channel = ">=0.5.15"
 flate2 = "1.0.35"
 libc = ">=0.2.39"
-linux-loader = { version = "0.13.0", features = ["bzimage", "elf", "pe"] }
+linux-loader = { version = "0.13.2", features = ["bzimage", "elf", "pe"] }
 log = "0.4.0"
 nix = { version = "0.30.1", features = ["fs", "term"] }
-vm-memory = { version = ">=0.13", features = ["backend-mmap"] }
-vmm-sys-util = ">=0.14"
+vm-memory = { version = ">=0.17.0, <0.17.2", features = ["backend-mmap"] }
```
dorindabassey: Just curious, why exclude 0.17.2+? Everywhere else just uses "0.17", or should they be consistent?

dorindabassey (commented):
code LGTM, just a few comments.

src/devices/src/virtio/console/process_rx.rs
mtjhrc: @dorindabassey  It should be the same in this regard actually, the old code would call it multiple times too - `try_access` invoked the closure for each slice. 

Not sure if there is actually a functional difference here, so far haven't noticed any, but this did look suspicious it might break something.

dorindabassey: Ah thanks for clarifying! But separately, looks like there might be an issue with partial reads within a single slice as per gemini.

src/vmm/Cargo.toml
d-e-s-o: I have recollection that it was necessary, because otherwise some transitive dependency would get bumped in a problematic way...but perhaps that's no longer an issue due to other constraints -- I can't seem to reproduce anymore. Changed.

src/devices/src/virtio/console/process_rx.rs
d-e-s-o: Perhaps it's best if we leave the code as it was for now? I only adjusted it reluctantly because `try_access()` has been deprecated. But it hasn't been removed yet and as per semver that can only happen with the next minor bump. So I've just `#[allow]`ed the deprecation now. Let me know what you think.

src/devices/src/virtio/console/process_tx.rs
```diff
@@ -76,22 +76,24 @@ fn write_desc_to_output(
...
+                    interrupt.signal_used_queue();
+                    output.wait_until_writable();
                 }
+                Err(e) => return Err(GuestMemoryError::IOError(e)),
             }
-        })
+        }
+    }
+    Ok(total)
 }
```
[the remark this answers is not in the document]
d-e-s-o: I reverted to the original as per https://github.com/containers/libkrun/pull/588#discussion_r3009762779 and just allow deprecated method usage.

src/devices/src/virtio/console/process_rx.rs
```diff
@@ -89,19 +89,18 @@ fn read_to_desc(
...
+            Err(e) if e.kind() == io::ErrorKind::WouldBlock => break,
+            Err(e) => return Err(GuestMemoryError::IOError(e)),
+        }
+    }
+    Ok(total)
 }
```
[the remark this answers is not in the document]
d-e-s-o: I reverted to the original as per https://github.com/containers/libkrun/pull/588#discussion_r3009762779 and just allow deprecated method usage.

d-e-s-o:
Is there something blocking these changes or can we move ahead and get them merged?

slp (approved):
LGTM, thanks!

saku3:
Do you have an estimate for when a release including this change might be available?
I’d like to try using it with [youki](https://github.com/youki-dev/youki).

nohajc:
I don't think libkrun is published on crates.io yet but you can already try this by importing from git directly.

d-e-s-o:
> I don't think libkrun is published on crates.io yet

Is there a plan to do that, though? Right now one cannot release a package depending on it due to Git dependency.

nohajc:
I know @slp intentionally designed libkrun as a dynamic library but steps are being taken to enable static linking too (and reliable cross-compilation of init on macOS).

From my POV, it would make sense to fully integrate it into the Rust ecosystem by publishing to crates at some point.

Personally, I can live with the current state as well though.

slp:
> > I don't think libkrun is published on crates.io yet
> 
> Is there a plan to do that, though? Right now one cannot release a package depending on it due to Git dependency.

TBH, I haven't considered that option. The repo is a workspace, where `src/libkrun` depends on `src/*` crates. To publish it in crates.io, we'd need to publish every crate in src. At the very least, that would require doing some crate renaming to avoid collisions.

That said, seems doable, perhaps for even 1.18. Is there an actual interest on having libkrun in crates.io?

nohajc:
For my use case, it would simplify version management. Right now, when linking against the homebrew dynamic library I have basically no control over it. Works mostly fine but sometimes things might break...

On the other hand, I'm fine importing the rust lib directly from git because my own app is not on crates.io either.

d-e-s-o:
> Is there an actual interest on having libkrun in crates.io?

Yes, absolutely. As I said, one can't publish a crate with Git dependencies, so a single dependency not being on `crates.io` makes the entire thing unpublishable there.

PR #590, opened 2026-03-16
init: remove TEE-specific code

jakecorrenti:
Since we never built the init binary in this repository in the first place, remove it.


mtjhrc:
I welcome this change, but whats the plan - it will be moved to libkrunfw? Will you reuse some parts of init.c from here? (I suppose because you are modularizing some parts into separate files.

jakecorrenti:
> I welcome this change, but whats the plan - it will be moved to libkrunfw? Will you reuse some parts of init.c from here? (I suppose because you are modularizing some parts into separate files.

Yes, the Confidential Computing init code will be moved to libkrunfw (PR still needs to be made). The Conf. Computing init code is practically the same as the standard variant but with the attestation bits. The goal here was to make as much code reusable so that a PR to the init binary here didn't have to result in a PR in libkrunfw at the same time. The Conf.

For 2.0 I was envisioning the init code being re-written in Rust (which I've started working on in the background) and having any shared code be a separate crate. Then this would allow the init crates in libkrun and libkrunfw to just import the crate and not have to deal with any funny scripts at compile time. Just a thought, and it hasn't been fully thought through yet.

jakecorrenti:
Closing this and will be doing it in one go with the Rust re-write so it's less work

PR #592, opened 2026-03-17
vmm: fix non_x86_64 && non-TEE memory region setup

yzewei:
## Background

In non-x86 builds that don't use TEE features, `vstate.rs` uses `cfg!()` to distinguish between TEE and non-TEE paths.

`cfg!()` only generates a compile-time constant; the code in the branch still participates in compilation and type checking. As a result, in a normal riscv64 build, code related to `guest_memfd` and memory attributes will also be flagged for errors, leading to compilation failures.

Verify CMD
`cargo check -p vmm --target riscv64gc-unknown-linux-gnu`

### DIFF
--- a/src/vmm/src/linux/vstate.rs
+++ b/src/vmm/src/linux/vstate.rs
@@ -42,6 +42,8 @@ use crate::resources::TeeConfig;
 use crate::vmm_config::machine_config::CpuFeaturesTemplate;
 #[cfg(target_arch = "x86_64")]
 use cpuid::{c3, filter_cpuid, t2, VmSpec};
+#[cfg(not(feature = "tee"))]
+use kvm_bindings::kvm_userspace_memory_region;
 #[cfg(target_arch = "x86_64")]
 use kvm_bindings::{
     kvm_clock_data, kvm_debugregs, kvm_irqchip, kvm_lapic_state, kvm_mp_state, kvm_pit_state2,
@@ -49,14 +51,13 @@ use kvm_bindings::{
     KVM_CLOCK_TSC_STABLE, KVM_IRQCHIP_IOAPIC, KVM_IRQCHIP_PIC_MASTER, KVM_IRQCHIP_PIC_SLAVE,
     KVM_MAX_CPUID_ENTRIES,
 };
-use kvm_bindings::{
-    kvm_create_guest_memfd, kvm_userspace_memory_region, kvm_userspace_memory_region2,
-    KVM_API_VERSION, KVM_MEM_GUEST_MEMFD, KVM_SYSTEM_EVENT_RESET, KVM_SYSTEM_EVENT_SHUTDOWN,
-};
+#[cfg(all(feature = "tee", target_arch = "x86_64"))]
+use kvm_bindings::{kvm_create_guest_memfd, kvm_userspace_memory_region2, KVM_MEM_GUEST_MEMFD};
 #[cfg(feature = "tee")]
 use kvm_bindings::{kvm_enable_cap, KVM_CAP_EXIT_HYPERCALL, KVM_MEMORY_EXIT_FLAG_PRIVATE};
-#[cfg(not(target_arch = "riscv64"))]
+#[cfg(all(feature = "tee", target_arch = "x86_64"))]
 use kvm_bindings::{kvm_memory_attributes, KVM_MEMORY_ATTRIBUTE_PRIVATE};
+use kvm_bindings::{KVM_API_VERSION, KVM_SYSTEM_EVENT_RESET, KVM_SYSTEM_EVENT_SHUTDOWN};
 use kvm_ioctls::{Cap::*, *};
 use utils::eventfd::EventFd;
 use utils::signal::{register_signal_handler, sigrtmin, Killable};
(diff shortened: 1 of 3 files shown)


src/vmm/src/linux/vstate.rs
```diff
@@ -746,6 +755,11 @@ impl Vm {
             self.guest_memfds.push((Range { start, end }, guest_memfd));
         }
 
+        #[cfg(all(feature = "tee", not(target_arch = "x86_64")))]
+        {
```
jakecorrenti: IMO if we're going to return an error because the arch+TEE combination is invalid, then this is much too late. I think it should probably happen around here: https://github.com/containers/libkrun/blob/788cf91b53b0d94d33cffa45f7670594041cf0a6/src/vmm/src/builder.rs#L633

src/vmm/src/linux/vstate.rs
```diff
@@ -681,7 +684,8 @@ impl Vm {
         // as of late 2025. Also, on other architectures like aarch64, guest_memfd in
         // general is unstable for now, so don't try to use it without a reason.
 
-        if cfg!(not(feature = "tee")) {
+        #[cfg(not(feature = "tee"))]
```
jakecorrenti: I wonder if it would be cleaner to write something along the lines of this:
```rust
impl Vm {
    ...

    #[cfg(not(feature = "tee"))]
    fn create_guest_physical_memory_slot(...) -> Result<()> {}

    #[cfg(feature = "tee")]
    fn create_guest_physical_memory_slot(...) -> Result<()> {}

    fn memory_region_set(...) -> Result<()>  {
        ...

        #[cfg(not(feature="tee"))]
        self.create_guest_physical_memory_slot(...)?;
        #[cfg(feature="tee")]
        self.create_guest_physical_memory_slot(...)?;

        ...
    }

    ...
}
```

WDYT?

jakecorrenti (changes requested):
Can you populate the commit message with your PR description?

src/vmm/src/linux/vstate.rs
yzewei: Agreed, that's definitely clearer.

yzewei:
> Can you populate the commit message with your PR description?

sure.

src/vmm/src/linux/vstate.rs
yzewei: Yes, the latest commit split it into three branches: tee + x86 non_tee and invalid tee + non_x86.

src/vmm/src/linux/vstate.rs
```diff
@@ -42,21 +42,22 @@ use crate::resources::TeeConfig;
...
-use kvm_bindings::{
-    kvm_create_guest_memfd, kvm_userspace_memory_region, kvm_userspace_memory_region2,
-    KVM_API_VERSION, KVM_MEM_GUEST_MEMFD, KVM_SYSTEM_EVENT_RESET, KVM_SYSTEM_EVENT_SHUTDOWN,
-};
+#[cfg(all(feature = "tee", target_arch = "x86_64"))]
+use kvm_bindings::{kvm_create_guest_memfd, kvm_userspace_memory_region2, KVM_MEM_GUEST_MEMFD};
 #[cfg(feature = "tee")]
 use kvm_bindings::{kvm_enable_cap, KVM_CAP_EXIT_HYPERCALL, KVM_MEMORY_EXIT_FLAG_PRIVATE};
-#[cfg(not(target_arch = "riscv64"))]
+#[cfg(all(feature = "tee", target_arch = "x86_64"))]
```
dorindabassey, line 57: If youre adding guards on this, maybe check for leftovers like in worker.rs - https://github.com/containers/libkrun/blob/788cf91b53b0d94d33cffa45f7670594041cf0a6/src/vmm/src/worker.rs#L11-L12

yzewei: Good catch. I found the same leftover in worker.rs and tightened that path as well so the memory-attribute / guest_memfd handling is consistently gated to x86_64 + TEE.

dorindabassey (approved):
code LGTM, Thanks.

slp (approved):
LGTM. @jakecorrenti you've requested changes, PTAAL.

src/vmm/src/linux/vstate.rs
```diff
@@ -661,90 +662,121 @@ impl Vm {
...
+        // ignore an invalid TEE configuration and create a normal VM instead.
+        Err(Error::InvalidTee)
+    }
+
+    fn memory_region_set(
+        &mut self,
+        guest_mem: &GuestMemoryMmap,
+        region: &GuestRegionMmap,
+    ) -> Result<()> {
+        let host_addr = guest_mem.get_host_address(region.start_addr()).unwrap() as u64;
```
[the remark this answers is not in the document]
slp: We're doing it like this pretty much everywhere.

jakecorrenti:
sorry @yzewei this one fell through the cracks. can you rebase before I take a look at it again?

yzewei:
RFR @jakecorrenti

jakecorrenti (changes requested):
Can you please rebase instead of doing merge commits?

yzewei:
> Can you please rebase instead of doing merge commits?

Ok.

jakecorrenti (approved):
LGTM, thanks!

PR #593, opened 2026-03-17
lib: add krun_set_init_binary, backcompat via new krun_init crate

ggoodman:
## Intent

Our goal with this PR is to prepare for a future Rust API where the main crate does **not** always embed or compile an `init` binary as part of being usable.

Today, `init.krun` is effectively a hard-wired implementation detail of the lower layers (especially `devices`), which makes sense for the current C-oriented flow, but creates friction for a slim Rust-first API design. We want to preserve the existing behavior for C consumers while moving toward a model where init payload is an explicit input, not an unconditional side effect of depending on core crates.

In short: keep the current default UX where needed, but stop forcing the init payload into every build shape.

## Approach

This PR splits policy from mechanism and introduces an explicit init contract.

First, it moves default init embedding/build responsibilities into a dedicated sibling crate (`krun_init`) instead of keeping that logic in `devices`. That lets us preserve current behavior (default payload is still available) without making `devices` itself always own the init artifact.

Second, it changes virtio-fs passthrough to accept init payload through configuration (`Option<Arc<[u8]>>`) rather than via a crate-level static include. The synthetic `/init.krun` inode is now exposed only when payload is present, which makes the behavior explicit and composable.

Third, it replaces the stringly cmdline check with a typed boot contract. We now express the requirement as `InitPolicy::InitKrunFromVirtioFs`, and enforce it centrally in `vmm::builder`. This keeps validation robust and avoids heuristic coupling to command-line text parsing.

Fourth, on the C side, it keeps backward-compatible defaults by auto-wiring the default init payload, and additionally introduces a runtime override API (`krun_set_init_binary`) so callers can inject their own init bytes for a context.

## Trade-offs

The main benefit is architectural clarity: core crates no longer implicitly force init embedding semantics, and the contract around `/init.krun` is explicit and validated in one place. This also gives us a clean migration path toward a Rust API where init choice is deliberate.

The cost is extra plumbing and a slightly larger configuration surface (`init_policy`, payload propagation through fs config). We also introduce another crate (`krun_init`) in the dependency graph, which is intentional but still additional structure to maintain.

Overall, this is a deliberate trade: modest complexity increase now in exchange for a cleaner long-term API boundary and less hidden policy in shared internals.

### DIFF
--- a/include/libkrun.h
+++ b/include/libkrun.h
@@ -110,6 +110,24 @@ int32_t krun_set_vm_config(uint32_t ctx_id, uint8_t num_vcpus, uint32_t ram_mib)
 int32_t krun_set_root(uint32_t ctx_id, const char *root_path);
 
 /**
+ * Sets the init binary payload that will be exposed as /init.krun inside the guest.
+ *
+ * IMPORTANT LIFETIME CONTRACT:
+ * The memory range [init_binary, init_binary + init_binary_len) is NOT copied.
+ * The caller MUST keep that memory valid and unchanged for the full VM lifetime
+ * in this context (until krun_start_enter() returns and the VM is torn down).
+ *
+ * Arguments:
+ *  "ctx_id"          - the configuration context ID.
+ *  "init_binary"     - pointer to init binary bytes.
+ *  "init_binary_len" - number of bytes at init_binary.
+ *
+ * Returns:
+ *  Zero on success or a negative error number on failure.
+ */
+int32_t krun_set_init(uint32_t ctx_id, const uint8_t *init_binary, size_t init_binary_len);
+
+/**
  * DEPRECATED. Use krun_add_disk instead.
  *
  * Sets the path to the disk image that contains the file-system to be used as root for the microVM.
--- a/src/devices/src/virtio/fs/linux/passthrough.rs
+++ b/src/devices/src/virtio/fs/linux/passthrough.rs
@@ -33,8 +33,6 @@ const EMPTY_CSTR: &[u8] = b"\0";
 const PROC_CSTR: &[u8] = b"/proc/self/fd\0";
 const INIT_CSTR: &[u8] = b"init.krun\0";
 
-static INIT_BINARY: &[u8] = include_bytes!(env!("KRUN_INIT_BINARY_PATH"));
-
 type Inode = u64;
 type Handle = u64;
 
@@ -328,6 +326,7 @@ pub struct Config {
     /// Table of exported FDs to share with other subsystems.
     pub export_table: Option<ExportTable>,
     pub allow_root_dir_delete: bool,
+    pub init_payload: Option<&'static [u8]>,
 }
 
 impl Default for Config {
@@ -343,6 +342,7 @@ impl Default for Config {
             export_fsid: 0,
             export_table: None,
             allow_root_dir_delete: false,
+            init_payload: None,
         }
     }
 }
@@ -439,7 +439,11 @@ impl PassthroughFs {
         Ok(PassthroughFs {
             inodes: RwLock::new(MultikeyBTreeMap::new()),
             next_inode: AtomicU64::new(fuse::ROOT_ID + 2),
-            init_inode: fuse::ROOT_ID + 1,
+            init_inode: if cfg.init_payload.is_some() {
+                fuse::ROOT_ID + 1
+            } else {
+                0
+            },
 
             handles: RwLock::new(BTreeMap::new()),
             next_handle: AtomicU64::new(1),
@@ -454,6 +458,12 @@ impl PassthroughFs {
             cap_fowner,
             cfg,
         })
+    }
+
+    fn init_payload(&self) -> io::Result<&[u8]> {
+        self.cfg
+            .init_payload
+            .ok_or_else(|| io::Error::from_raw_os_error(libc::ENOENT))
     }
 
     fn open_inode(&self, inode: Inode, mut flags: i32) -> io::Result<File> {
@@ -996,7 +1006,7 @@ impl FileSystem for PassthroughFs {
 
         if self.init_inode != 0 && name == init_name {
             let mut st: libc::stat64 = unsafe { mem::zeroed() };
-            st.st_size = INIT_BINARY.len() as i64;
+            st.st_size = self.init_payload()?.len() as i64;
             st.st_ino = self.init_inode;
             st.st_mode = 0o100_755;
 
@@ -1235,13 +1245,16 @@ impl FileSystem for PassthroughFs {
     ) -> io::Result<usize> {
         debug!("read: {inode:?}");
         if inode == self.init_inode {
+            let init_payload = self.init_payload()?;
             let off: usize = offset.try_into().map_err(|_| einval())?;
-            let len = if off + (size as usize) < INIT_BINARY.len() {
-                size as usize
-            } else {
-                INIT_BINARY.len() - off
-            };
-            return w.write(&INIT_BINARY[off..(off + len)]);
+            if off >= init_payload.len() {
+                // Read past EOF should return 0 bytes.
+                return Ok(0);
+            }
+
+            let remaining = init_payload.len() - off;
+            let len = remaining.min(size as usize);
+            return w.write(&init_payload[off..(off + len)]);
         }
 
         let data = self
@@ -2088,6 +2101,7 @@ impl FileSystem for PassthroughFs {
         debug!("setupmapping: ino {inode:?} addr={addr:x} len={len}");
 
         if inode == self.init_inode {
+            let init_payload = self.init_payload()?;
             let ret = unsafe {
                 libc::mmap(
                     addr as *mut libc::c_void,
@@ -2102,15 +2116,15 @@ impl FileSystem for PassthroughFs {
                 return Err(io::Error::last_os_error());
             }
 
-            let to_copy = if len as usize > INIT_BINARY.len() {
-                INIT_BINARY.len()
+            let to_copy = if len as usize > init_payload.len() {
+                init_payload.len()
             } else {
                 len as usize
             };
             unsafe {
                 libc::memcpy(
                     addr as *mut libc::c_void,
-                    INIT_BINARY.as_ptr() as *const _,
+                    init_payload.as_ptr() as *const _,
                     to_copy,
                 )
             };
--- a/src/devices/src/virtio/fs/macos/passthrough.rs
+++ b/src/devices/src/virtio/fs/macos/passthrough.rs
@@ -37,8 +37,6 @@ const SECURITY_CAPABILITY: &[u8] = b"security.capability\0";
 
 const UID_MAX: u32 = u32::MAX - 1;
 
-static INIT_BINARY: &[u8] = include_bytes!(env!("KRUN_INIT_BINARY_PATH"));
-
 type Inode = u64;
 type Handle = u64;
 
@@ -517,6 +515,7 @@ pub struct Config {
     /// Table of exported FDs to share with other subsystems. Not supported for macos.
     pub export_table: Option<ExportTable>,
     pub allow_root_dir_delete: bool,
+    pub init_payload: Option<&'static [u8]>,
 }
 
 impl Default for Config {
@@ -532,6 +531,7 @@ impl Default for Config {
             export_fsid: 0,
             export_table: None,
             allow_root_dir_delete: false,
+            init_payload: None,
         }
     }
 }
@@ -580,7 +580,11 @@ impl PassthroughFs {
         Ok(PassthroughFs {
             inodes: RwLock::new(MultikeyBTreeMap::new()),
             next_inode: AtomicU64::new(fuse::ROOT_ID + 2),
-            init_inode: fuse::ROOT_ID + 1,
+            init_inode: if cfg.init_payload.is_some() {
+                fuse::ROOT_ID + 1
+            } else {
+                0
+            },
 
             handles: RwLock::new(BTreeMap::new()),
             next_handle: AtomicU64::new(1),
@@ -592,6 +596,12 @@ impl PassthroughFs {
             announce_submounts: AtomicBool::new(false),
             cfg,
         })
+    }
+
+    fn init_payload(&self) -> io::Result<&[u8]> {
+        self.cfg
+            .init_payload
+            .ok_or_else(|| linux_error(io::Error::from_raw_os_error(libc::ENOENT)))
     }
 
     fn inode_to_handle(&self, inode: Inode, supports_fd: bool) -> io::Result<InodeHandle> {
@@ -1205,7 +1215,7 @@ impl FileSystem for PassthroughFs {
 
         if self.init_inode != 0 && name == _init_name {
             let mut st: bindings::stat64 = unsafe { mem::zeroed() };
-            st.st_size = INIT_BINARY.len() as i64;
+            st.st_size = self.init_payload()?.len() as i64;
             st.st_ino = self.init_inode;
             st.st_mode = 0o100_755;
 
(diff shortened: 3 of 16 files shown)


include/libkrun.h
```diff
@@ -109,6 +109,22 @@ int32_t krun_set_vm_config(uint32_t ctx_id, uint8_t num_vcpus, uint32_t ram_mib)
...
+ *
+ * Arguments:
+ *  "ctx_id"          - the configuration context ID.
+ *  "init_binary"     - pointer to init binary bytes.
+ *  "init_binary_len" - number of bytes at init_binary.
+ *
+ * Returns:
+ *  Zero on success or a negative error number on failure.
+ */
+int32_t krun_set_init_binary(uint32_t ctx_id, const uint8_t *init_binary, size_t init_binary_len);
```
tylerfanelli: `krun_set_init` would be a bit more succinct.

```suggestion
int32_t krun_set_init(uint32_t ctx_id, const uint8_t *bin, size_t size);
```

src/devices/src/virtio/fs/macos/passthrough.rs
```diff
@@ -517,6 +515,7 @@ pub struct Config {
     /// Table of exported FDs to share with other subsystems. Not supported for macos.
     pub export_table: Option<ExportTable>,
     pub allow_root_dir_delete: bool,
+    pub init_payload: Option<Arc<[u8]>>,
```
tylerfanelli: Should this perhaps be an `Arc<[u8]>` rather than `Option<Arc<[u8]>>`, in which if there is no init binary set, we fall back to the default?

src/devices/src/virtio/fs/linux/passthrough.rs
```diff
@@ -1235,13 +1246,17 @@ impl FileSystem for PassthroughFs {
     ) -> io::Result<usize> {
         debug!("read: {inode:?}");
         if inode == self.init_inode {
+            let init_payload = self.init_payload()?;
             let off: usize = offset.try_into().map_err(|_| einval())?;
-            let len = if off + (size as usize) < INIT_BINARY.len() {
+            if off >= init_payload.len() {
+                return Ok(0);
+            }
```
tylerfanelli, line 1253: Can you explain why zero is returned?

src/devices/src/virtio/fs/macos/passthrough.rs
```diff
@@ -1457,15 +1468,19 @@ impl FileSystem for PassthroughFs {
...
         debug!("read: {inode:?}");
         if inode == self.init_inode {
+            let init_payload = self.init_payload()?;
             let off: usize = offset
                 .try_into()
                 .map_err(|_| io::Error::from_raw_os_error(libc::EINVAL))?;
-            let len = if off + (size as usize) < INIT_BINARY.len() {
+            if off >= init_payload.len() {
+                return Ok(0);
+            }
```
tylerfanelli, line 1477: Ditto here, why return zero?

src/devices/src/virtio/fs/macos/passthrough.rs
ggoodman: Wouldn't that result in any binary built off this now forced to include the bytes of the default init? My intention here is to avoid that by putting it in another compilation unit and forcing the user to provide it.

But to maintain parity with the c API, I did the default fallback.

tylerfanelli: Yea, I guess there is no way to "fall back" without including the bytes within the compilation.

But are we requiring that an init binary always be set now?

ggoodman: If we are that is not the intention. My intention was as follows:

1. For the current c api, always fall back to the current init.
2. For the speculative rust api, require that an init binary be configured. Document that the built-in init binary is in the `libkrun_init` crate.

ggoodman: The new fallback mechanism in the c api works based on this: https://github.com/ggoodman/libkrun/blob/b9881fd56291a9c67ebd556ec3cd4177536166ee/src/libkrun/src/lib.rs#L92

src/devices/src/virtio/fs/macos/passthrough.rs
ggoodman: Added a comment. Read past EOF should return 0 bytes read.

src/devices/src/virtio/fs/linux/passthrough.rs
ggoodman: Added a comment. Read past EOF should return 0 bytes read.

include/libkrun.h
ggoodman: Updated

src/libkrun/src/lib.rs
```diff
@@ -603,6 +623,35 @@ pub unsafe extern "C" fn krun_set_root(ctx_id: u32, c_root_path: *const c_char)
...
+pub unsafe extern "C" fn krun_set_init(
+    ctx_id: u32,
+    init_binary: *const u8,
+    init_binary_len: size_t,
+) -> i32 {
+    if init_binary.is_null() || init_binary_len == 0 {
+        return -libc::EINVAL;
+    }
+
+    let payload = Arc::<[u8]>::from(slice::from_raw_parts(init_binary, init_binary_len));
```
d-e-s-o: I feel a bit uneasy about this potentially sizable copy (and heap allocation; for something that with great likelihood is already statically included in the binary anyway). It strikes me as something we should try and avoid, if at all possible. Have you considered replacing the `Arc<[u8]>` construct with a `&'static [u8]`? Would it impede usability significantly?

mtjhrc: I think, I would prefer the `&'static [u8]` too. 

`'static` is currently the correct lifetime, because we exit the process once VM quits, but this will probably change in the future and this may become more unsafe. For a C API like we have now, having a huge static buffer be caller owned is reasonable behavior IMO. (I wonder though if it will be possible to design the Rust API so it's possible to name a `'vmm` lifetime...?)

ggoodman: OK, so I have tried addressing this in fcd6927 without trying to pretend to own the memory range supplied to the c api by using an internal enum of owned vs static byte slices. The init binary that is now embedded in `krun_init` is of the static variant allowing a zero-copy use-case. However, if folks supply an init binary via `krun_set_init()`, it will be treated as owned and will result in a copy.

Later, if we have a native rust api, this will allow embedders of libkrun to use the `include_bytes` macro and the `Static()` variant if they want.

d-e-s-o: > 'static is currently the correct lifetime, because we exit the process once VM quits, but this will probably change in the future and this may become more unsafe.

I'd think `'static` would be correct no matter what, right? It would just over-constrain. However, I don't think you can have non-static lifetimes on global data in Rust (and at least currently there is still a bunch of global state, it seems).

That being said, the most flexibility for users (and, hence, future-proof API design) would perhaps be achieved be some custom `enum` ala:
```rust
enum InitBinary {
  Static(&'static [u8]),
  Dynamic(Arc<[u8]>),
}
```
Personally I'd probably still err on the side of just using `&'static [u8]` to force callers to think about what they are doing, but I can't anticipate how this API will be used most frequently.

d-e-s-o: > OK, so I have tried addressing this in [fcd6927](https://github.com/containers/libkrun/commit/fcd69278410492707fda649cbeffa40419207604) without trying to pretend to own the memory range supplied to the c api by using an internal enum of owned vs static byte slices.

Thanks! Do we still need the `LazyLock` thingy? Constructing an enum variant with static data in a const context should be fine, but haven't tried removing.

ggoodman: Good call. Dropped it in 496baaf.

mtjhrc: > I'd think 'static would be correct no matter what, right? It would just over-constrain.

Yes exactly.

> However, if folks supply an init binary via krun_set_init(), it will be treated as owned and will result in a copy.

I don't know that feels a bit weird to me, I think it sense for a big read-only buffer like this be caller owned.

ggoodman: > I don't know that feels a bit weird to me, I think it sense for a big read-only buffer like this be caller owned.

It was my intention to treat `::Owned()` as caller-owned. This might be a case of poor naming.

For the owned case, we treat the memory as owned by the caller which is why we copy it.

An embedder of the future rust library would use the `::Static()` variant.

Right now, the new c API will systematically treat the memory as caller owned and thus use `::Owned()`.

mtjhrc: Ah, we misunderstood each other. 
By "caller owned" I meant that caller owns the memory (possibly dynamically allocated or their static memory) and guarantees it will exist during the VMM lifetime. In that case `krun_set_init` doesn't need to copy it (so it actually corresponds to your `static` case). 

Both have pros and cons, I am not really in strongly in favor of one or the other.

d-e-s-o: > By "caller owned" I meant that caller owns the memory (possibly dynamically allocated or their static memory) and guarantees it will exist during the VMM lifetime.

Personally, I'd also prefer these semantics.

ggoodman: Oh I see. I did not anticipate that as the default.

In that case, perhaps the union type is not necessary and the ownership requirements can simply be documented on the c api.

It would state that the caller guarantees that the memory will outlive the vm lifetime.

So that leaves some options:

1. Status quo.
2. Keep the union type with the expectation that `::Owned()` may be helpful for some use-cases of a future rust api _but switch the new c api method to assume `::Static()`.
3. Rip out the whole union type and switch everything conceptually to the `::Static()`.

mtjhrc: > 3. Rip out the whole union type and switch everything conceptually to the ::Static().

I'd go for this

ggoodman: ...then so shall it be (as of b89d788)!

ggoodman:
@mtjhrc / @slp is this something you're willing to entertain in 1.x?

mtjhrc:
Personally, I would be ok with this for 1.18 (last feature release of 1.x branch), because the feature is small and isolated. Please squash the commits and directly introduce just the final simple &'static version only. (instead of the Arc/lazylock/enum sidequests).

ggoodman:
> Please squash the commits and directly introduce just the final simple &'static version only. (instead of the Arc/lazylock/enum sidequests).

Rebased and pushed.

I saw that you merged the crate PR (#588) and I strongly believe that this needs to land in the same version as that or it will result in a breaking change.

src/libkrun/src/lib.rs
```diff
@@ -603,6 +618,38 @@ pub unsafe extern "C" fn krun_set_root(ctx_id: u32, c_root_path: *const c_char)
...
+        Entry::Occupied(mut ctx_cfg_entry) => {
+            let ctx_cfg = ctx_cfg_entry.get_mut();
+            ctx_cfg.set_init_binary(payload);
+
+            for fs_cfg in &mut ctx_cfg.vmr.fs {
+                fs_cfg.init_payload = Some(payload);
+            }
+        }
+        Entry::Vacant(_) => return -libc::ENOENT,
+    }
```
mtjhrc, line 648: Instead of having to save the `init_binary` in both the `ctx_cfg` and `fs_cfg`, I would suggest this just sets the init payload for fs device with tag "/dev/root".
If such fs device doesn't exist this returns an error.

ggoodman: @mtjhrc, I took a stab at making the suggested change and came to the conclusion that the juice is not worth the squeeze. 

The main downside is that it would result in `krun_set_init` becoming _order-dependent_; embedders would _need_ to have set up their root filesystem before calling `krun_set_init`. I think these two operations are probably conceptually orthogonal from a user's perspective so the order-dependence creates an unnecessary failure mode.

That being said, I'm not at all an expert on this codebase so maybe I'm not seeing the better approach or I'm projecting constraints that the maintainers don't impose upon themselves.

LMK.

ggoodman: CI failure a flake? I can't re-run the tests myself.

src/devices/src/virtio/fs/linux/passthrough.rs
```diff
@@ -456,6 +460,10 @@ impl PassthroughFs {
         })
     }
 
+    fn init_payload(&self) -> io::Result<&[u8]> {
+        self.cfg.init_payload.ok_or_else(ebadf)
```
d-e-s-o: Is `EBADF` really an appropriate error code to return here? I feel like `ENOENT` is more common for this kind of condition, perhaps `ENOMEDIUM` if we'd want something more unique.

src/krun_init/build.rs
```diff
@@ -39,20 +39,35 @@ fn build_default_init() -> PathBuf {
         .unwrap_or_else(|e| panic!("failed to execute {cc}: {e}"));
 
     if !status.success() {
-        panic!("failed to compile init/init.c: {status}");
+        return None;
```
d-e-s-o: I'd probably keep the existing behavior: if the build fails that should be signaled with a dedicated error message, in my opinion. That decouples failure from optionality more cleanly, I'd say.

src/libkrun/src/lib.rs
```diff
@@ -232,6 +236,16 @@ impl ContextConfig {
...
+    #[cfg(not(feature = "tee"))]
+    fn set_init_binary(&mut self, init_binary: &'static [u8]) {
+        self.init_payload = Some(init_binary);
+    }
+
+    #[cfg(not(feature = "tee"))]
+    fn get_init_binary(&self) -> &'static [u8] {
+        self.init_payload.unwrap_or(DEFAULT_INIT_PAYLOAD)
+    }
+
```
d-e-s-o, line 248: Not familiar with the `tee` feature per-se, but just pointing out that Cargo features should be [additive](https://doc.rust-lang.org/cargo/reference/features.html#:~:text=A%20consequence%20of%20this%20is%20that%20features%20should%20be%20additive), or problems await. It seems that is violated here, because enabling of a feature removes functionality. Perhaps it's fine if it's a niche thing for a very specific use case (or perhaps there is precedent in the repository already), but just pointing that out.

mtjhrc: Yes it violates it here, but unfortunately so do many other places in libkrun code base. Yes this means this will cause problems if multiple crates in the compilation use libkrun and features get unified :face_with_diagonal_mouth:  Also I think some function might do something different based on features being enabled. This is something we need to fix, but I don't us being able to do this in libkrun 1.x.

src/krun_init/build.rs
ggoodman: Agree. Working on it.

ggoodman: I'll rebase this if it looks good. I kept it separate to make it easy for you to review in isolation: 58210705030555aa01a7a086b83df0d69e794e7a

d-e-s-o: Seems fine to me, thanks!

src/devices/src/virtio/fs/linux/passthrough.rs
ggoodman: Done in e129fd6. Will rebase if you confirm that it looks good to you, @d-e-s-o.

src/libkrun/src/lib.rs
mtjhrc: > I think these two operations are probably conceptually orthogonal from a user's perspective so the order-dependence creates an unnecessary failure mode

Yeah that is a valid point too. Yeah I guess it's fine currently, we'll change it in libkrun 2.x anyway.

The current API is a mess, some things are order dependent some are not. Ideally this would take a handle to some sort of filesystem device object (some function already work on handles, well IDs e.g. krun_add_console_port_* ), but filesystems creation functions don't return the ID (index).

> CI failure a flake? I can't re-run the tests myself.

Yes some sort of race condition causes an ENOMEM in the guest kernel (we had a similar issue before related to host kernel, but this seems different). Unrelated to this PR.

src/devices/src/virtio/fs/linux/passthrough.rs
d-e-s-o: Looks good as well, thanks!

ggoodman: Rebased. Thanks.

src/krun_init/build.rs
ggoodman: Rebased. Thanks.

ggoodman:
Need me to rebase off main?

d-e-s-o:
FWIW, I'd love to see something like this (specifying memory range with binary) possible for the kernel as well. I see that on `x86_64` we already have a path that just `mmap`s the kernel and boots from that (though it still is behind an API that forces a path input).

Right now, even if a kernel is already present in memory, one seems to be forced to write it out into a file first in order to boot. For a MicroVM where quick boot times come into the picture, it would be good to be given the option to side-step that overhead.

Is there any interest for such an API?

ggoodman:
I am interested and it was a next step that I asked about on the Matrix chat. I understood that there is interest but that it would be more of a 2.0 thing (at least if implemented similar to this PR).

mtjhrc:
I believe this has already been superseded by #673 and  #714 which have been released in 1.19.0.

1. (optional) disable init-blob feature at compile time (cargo feature)
2. (optional if you disabled it at compile time, but required if you haven't) call `krun_disable_implicit_init()`
3. Call `krun_fs_add_overlay_file` with `"init.krun"` (or inject other virtual file(s))

PR #597, opened 2026-03-20
virtio/fs/macos: fix fallocate semantics

slp:
The translation of Linux's fallocate semantics to macOS was badly broken. While we can't fully preserve the original semantics, since macOS doesn't have posix_fallocate and F_PREALLOCATE doesn't allow to allocate arbitrary ranges, we can do something close enough to work fine for the vast majority of guest applications.

### DIFF
--- a/src/devices/src/virtio/fs/macos/passthrough.rs
+++ b/src/devices/src/virtio/fs/macos/passthrough.rs
@@ -2209,7 +2209,7 @@ impl FileSystem for PassthroughFs {
         _ctx: Context,
         inode: Inode,
         handle: Handle,
-        _mode: u32,
+        mode: u32,
         offset: u64,
         length: u64,
     ) -> io::Result<()> {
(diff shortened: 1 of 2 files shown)


slp:
@mtjhrc @dorindabassey @mz-pdm PTAL. If you can't test on macOS, a visual inspection is still appreciated.

JAORMX:
@slp I'm still getting acquainted with the codebase. I ran this through Claude, so take it with a grain of salt, but it poitned out the following:

> in the keep_size path, st.st_blocks * st.st_blksize overestimates disk usage. st_blocks is in 512-byte units per POSIX, not st_blksize units. On APFS (st_blksize = 4096) this inflates the value by 8x, causing needed allocations to be skipped. Should be st.st_blocks * 512.

src/devices/src/virtio/fs/macos/passthrough.rs
```diff
@@ -2224,36 +2224,79 @@ impl FileSystem for PassthroughFs {
...
+                // blocks to the file's end.
+                //
+                // The best thing we can do here is extend the file to (offset + length).
+                // This doesn't adhere to the same semantics, but should work fine (albeit
+                // less performant) for most guest applications.
+                let st = fstat(fd, true)?;
+                let new_length = (offset + length) as i64;
+
+                if keep_size {
+                    // Check the number of allocate blocks instead of the file size.
```
mz-pdm: s/allocate/allocated/ ?

mz-pdm:
> @slp I'm still getting acquainted with the codebase. I ran this through Claude, so take it with a grain of salt, but it poitned out the following:
> 
> > in the keep_size path, st.st_blocks * st.st_blksize overestimates disk usage. st_blocks is in 512-byte units per POSIX, not st_blksize units. On APFS (st_blksize = 4096) this inflates the value by 8x, causing needed allocations to be skipped. Should be st.st_blocks * 512.

According to https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/fstat.2.html, Claude is right. `st_blksize` is the optimal I/O block size and `st_blocks` is the number of 512-byte units.

Other than that (and the possible typo in the comment), the PR looks good to me (but I know nothing about the topic, just using available documentation).

src/devices/src/virtio/fs/macos/passthrough.rs
slp: Fixed, thanks!

slp:
> @slp I'm still getting acquainted with the codebase. I ran this through Claude, so take it with a grain of salt, but it poitned out the following:
> 
> > in the keep_size path, st.st_blocks * st.st_blksize overestimates disk usage. st_blocks is in 512-byte units per POSIX, not st_blksize units. On APFS (st_blksize = 4096) this inflates the value by 8x, causing needed allocations to be skipped. Should be st.st_blocks * 512.

Fixed, thanks!

src/devices/src/virtio/fs/macos/passthrough.rs
```diff
@@ -2224,36 +2224,79 @@ impl FileSystem for PassthroughFs {
...
-        if res == 0 {
-            Ok(())
-        } else {
-            Err(linux_error(io::Error::last_os_error()))
+                let res = unsafe { libc::fcntl(fd, libc::F_PUNCHHOLE, &mut hole as *mut _) };
+                if res < 0 {
+                    return Err(linux_error(io::Error::last_os_error()));
+                }
+            }
+            _ => unreachable!(),
```
mtjhrc, line 2296: This doesn't seem great. I think we should maybe `log:.warn!` this and return `EOPNOTSUPP`.

mtjhrc: Wait nevermind I missed the check in the code.

dorindabassey:
Can't test on macOS but the code LGTM!

mz-pdm:
LGTM AFAICT.

mtjhrc (approved):
LGTM, tested this a bit on macOS and seems to work (and least on the surface level).

PR #603, opened 2026-03-24
tests: Add virtio-net tests + supporting testing framework improvements

mtjhrc:
Changes:
  - Move network namespace isolation (unshare) from a single global wrapper to per-test isolation in the runner                         
  - Test runner can now clean up background processes (e.g. gvproxy, vmnet-helper) after each test
  - Add configurable per-test timeout to prevent hanging the suite                                                        
  - Tests can now specify a Containerfile to build a guest rootfs via podman — the runner builds the image, exports it, and mounts it as the guest's virtiofs root. Used by iperf3 tests to get a Fedora-based rootfs with iperf3 pre-installed.                              
  - Introduce `TestOutcome::Report` so tests can produce structured output (terminal text + GitHub-flavored markdown for CI summaries) instead of just pass/fail                                                                                                             
  - Introduce functional virtio-net tests for passt, tap, gvproxy, and vmnet-helper backends using guest DHCP
  - Introduce parametrized iperf3 performance tests that run iperf3 client in the guest against a host server and report throughput

### DIFF
--- a/tests/test_cases/src/test_net/vmnet_helper.rs
+++ b/tests/test_cases/src/test_net/vmnet_helper.rs
@@ -1,0 +1,185 @@
+//! vmnet-helper backend for virtio-net test (macOS only)
+
+use crate::{krun_call, ShouldRun, TestSetup};
+use krun_sys::{
+    NET_FEATURE_CSUM, NET_FEATURE_GUEST_CSUM, NET_FEATURE_GUEST_TSO4, NET_FEATURE_HOST_TSO4,
+};
+use nix::libc;
+use std::ffi::CString;
+use std::io::{BufRead, BufReader, Read};
+use std::process::{Command, Stdio};
+
+type KrunAddNetUnixgramFn = unsafe extern "C" fn(
+    ctx_id: u32,
+    c_path: *const std::ffi::c_char,
+    fd: i32,
+    c_mac: *mut u8,
+    features: u32,
+    flags: u32,
+) -> i32;
+
+fn get_krun_add_net_unixgram() -> KrunAddNetUnixgramFn {
+    let symbol = CString::new("krun_add_net_unixgram").unwrap();
+    let ptr = unsafe { libc::dlsym(libc::RTLD_DEFAULT, symbol.as_ptr()) };
+    assert!(!ptr.is_null(), "krun_add_net_unixgram not found");
+    unsafe { std::mem::transmute(ptr) }
+}
+
+const VMNET_HELPER_PATH: &str = match option_env!("VMNET_HELPER_PATH") {
+    Some(path) => path,
+    None => "/opt/homebrew/opt/vmnet-helper/libexec/vmnet-helper",
+};
+
+fn vmnet_helper_path() -> Option<&'static str> {
+    std::path::Path::new(VMNET_HELPER_PATH)
+        .exists()
+        .then_some(VMNET_HELPER_PATH)
+}
+
+/// Parse a MAC address string like "1e:d4:d1:27:4b:bf" into 6 bytes.
+fn parse_mac(s: &str) -> Option<[u8; 6]> {
+    let parts: Vec<&str> = s.split(':').collect();
+    if parts.len() != 6 {
+        return None;
+    }
+    let mut mac = [0u8; 6];
+    for (i, part) in parts.iter().enumerate() {
+        mac[i] = u8::from_str_radix(part, 16).ok()?;
+    }
+    Some(mac)
+}
+
+struct VmnetConfig {
+    fd: i32,
+    mac: [u8; 6],
+    pid: u32,
+}
+
+/// Start vmnet-helper with `--fd 3`, wait for its JSON config on stdout,
+/// and return the fd + MAC address from vmnet.
+///
+/// Creates a `SOCK_DGRAM` socketpair, passes one end to vmnet-helper as fd 3
+/// (matching what `vmnet-client` does), and returns the other end for use
+/// with `krun_add_net_unixgram`.
+fn start_vmnet_helper(log_path: &std::path::Path) -> std::io::Result<VmnetConfig> {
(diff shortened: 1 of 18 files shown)


tests/test_cases/src/test_net/vmnet_helper.rs
```diff
@@ -0,0 +1,179 @@
...
+    }
+    let (our_fd, helper_fd) = (fds[0], fds[1]);
+
+    let log_file = std::fs::File::create(log_path)?;
+
+    let mut child = Command::new(helper)
+        .arg("--fd")
+        .arg(helper_fd.to_string())
+        // TODO: Enable when TSO is working.
+        // .arg("--enable-tso")
```
nirs: What do you mean by "When TSO is working"?

`--enable-tso` and `--enable-checksum-offload` must be specified together. This will not work without --enable-tso.

tests/test_cases/src/test_net/vmnet_helper.rs
```diff
@@ -0,0 +1,179 @@
...
+        .ok_or_else(|| std::io::Error::other(format!("invalid MAC address: {mac_str}")))?;
+
+    // Increase socket buffer sizes so libkrun's Unixgram backend (which uses
+    // the fd path and does NOT set these) can batch frames without drops.
+    let buf_size: libc::c_int = 7 * 1024 * 1024;
+    unsafe {
+        libc::setsockopt(
+            our_fd,
+            libc::SOL_SOCKET,
+            libc::SO_SNDBUF,
```
nirs: Send buffer size is not used for buffering. It determines the maximum frame size. Should be 65550 when using offloading.

tests/test_cases/src/test_net/vmnet_helper.rs
```diff
@@ -0,0 +1,179 @@
...
+    let mut config = start_vmnet_helper(&vmnet_log)?;
+    test_setup.register_cleanup_pid(config.pid);
+
+    unsafe {
+        krun_call!(get_krun_add_net_unixgram()(
+            ctx,
+            std::ptr::null(),
+            config.fd,
+            config.mac.as_mut_ptr(),
+            // TODO: Enable TSO (HOST_TSO4, GUEST_TSO4, etc.) — currently hangs.
```
nirs: I test this with krunkit - works fine
https://github.com/containers/krunkit/blob/02d4161f19b172c85d686ce2532b09e6ad1dd398/src/virtio.rs#L511

tests/test_cases/src/test_net/vmnet_helper.rs
```diff
@@ -0,0 +1,180 @@
...
+        }
+    }
+
+    let log_file = std::fs::File::create(log_path)?;
+
+    let mut child = Command::new(helper)
+        .arg("--fd")
+        .arg(helper_fd.to_string())
+        .arg("--enable-tso")
+        .arg("--enable-checksum-offload")
```
nirs, line 105: It will be useful to test with and without offloading. Offloading does not work for use cases, for example when clients is in a Kubernets pod network. We get very low bandwidth (1000x slower) and huge amount of retransmits.

https://github.com/nirs/vmnet-helper/blob/615b5bb5dc4def4e9453e7f4caaa7ff9d7cdd3b3/performance/2026-02/M2/minikube/report.md#pod-network---iperf3-on-vm

tests/test_cases/src/test_net/vmnet_helper.rs
mtjhrc: Fixed, it was using wrong socket buffer size, causing `sendmsg_x` to fail in vmnet-helper.

tests/test_cases/src/test_net/vmnet_helper.rs
mtjhrc: Fixed - same socket buffer size issue.

tests/test_cases/src/test_net/vmnet_helper.rs
mtjhrc: Hmm, makes sense. 

Also I am still unsure how I want to go about the performance testing. I mean if we should have many fixed variants of the tests here or just make them more parametrized and make the user pick. 
Thing is we probably won't be running these performance tests in a CI for the foreseeable future anyway (no full macOS CI, not sure how consistent the performance baseline is on Linux...)

I put the performance tests here mostly because it's a really convenient way to run them locally for me when making changes to libkrun code (and reviewing PRs).

nirs: You can check vmnet-helper benchmarking infrastructure. It uses yaml to describe benchmarks parameters:
https://github.com/nirs/vmnet-helper/tree/main/benchmarks

and plots parameters:
https://github.com/nirs/vmnet-helper/blob/main/plots/offloading.yaml

The bench run command read the yaml file and run all the benchmarks, saving results to json files:
https://github.com/nirs/vmnet-helper/blob/main/bench

The bench plot command read plot yaml files describing the plots and read data generated by the bench run.

The benchmarks are very noisy since we run multiple vms and we cannot control what macOS run in the background.

To compare results for PR you need to run the same benchmark twice, once with the previous version and once with the change. The run must be long enough to mitigate random noise during the run.

mtjhrc: Yes the graphs and everything is nice! But makes we wonder if it is necessary to replicate the whole testing infrastructure here (but I guess we want that for other back-ends on Linux).
Anyways such improvements are out-of-scope for this PR, but they can be added later.

nirs: Network benchmarks for different programs and network proxies sounds like a separate project. This can be used by vfkit, krunkit, qemu, lima, minikube, vmnet-helper, gvproxy, passt and more.

slp (approved):

PR #605, opened 2026-03-24
libkrun: Add vhost-user sound device support

dorindabassey:
Add support for vhost-user sound devices following
QEMU's vhost-user-snd implementation. This allows
libkrun VMs to use external vhost-user sound backends
(e.g., vhost-device-sound) for audio playback and capture.

Note: The ALSA default device in the guest may not work without
additional configuration. Use explicit device specification
(e.g., 'aplay -D hw:0,0 <file>') or create /etc/asound.conf
in the guest to set the default device.

Depends on #527 and https://github.com/containers/libkrunfw/pull/114

### DIFF
--- a/src/devices/src/virtio/vhost_user/device.rs
+++ b/src/devices/src/virtio/vhost_user/device.rs
@@ -1,0 +1,548 @@
+// Copyright 2026, Red Hat Inc. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Generic vhost-user device wrapper.
+//!
+//! This module provides a wrapper around the vhost crate's Frontend,
+//! adapting it to work with libkrun's VirtioDevice trait.
+
+use std::io::{self, ErrorKind, Result as IoResult};
+use std::os::fd::AsRawFd;
+use std::os::unix::net::UnixStream;
+use std::sync::{Arc, Mutex};
+
+use log::{debug, error, warn};
+use polly::event_manager::{EventManager, Subscriber};
+use utils::epoll::{EpollEvent, EventSet};
+use utils::eventfd::{EventFd, EFD_NONBLOCK};
+use vhost::vhost_user::{Frontend, VhostUserFrontend, VhostUserProtocolFeatures};
+use vhost::{VhostBackend, VhostUserMemoryRegionInfo, VringConfigData};
+use vm_memory::{Address, GuestMemory, GuestMemoryMmap, GuestMemoryRegion};
+
+use crate::virtio::{
+    ActivateError, ActivateResult, DeviceQueue, DeviceState, InterruptTransport, QueueConfig,
+    VirtioDevice,
+};
+
+/// VHOST_USER_F_PROTOCOL_FEATURES (bit 30) is a backend-only feature
+/// that enables vhost-user protocol extensions. It's not a virtio feature.
+const VHOST_USER_F_PROTOCOL_FEATURES: u64 = 1 << 30;
+
+/// Generic vhost-user device wrapper.
+///
+/// This wraps a vhost-user backend connection and implements the VirtioDevice
+/// trait, allowing it to be used like any other virtio device in libkrun.
+pub struct VhostUserDevice {
+    /// Vhost-user frontend connection
+    frontend: Arc<Mutex<Frontend>>,
+
+    /// Device type (e.g., VIRTIO_ID_RNG = 4)
+    device_type: u32,
+
+    /// Device name for logging
+    device_name: String,
+
+    /// Queue configurations
+    queue_configs: Vec<QueueConfig>,
+
+    /// Available features from the backend
+    avail_features: u64,
+
+    /// Whether the backend supports protocol features
+    has_protocol_features: bool,
+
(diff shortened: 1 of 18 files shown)


src/devices/src/virtio/vhost_user/device.rs
```diff
@@ -154,6 +153,12 @@ impl VhostUserDevice {
             })
             .collect();
 
+        const VIRTIO_ID_SOUND: u32 = 25;
+        let config_size = match device_type {
+            VIRTIO_ID_SOUND => 16, // sizeof(struct virtio_snd_config)
+            _ => 256,
+        };
```
mtjhrc: This is quite unfortunate, if we need to know the size of the config, I think it should be configured the same way as number and sizes of virtqueues are, preferably specified in the C API, with a constant for the known value.

But if we keep adding more parameters to `krun_add_vhost_device()`, maybe it makes sense for that function to accept a whole `struct krun_vhost_config`. (and provide entire configured values of the `krun_vhost_config` struct for each device type).

mtjhrc (commented):
Please split the commit and improve the commit messages a bit, it's quite hard to follow. It goes unnecessarily into internals of the virtio-sound spec ("The virtio-snd configuration space is read-only and contains device capabilities (jacks, streams, chmaps, controls).", but doesn't mention actual implementation details (I am wondering why do we need `config_cache`? - or at least mention existence of it).

Maybe split it into like 3 commits: 
1. add config support for vhost user devices (this is mostly general, we need that for other vhost-user devices)
2. add the vhost-sound #define constants (and possibly other sound specific things) 
3. add support in the chroot_vm example (you can mention how to get `aplay` to work in that commit message or in a comment somewhere that is great!)

src/devices/src/virtio/vhost_user/device.rs
dorindabassey: humm, I don't think this is something that should be user-configurable, because what's the point if `config_size` is Fixed by the virtio spec, it's different from num_queues and queue_sizes. I'd leave the current implementation as is correct. We could add a documentation constant to be clear.

dorindabassey:
> Please split the commit and improve the commit messages a bit, it's quite hard to follow. 

Fixed, Thanks!

src/devices/src/virtio/vhost_user/device.rs
mtjhrc: > because what's the point if config_size is Fixed by the virtio spec

Main problem is that it can change with new features introduced by future protocol versions, so assuming the size based on device type is very brittle. One example is `VIRTIO_SND_F_CTLS` which adds another field to the struct for virtio-sound (introduced in virtio 1.3 spec). 

Correct me if I am wrong, but the only reason you need to know the size of the config space is the cache? In that case maybe we could avoid knowing the size of the config space and implement the caching like this:
```
guest requests 0..4 of config -> CACHE MISS: we read into 0..4 cache -> return from cache
guest requests 7..9 of config -> CACHE MISS: we read into 4..9 into cache -> return from cache
guest requests 4..8 of config -> CACHE HIT -> return from cache
```

basically you just have a `cache: Vec<u8>` and extend it whenever guest tries to read past the end of the cache. 

btw do other vmms (like QEMU) also cache the config space?

dorindabassey: no, QEMU doesn't cache, they re-fetch every time. But since the config space is small maybe caching makes sense.
I've changed it.

mtjhrc: To be honest that changes my view on this - we probably shouldn't cache either. Neither QEMU nor crosvm cache the config space - they re-fetch every time. 

Caching seems like more trouble than it's worth - we now have to worry about invalidation (and the current code doesn't seem to invalidate the cache at all). Simplest and trivially correct solution is just not to cache.

dorindabassey: yeah I agree. Fixed, I've removed the config space caching.

src/devices/src/virtio/vhost_user/device.rs
```diff
@@ -0,0 +1,479 @@
...
+                let offset_usize = offset as usize;
+                let end = offset_usize + data.len();
+                let mut config_buf = vec![0u8; end];
+
+                match frontend.get_config(
+                    0,
+                    end as u32,
+                    VhostUserConfigFlags::empty(),
+                    &mut config_buf,
+                ) {
```
mtjhrc: The intermediary buffer doesn't seem useful - is there a reason why we can't just do this?
```suggestion
                match frontend.get_config(
                    offset as u32,
                    data.len() as u32,
                    VhostUserConfigFlags::empty(),
                    data,
                ) {
```

Sidenote: The protocol for VHOST_USER_GET_CONFIG is weird - I can't fathom why are we even sending the buffer (the last parameter of frontend.get_config) attached to the msg... But that concern is beyond libkrun project, nothing we can do about it directly here.

dorindabassey: right! there's no need for the intermediary buffer.

mtjhrc:
Merged as part of #642

PR #610, opened 2026-03-27
virtiofs: split exit-code reporting and VM-exit request

yzewei:
Split guest exit-code reporting from VM-exit signaling in the virtio-fs path.

Guest exit status reporting and VM termination are different operations and
should not depend on reboot semantics.

Keep KRUN_EXIT_CODE_IOCTL for recording the guest exit status, and add
KRUN_EXIT_VM_IOCTL for explicitly requesting VM exit. Update init.krun to
call sync() and send the explicit VM-exit request instead of rebooting the
guest, and thread the exit request through the virtio-fs device/server/worker
path so the worker signals exit_evt only for the explicit VM-exit request.

This keeps the change focused on the virtio-fs exit protocol and avoids
mixing in unrelated init cleanups.

### DIFF
--- a/init/init.c
+++ b/init/init.c
@@ -14,6 +14,7 @@
 #include <net/if.h>
 #include <sys/ioctl.h>
 #include <sys/mount.h>
+#include <sys/reboot.h>
 #include <sys/resource.h>
 #include <sys/socket.h>
 #include <sys/stat.h>
@@ -32,6 +33,7 @@
 
 #define KRUN_EXIT_CODE_IOCTL 0x7602
 #define KRUN_REMOVE_ROOT_DIR_IOCTL 0x7603
+#define KRUN_EXIT_VM_IOCTL 0x7604
 
 #define KRUN_MAGIC "KRUN"
 #define KRUN_FOOTER_LEN 12
@@ -1003,6 +1005,41 @@ void set_exit_code(int code)
     close(fd);
 }
 
+static void request_vm_exit(void)
+{
+    int fd;
+    int ret;
+    int virtiofs_check;
+
+    virtiofs_check = is_virtiofs("/");
+    if (virtiofs_check < 0) {
+        printf("Warning: Could not determine filesystem type for root\n");
+    }
+
+    if (virtiofs_check == 0) {
+        return;
+    }
+
+    fd = open("/", O_RDONLY);
+    if (fd < 0) {
+        perror("Couldn't open root filesystem to request VM exit");
+        return;
+    }
+
+    ret = ioctl(fd, KRUN_EXIT_VM_IOCTL, 0);
+    if (ret < 0) {
+        perror("Error using the ioctl to request VM exit");
+    }
+
+    close(fd);
+}
+
+static void shutdown_vm(void)
+{
+    sync();
+    request_vm_exit();
+}
+
 int try_mount(const char *source, const char *target, const char *fstype,
               unsigned long mountflags, const void *data)
 {
@@ -1241,13 +1278,12 @@ int main(int argc, char **argv)
             // Not the first child, ignore it.
         };
 
-        // The workload's entrypoint has exited, record its exit code and exit
-        // ourselves.
         if (WIFEXITED(status)) {
             set_exit_code(WEXITSTATUS(status));
         } else if (WIFSIGNALED(status)) {
             set_exit_code(WTERMSIG(status) + 128);
         }
+        shutdown_vm();
     }
 
     return 0;
--- a/src/devices/src/virtio/fs/device.rs
+++ b/src/devices/src/virtio/fs/device.rs
@@ -2,7 +2,7 @@
 use crossbeam_channel::Sender;
 use std::cmp;
 use std::io::Write;
-use std::sync::atomic::{AtomicI32, AtomicU64, Ordering};
+use std::sync::atomic::{AtomicBool, AtomicI32, AtomicU64, Ordering};
 use std::sync::Arc;
 use std::thread::JoinHandle;
 
@@ -49,6 +49,8 @@ pub struct Fs {
     worker_thread: Option<JoinHandle<()>>,
     worker_stopfd: EventFd,
     exit_code: Arc<AtomicI32>,
+    exit_request: Arc<AtomicBool>,
+    exit_evt: EventFd,
     #[cfg(target_os = "macos")]
     map_sender: Option<Sender<WorkerMessage>>,
 }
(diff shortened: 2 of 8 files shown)


mtjhrc:
Hmm, I'm not sure about this. 

> When a virtual machine (VM) passes its final exit status via the KRUN_EXIT_CODE_IOCTL function on the virtio-fs root filesystem, the host records it in the shared exit_code AtomicI32, but does not signal it to the VMM's EventFd.

I'd say this was intended design, the ioctl wasn't meant as a general virtual machine shutdown mechanism, it is just a transport to extract the exit value from the VM. (@slp correct me if I am wrong)

Perhaps we could change it but it doesn't seem great, but I have no idea what is a good way to deal with this across all architectures (It would be nice to not have to bypass the kernel).

yzewei:
> Hmm, I'm not sure about this.
> 
> > When a virtual machine (VM) passes its final exit status via the KRUN_EXIT_CODE_IOCTL function on the virtio-fs root filesystem, the host records it in the shared exit_code AtomicI32, but does not signal it to the VMM's EventFd.
> 
> I'd say this was intended design, the ioctl wasn't meant as a general virtual machine shutdown mechanism, it is just a transport to extract the exit value from the VM. (@slp correct me if I am wrong)
> 
> Perhaps we could change it but it doesn't seem great, but I have no idea what is a good way to deal with this across all architectures (It would be nice to not have to bypass the kernel).

That's true. Could we retain ioctl as the transmission method, allowing the guest/initialization path to explicitly perform the actual shutdown after logging the exit code?

This way, the virtual machine manager can still exit via the existing exit event path.

yzewei:
I updated the commit description and completed the code modifications. PTAL!

init/init.c
```diff
@@ -1262,13 +1256,16 @@ int main(int argc, char **argv)
             // Not the first child, ignore it.
         };
 
-        // The workload's entrypoint has exited, record its exit code and exit
-        // ourselves.
+        // The workload's entrypoint has exited. Record its exit code before
+        // requesting a kernel-mediated shutdown so the VMM can later pick up
+        // the workload's final status when it handles the exit event.
+        printf("Before Recording workload exit code\n");
```
dorindabassey: This looks like debug code that should be removed, no?

init/init.c
```diff
@@ -1024,6 +1004,20 @@ void set_exit_code(int code)
...
+    sync();
+
+    /*
+     * Use a kernel-mediated reboot here rather than RB_POWER_OFF.
+     * On x86 guests, RB_POWER_OFF may fall back to halt, which does not
+     * trigger libkrun's VM exit path, while reboot/reset does.
+     */
+    if (reboot(RB_AUTOBOOT) < 0) {
+        printf("Error powering off the VM, trying to reboot instead\n");
+    }
```
dorindabassey: smalll nit: the error message is a bit confusing (says "trying to reboot instead" when it already called reboot) maybe better wording? `printf("Error: reboot() failed: %s\n", strerror(errno));`

dorindabassey (commented):
Everything else looks fine, just a few comments. Thanks!

init/init.c
yzewei: The printed information was not clear enough, I have revised it.

init/init.c
yzewei: Redundant debugging code has been removed.

yzewei:
@dorindabassey @mtjhrc I'm very sorry, but this solution has to be changed. I found that simply rebooting is unreliable when running it on a LoongArch machine.

Due to issues encountered during testing, the solution approach has changed.

It no longer emits the `exit_evt` signal from the guest exit code path. The decoupling is now in place:

- Guest exit code reporting

- Explicit virtual machine exit request

The reason is that relying on `reboot()` as the virtual machine exit mechanism is unreliable on the LoongArch architecture: the guest might get stuck in the restart path instead of exiting promptly.

The current model is:

- `KRUN_EXIT_CODE_IOCTL`: Only logs the exit status

- `KRUN_EXIT_VM_IOCTL`: Explicitly requests the virtual machine manager to exit

- `shutdown_vm()`: `sync()` + explicit virtual machine exit request, without using `reboot()`

Please submit an updated version for review.

dorindabassey (dismissed):
LGTM

slp:
@yzewei Could you please explain a bit more why relying on `reboot()` is unreliable on LoongArch?

yzewei:
> @yzewei Could you please explain a bit more why relying on `reboot()` is unreliable on LoongArch?

I'm so sorry, I just saw your question.  This is because the loongarch kvm kernel module does not provide a loongarch paravirt exit event (currently, KVM_HCALL_USER_SERVICE is thrown as KVM_EXIT_HYPERCALL by KVM), causing KVM to be unable to translate reboot into a vmm-usable exit event. The same applies to the guest kernel code (machine_restart/power_off currently does not actively trigger PV hypercalls). I will try to add this functionality to the loongarch kernel code, but it seems the current PR should also exist? At least it solves the semantic separation problem between exit_code (data) and vm_exit (control) and ensures that exiting can be completed correctly even for the loongarch architecture lacking the above logic?

slp:
> At least it solves the semantic separation problem between exit_code (data) and vm_exit (control) and ensures that exiting can be completed correctly even for the loongarch architecture lacking the above logic?

Not really. The current shut down flow goes like this:

1. Workloads entry point exits.
2. init.c notices the workload has exited, registers the workload's exit code using the virtio-fs ioctl, and exits itself.
3. kernel notices PID1 has exited and triggers an emergency (but otherwise ordered) shutdown.

IMO this is already a pretty natural shut down flow, involving all 3 components at the right pace.

yzewei:
> > At least it solves the semantic separation problem between exit_code (data) and vm_exit (control) and ensures that exiting can be completed correctly even for the loongarch architecture lacking the above logic?
> 
> Not really. The current shut down flow goes like this:
> 
> 1. Workloads entry point exits.
> 2. init.c notices the workload has exited, registers the workload's exit code using the virtio-fs ioctl, and exits itself.
> 3. kernel notices PID1 has exited and triggers an emergency (but otherwise ordered) shutdown.
> 
> IMO this is already a pretty natural shut down flow, involving all 3 components at the right pace.

Thank you for your clarification. I agree that the natural shutdown process (workload exits -> init logs exit code -> init exits -> kernel handles PID1 exit) is the correct direction.

Therefore, the explicit virtual machine exit path in this PR is inconsistent with this design, so I will close this PR.

For the LoongArch shutdown issue (host/KVM side), I will try to complete the feature completion on the kernel side.

PR #613, opened 2026-04-01
libkrun: vhost-user-vsock support

dorindabassey:
libkrun: Add API constants and example for vhost-user vsock

Add public API constants for vhost-user vsock devices and example
usage in chroot_vm. The underlying support already exists via the
generic VhostUserDevice wrapper.

Example integration:
- Added --vhost-user-vsock option to chroot_vm
- Calls krun_disable_implicit_vsock() to avoid conflict with
  built-in device
- Skips TSI port mapping when vhost-user-vsock is active

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -94,14 +94,14 @@ name = "arch"
 version = "0.1.0"
 dependencies = [
  "arch_gen",
- "kvm-bindings",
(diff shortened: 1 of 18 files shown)


mtjhrc:
Merged as part of #642

PR #616, opened 2026-04-01
init/aws-nitro: support partial read/write ethernet packet forwarding

jakecorrenti:
Ensure the requested number of bytes is forwarded to/from the host vsock providing the network access and the guest TAP device routing the application network traffic.

To write this code I used Cursor and the claude-4.6-opus-high model.

### Reproducer
To reproduce the issue, have the `examples/nitro.c` example execute the following:

```bash
TOKEN=$(curl -sf -X PUT 'http://169.254.169.254/latest/api/token' \
          -H 'X-aws-ec2-metadata-token-ttl-seconds: 21600')
echo "IMDSv2 token acquired"
ROLE=$(curl -sf 'http://169.254.169.254/latest/meta-data/iam/security-credentials/' \
              -H "X-aws-ec2-metadata-token: $TOKEN")
echo "--- IAM role: $ROLE ---"
curl -sf "http://169.254.169.254/latest/meta-data/iam/security-credentials/$ROLE" \
                -H "X-aws-ec2-metadata-token: $TOKEN" || { echo "error: failed to fetch credentials for role: $ROLE" >&2; exit 1; }
```

Without the above, about 90% of the time the final `curl` command will take about 15 seconds before failing because the client connection is closed. With the new diff, it should work 100% of the time.

### DIFF
--- a/init/aws-nitro/device/net_tap_afvsock.c
+++ b/init/aws-nitro/device/net_tap_afvsock.c
@@ -6,6 +6,7 @@
 #include <signal.h>
 #include <stdatomic.h>
 #include <stdbool.h>
+#include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
@@ -32,6 +33,58 @@
 #define TUN_DEV_MINOR 200
 
 /*
+ * The Extended Ethernet Frame header is 14 bytes, representing the Destination
+ * Address (6 bytes), Source Address (6 bytes) and the Ethertype (2 bytes).
+ */
+#define ETH_HEADER_LEN 14
+
+#define PROXY_HEADER_LEN 4
+
+/*
+ * Read exactly n bytes into the buffer, retrying on partial reads.
+ * Returns n on success, 0 on clean EOF, or -1 on error.
+ */
+static ssize_t read_exact(int fd, void *buf, size_t num_bytes)
+{
+    ssize_t total = 0;
+    ssize_t bytes_read;
+
+    while (total < num_bytes) {
+        bytes_read = read(fd, (char *)buf + total, num_bytes - total);
+        if (bytes_read < 0) {
+            if (errno == EINTR)
+                continue;
+            return -1;
+        }
+        if (bytes_read == 0)
+            return (total > 0) ? -1 : 0;
+        total += bytes_read;
+    }
+    return total;
+}
+
+/*
+ * Write exactly n bytes from the buffer to the fd, retrying on partial writes.
+ * Returns n on success, or -1 on error.
+ */
+static ssize_t write_all(int fd, const void *buf, size_t num_bytes)
+{
+    ssize_t total = 0;
+    ssize_t bytes_written;
+
+    while (total < num_bytes) {
+        bytes_written = write(fd, (const char *)buf + total, num_bytes - total);
+        if (bytes_written < 0) {
+            if (errno == EINTR)
+                continue;
+            return -1;
+        }
+        total += bytes_written;
+    }
+    return total;
+}
+
+/*
  * Forward ethernet packets to/from the host vsock providing network access and
  * the guest TAP device routing application network traffic.
  */
@@ -43,7 +96,7 @@ static int tap_vsock_forward(int tun_fd, int vsock_fd, int shutdown_fd,
     bool event_found;
     struct ifreq ifr;
     int ret, sock_fd;
-    unsigned int sz;
+    uint32_t sz;
     ssize_t nread;
 
     /*
@@ -53,7 +106,7 @@ static int tap_vsock_forward(int tun_fd, int vsock_fd, int shutdown_fd,
     sock_fd = socket(AF_INET, SOCK_DGRAM, 0);
     if (sock_fd < 0) {
         perror("creating INET socket to get TAP MTU");
-        return -errno;
+        return -1;
     }
 
     memset(&ifr, 0, sizeof(struct ifreq));
@@ -63,22 +116,29 @@ static int tap_vsock_forward(int tun_fd, int vsock_fd, int shutdown_fd,
     if (ret < 0) {
         close(sock_fd);
         perror("fetch MTU of TAP device");
-        exit(-errno);
+        return -1;
     }
 
     close(sock_fd);
 
-    buf = (unsigned char *)malloc(ifr.ifr_mtu);
+    uint32_t eth_frame_size = ifr.ifr_mtu + ETH_HEADER_LEN;
+    buf = (unsigned char *)malloc(eth_frame_size);
     if (buf == NULL) {
         perror("allocate buffer for TAP/vsock communication");
-        exit(-1);
+        return -1;
     }
 
-    // Forward the MTU to the host for it to allocate a corresponding buffer.
-    ret = write(vsock_fd, (void *)&ifr.ifr_mtu, sizeof(int));
-    if (ret < sizeof(int)) {
-        perror("write TAP device MTU to host");
-        exit(-errno);
+    // Forward the max ethernet frame size to the host for it to allocate a
+    // corresponding buffer.
+
+    // To avoid issues where the host endianness and the enclave endianness is
+    // different, convert to big endian to pass the max ethernet frame size to
+    // the host.
+    uint32_t eth_frame_size_be = htonl(eth_frame_size);
+    ret = write_all(vsock_fd, &eth_frame_size_be, sizeof(eth_frame_size));
+    if (ret < 0) {
+        perror("write max ethernet frame size to host");
+        goto cleanup;
     }
 
     pfds[0].fd = vsock_fd;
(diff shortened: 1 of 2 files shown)


jakecorrenti:
cc @tylerfanelli

init/aws-nitro/device/net_tap_afvsock.c
```diff
@@ -97,14 +139,17 @@ static int tap_vsock_forward(int tun_fd, int vsock_fd, int shutdown_fd,
...
 
             unsigned int len = htonl(sz);
 
-            nread = read(vsock_fd, buf, len);
-            write(tun_fd, buf, nread);
+            nread = read_exact(vsock_fd, buf, len);
+            if (nread <= 0)
+                exit(0);
+
+            write_all(tun_fd, buf, nread);
```
tylerfanelli: Should we check for `-1` here? Along with all `write_all` calls?

init/aws-nitro/device/net_tap_afvsock.c
```diff
@@ -97,14 +139,17 @@ static int tap_vsock_forward(int tun_fd, int vsock_fd, int shutdown_fd,
         event_found = false;
         // Event on vsock. Read the frame and write it to the TAP device.
         if (pfds[0].revents & POLLIN) {
-            nread = read(vsock_fd, &sz, 4);
-            if (nread != 4)
+            nread = read_exact(vsock_fd, &sz, 4);
+            if (nread <= 0)
```
tylerfanelli: The only situation where I think `nread` will be zero is if we `read_exact` with 0 as input, which I don't think would be a valid call. Should we instead check for:

```
if (nread < 0)
```

Instead?

jakecorrenti: yes most likely. I kept this a draft because I haven't gone through the code with a fine-tooth comb the agent wrote yet, I only verified it worked. 

everything still needs to get refactored :)

tylerfanelli: No worries, mark as ready when you see fit.

init/aws-nitro/device/net_tap_afvsock.c
```diff
@@ -97,25 +150,49 @@ static int tap_vsock_forward(int tun_fd, int vsock_fd, int shutdown_fd,
         event_found = false;
         // Event on vsock. Read the frame and write it to the TAP device.
         if (pfds[0].revents & POLLIN) {
-            nread = read(vsock_fd, &sz, 4);
-            if (nread != 4)
+            nread = read_exact(vsock_fd, &sz, 4);
```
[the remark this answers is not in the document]
jakecorrenti: ~@tylerfanelli where did you get the # 4 here from? Is it documented somewhere in the kernel?~ Found it in virtio/net/unixstream.rs

init/aws-nitro/device/net_tap_afvsock.c
```diff
@@ -97,25 +150,49 @@ static int tap_vsock_forward(int tun_fd, int vsock_fd, int shutdown_fd,
...
 
             event_found = true;
         }
 
         // Event on the TAP device. Read the frame and write it to the vsock.
         if (pfds[1].revents & POLLIN) {
-            nread = read(tun_fd, buf, ifr.ifr_mtu);
+            nread = read(tun_fd, buf, buffer_size);
             if (nread > 0) {
                 sz = htonl(nread);
```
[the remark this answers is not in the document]
jakecorrenti: I think reading exactly one ethernet frame makes sense here

init/aws-nitro/device/net_tap_afvsock.c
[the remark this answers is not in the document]
jakecorrenti: I'm going to keep this as is for the time being and it can be fixed in another PR to clean up

init/aws-nitro/device/net_tap_afvsock.c
```diff
@@ -31,6 +32,64 @@
...
+    while (total < n) {
+        ssize_t w = write(fd, (const char *)buf + total, n - total);
+        if (w <= 0) {
+            if (w < 0 && errno == EINTR)
+                continue;
+
+            if (w == 0)
+                errno = EIO;
+
+            return -1;
```
tylerfanelli: Not sure if exactly correct, but trying to avoid nested ifs.

```suggestion
        if (w < 0 && errno != EINTR)
             return -1;
        else if (w == 0) {
            errno = EIO
            return -1;
        }
```

jakecorrenti: No that isn't logically the same because if `w < 0 && errno == EINTR` is true, then it falls through and decreases the `total` value which would be incorrect since we didn't actually write any data nor do we want to say we wrote less than we actually did.

jakecorrenti:
@tylerfanelli let me know if you think the latest revision is more readable -- failures don't have to do with this PR

init/aws-nitro/device/net_tap_afvsock.c
```diff
@@ -31,6 +32,64 @@
...
+{
+    size_t total = 0;
+
+    while (total < n) {
+        ssize_t w = write(fd, (const char *)buf + total, n - total);
+        if (w < 0 && errno == EINTR)
+            continue;
+
+        if (w == 0)
+            errno = EIO;
```
tylerfanelli: What's the purpose of setting errno if you never return an error? Is this used on another iteration?

init/aws-nitro/device/net_tap_afvsock.c
```diff
@@ -31,6 +32,64 @@
...
+/*
+ * Read exactly n bytes into the buffer, retrying on partial reads.
+ * Returns n on success, 0 on clean EOF, or -1 on error.
+ */
+static ssize_t read_exact(int fd, void *buf, size_t n)
+{
+    size_t total = 0;
+
+    while (total < n) {
+        ssize_t r = read(fd, (char *)buf + total, n - total);
```
tylerfanelli: `r` should be declared at the top of the function. Also, maybe something like `tot_read` would be a better name.

init/aws-nitro/device/net_tap_afvsock.c
```diff
@@ -31,6 +32,64 @@
...
+        if (r < 0) {
+            if (errno == EINTR)
+                continue;
+            return -1;
```
tylerfanelli: Try:

```suggestion
        if (r < 0 && errno == EINTR) {
            continue;
```

And check the standard `r < 0` as the final `else if`.

init/aws-nitro/device/net_tap_afvsock.c
```diff
@@ -31,6 +32,64 @@
...
+        } else if (r == 0) {
+            if (total > 0) {
+                errno = EIO;
+                return -1;
+            }
+            return 0;
```
tylerfanelli: Why return 0? Shouldn't `total` be returned even if it's less than `n`?

init/aws-nitro/device/net_tap_afvsock.c
```diff
@@ -31,6 +32,64 @@
...
+/*
+ * Write exactly n bytes from the buffer to the fd, retrying on partial writes.
+ * Returns n on success, or -1 on error.
+ */
+static ssize_t write_all(int fd, const void *buf, size_t n)
+{
+    size_t total = 0;
+
+    while (total < n) {
+        ssize_t w = write(fd, (const char *)buf + total, n - total);
```
tylerfanelli: w should be declared at the top of the function. Also, maybe something like `tot_written` would be a better name.

init/aws-nitro/device/net_tap_afvsock.c
```diff
@@ -31,6 +32,64 @@
...
+        if (w < 0 && errno == EINTR)
+            continue;
+
+        if (w == 0)
+            errno = EIO;
+
+        if (w <= 0)
+            return -1;
```
tylerfanelli: You already check if `w == 0`, so this can be just a check for `w < 0`.

```suggestion
        if (w < 0)
            return -1;
```

tylerfanelli (changes requested):

init/aws-nitro/device/net_tap_afvsock.c
jakecorrenti: because if we got to this point then the bytes we've read is 0 and `total` is also 0. We could return `total` here but returning 0 is explicit and easy to read IMO.

init/aws-nitro/device/net_tap_afvsock.c
jakecorrenti: Because we only want `errno` to be set if we wrote 0 bytes. We then return -1 in the `if` statement immediately below to prevent a nested `if` where we check `w <= 0` and then inside that block also check `w == 0` to set errno

init/aws-nitro/device/net_tap_afvsock.c
jakecorrenti: see comment above.

init/aws-nitro/device/net_tap_afvsock.c
tylerfanelli: But you're explicitly setting errno, when it usually should be set by the call to `write`. Usually the caller would not modify errno.

jakecorrenti:
@tylerfanelli PTAL when you get the chance. IMO in its current state the nested ifs in `{read, write}_exact()` are preferred here so that we don't duplicate the `if (bytes < 0)` conditional.

tylerfanelli (approved):
LGTM!

PR #619, opened 2026-04-03
tests: create /dev /proc /sys in temporary rootfs before boot

yzewei:
Create mountpoint directories in test rootfs setup so early init mounts (devtmpfs/proc/sysfs) do not fail with ENOENT.

This avoids boot-time errors such as:
`devtmpfs: error mounting -2`.

### DIFF
--- a/src/vmm/src/vmm_config/kernel_cmdline.rs
+++ b/src/vmm/src/vmm_config/kernel_cmdline.rs
@@ -5,10 +5,10 @@ use std::fmt::{Display, Formatter, Result};
 
 #[cfg(target_os = "linux")]
 pub const DEFAULT_KERNEL_CMDLINE: &str = "reboot=k panic=-1 panic_print=0 nomodule console=hvc0 \
-                                          rootfstype=virtiofs rw quiet no-kvmapf";
+                                          root=/dev/root rootfstype=virtiofs rw quiet no-kvmapf";
 #[cfg(target_os = "macos")]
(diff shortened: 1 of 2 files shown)


src/vmm/src/vmm_config/kernel_cmdline.rs
```diff
@@ -5,10 +5,10 @@ use std::fmt::{Display, Formatter, Result};
 
 #[cfg(target_os = "linux")]
 pub const DEFAULT_KERNEL_CMDLINE: &str = "reboot=k panic=-1 panic_print=0 nomodule console=hvc0 \
-                                          rootfstype=virtiofs rw quiet no-kvmapf";
```
mtjhrc, line 8: This change isn't mentioned in the commit which should only touch tests, this seems to be accidentally committed.

mtjhrc:
Isn't this already handled by init?

https://github.com/containers/libkrun/blob/7c5292cfddd207940458958befaa607b301f4fbe/init/init.c#L395-L406

yzewei:
> Isn't this already handled by init?
> 
> https://github.com/containers/libkrun/blob/7c5292cfddd207940458958befaa607b301f4fbe/init/init.c#L395-L406

Thanks for the review.

I reverted the accidental kernel cmdline change and re-ran the test from a clean state.
I could not reproduce the original `virtio-fs: tag <> not found` / `devtmpfs: error mounting -2` issue.


Given I can't reproduce the reported behavior anymore, I'll close this PR for now.
If I hit it again, I'll reopen with a minimal reproducer and logs.

PR #622, opened 2026-04-08
vmm: Fix bzip2 magic header detection

d-e-s-o:
The bzip2 magic byte search uses `.windows(4)` to produce 4-byte slices, but compares them against the 3-byte pattern `[b'B', b'Z', b'h']`. As a result, we will never see a match. Thus, the bzip2 header is never found, and loading always fails with ImageBz2Invalid (surfacing as EINVAL to callers).
Fix the issue by using `.windows(3)`, matching how the `gzip` path handles its 3-byte magic.


slp (approved):
LGTM, thanks!

PR #623, opened 2026-04-09
virtiofs: support read-only mounts

JAORMX:
This PR adds a read-only virtio-fs mode to the public API and wires the new flag through the virtio-fs device configuration, worker setup, and VMM builder so callers can expose shared directories without allowing guest writes.

The implementation wraps the passthrough backend in a read-only filesystem layer that rejects mutating FUSE operations with `EROFS`, disables writeback cache, and still preserves the non-mutating control ioctls the guest uses. On macOS, `setupmapping` now opens the backing file with `O_RDONLY` for read-only mappings, which fixes read-only DAX mounts.

I also changed virtio-fs worker creation to return an activation error instead of panicking if the backend cannot be initialized.

### DIFF
--- a/include/libkrun.h
+++ b/include/libkrun.h
@@ -323,6 +323,26 @@ int32_t krun_add_virtiofs2(uint32_t ctx_id,
                            const char *c_path,
                            uint64_t shm_size);
 
+/**
+ * Adds an independent virtio-fs device pointing to a host's directory with a tag. This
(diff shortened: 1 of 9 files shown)


slp:
Tested RW implementation and found no regressions. Code also LGTM. Let's see Gemini's opinion.

slp (approved):

slp:
Thanks for the PR! Let's get it merged.

PR #628, opened 2026-04-13
libkrun: vhost-user-console support

dorindabassey:
libkrun: Add API constants and example for vhost-user console

Add public API constants for vhost-user console devices and example
usage in chroot_vm. The underlying support already exists via the
generic VhostUserDevice wrapper.

Console devices require 4 queues for multiport support (receiveq,
transmitq, control receiveq, control transmitq).

Example integration:
- Added --vhost-user-console option to chroot_vm
- Available as /dev/hvc1 in the guest (implicit console uses hvc0)
- Can be used with vhost-device-console for remote console access

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -94,14 +94,14 @@ name = "arch"
 version = "0.1.0"
 dependencies = [
  "arch_gen",
- "kvm-bindings",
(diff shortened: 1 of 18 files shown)


mtjhrc:
Merged as part of #642

PR #630, opened 2026-04-13
Implement support for nested virt on Linux

### LINKED ISSUE #625: nested virt on linux
It seems as if Nested virt for MacOS was implemented (https://github.com/containers/libkrun/pull/288), but nested virt appears [absent on Linux](https://github.com/containers/libkrun/blob/44312fdfa105256f1988ea8cdfbde8ed1532c944/include/libkrun.h#L971).

The use case here is being able to have `/dev/kvm` inside your containers started with say `podman run --runtime=krun`.

slp:
So far, we supported enabling nested virt on macOS but not on Linux. Let's do that now.

Fixes: #625


jakecorrenti (approved):
LGTM!

PR #631, opened 2026-04-13
devices/mmio: Read QueueSizeMax from queue_config instead of actual queues

### LINKED ISSUE #626: macOS host: virtio-net connection broken in FreeBSD guest
Since commit c87386c, network doesn't work in my FreeBSD microVM.

Steps to reproduce:

1. Install anylinuxfs

```
brew tap nohajc/anylinuxfs
brew install anylinuxfs
```

2. Copy the binaries somewhere else and download entitlements plist

```
cp -r /opt/homebrew/opt/anylinuxfs/ somewhere-else
curl -LO https://raw.githubusercontent.com/nohajc/anylinuxfs/refs/heads/main/anylinuxfs.entitlements
```

3. Patch (and re-sign) the `anylinuxfs` binary to link against libkrun built from main.

```
cd somewhere-else
install_name_tool -change /opt/homebrew/opt/libkrun/lib/libkrun.1.dylib /usr/local/lib/libkrun.1.dylib bin/anylinuxfs
codesign --entitlements "anylinuxfs.entitlements" --force -s - bin/anylinuxfs
```

4. Start the microVM (this will automatically bootstrap an OS image) and try internet connectivity

```
bin/anylinuxfs shell -i freebsd
/init-network.sh && fetch -qo - http://ifconfig.co
```

Normally, you should get your public IP from the web service. With the offending commit, I get this
```
fetch: http://ifconfig.co: Transient resolver failure
```

Network init script is very simple:

```
#!/bin/sh

ifconfig vtnet0 inet 192.168.127.2/24
route add default 192.168.127.1
```

DNS configuration is already in `/etc/resolv.conf`

```
nameserver 192.168.127.1
```

The IPs are hardcoded for gvproxy which handles the network on the host side.

Note that if this is released to homebrew, it will break my setup for mounting ZFS drives on macOS. The same issue prevents me from setting up [gvproxy port-forwarding](https://github.com/containers/gvisor-tap-vsock?tab=readme-ov-file#port-forwarding) from the guest side (i.e. sending a POST request to `http://192.168.127.1/services/forwarder/expose`).

mtjhrc:
Commit c87386cc refactored queue ownership so transport queues are
moved to the device backend on activation (self.queues = None).
This caused QueueSizeMax to return 0 if device state prevented
recreating the queues (e.g. failed reset), making FreeBSD give up.
 
This change restores the previous behavior where QueueSizeMax can always be reported regardless of device status. 

Note: there is still an underlying problem - the lack of proper device reset support in the net device which would have probably fixed this too (but we need to implement reset for every device).

Fixes: #626

### DIFF
--- a/src/devices/src/virtio/mmio.rs
+++ b/src/devices/src/virtio/mmio.rs
@@ -391,7 +391,10 @@ impl BusDevice for MmioTransport {
                         }
                         features
                     }
-                    0x34 => self.with_queue(0, |q| u32::from(q.get_max_size())),
+                    0x34 => self
+                        .queue_config
+                        .get(self.queue_select as usize)
+                        .map_or(0, |c| c.size as u32),
                     0x44 => self.with_queue(0, |q| q.ready as u32),
                     0x60 => self.interrupt.status().load(Ordering::SeqCst) as u32,
                     0x70 => self.device_status,


nohajc:
Thanks for the fix!

slp:
> Note: the underlying problem is still the lack of proper device reset support in the net backend — FreeBSD guests hit a permanent FAILED state.

I'm confused by this part of the commit message. Does the network work on FreeBSD with this fix?

mtjhrc:
> I'm confused by this part of the commit message. Does the network work on FreeBSD with this fix?

Yes, but the device state stays FAILED.

nohajc:
> This change restores the previous behavior

So, the FAILED state issue was always present?

slp:
> > I'm confused by this part of the commit message. Does the network work on FreeBSD with this fix?
> 
> Yes, but the device state stays FAILED.

Gotcha. Let's merge this one for 1.x and we'll fix it properly in 2.x.

slp (approved):
LGTM, thanks!

PR #636, opened 2026-04-14
Add GitHub workflow for cross-compilation and fix build issues for macOS and FreeBSD

nohajc:
I figured it would be a good idea to run automated builds for our cross-compilation scenarios. So the new pull request workflow tests the following combinations:

1. macOS host, Linux init
2. macOS host, FreeBSD init
3. Linux host, FreeBSD init

Of course I immediately found out my newly merged FreeBSD init port didn't compile, so the fix is included (it was just a broken Makefile rule).

I also tried to fix various issues with building unit tests on macOS which popped up in my VS Code. Most of the changes were suggested by AI with the following explanation:

```
test_get_supported_cpuid — x86_64-specific, removed entirely
Vm::supported_cpuid() is a KVM concept (fetching the CPUID leaf table the hypervisor supports). HVF on ARM has no such thing. The method doesn't exist on the macOS Vm struct.

test_setup_irqchip — KVM concept, removed entirely
The KVM Vmm had an explicit setup_irqchip() call to configure the in-kernel interrupt controller before vCPUs could be created. HVF doesn't have this: the GIC is set up differently and is not part of Vm. The method simply doesn't exist here.

test_kvm_context — KVM concept, removed entirely
KvmContext is a wrapper around the /dev/kvm file descriptor. HVF has no equivalent — Vm::new(nested_enabled) talks to Hypervisor.framework directly via dlopen. There's nothing to test here.

test_vcpu_kick + test_vcpu_rtsig_offset — signal-based mechanism, removed entirely
On Linux/KVM, to interrupt a vCPU that's running inside KVM_RUN, a real-time signal (SIGRTMIN + offset) is sent to the vCPU thread, and the KVM signal handler sets kvm_run.immediate_exit = 1. The test verified this entire signal delivery path by:

Spawning a thread, starting it in a KVM inner loop
Sending the RTSIG from the outer thread
Checking that kvm_run.immediate_exit was set
On macOS/HVF the entire mechanism is stubbed out — register_kick_signal_handler() is an empty function, KvmRunWrapper doesn't exist on HVF, and vcpu.fd doesn't exist. Interrupting an HVF vCPU is done through hvf::vcpu_request_exit() instead.

test_vm_save_restore_state / test_vcpu_save_restore_state — x86_64 only, gated already
These were already #[cfg(target_arch = "x86_64")] and test KVM save/restore state (PIT, IOAPIC, MSRs, CPUID). None of that exists on the ARM64/HVF path so they just couldn't compile.

What changed inside test_vcpu_tls
The original test had three extra assertions using a run_on_thread_local() method that doesn't exist in this codebase's Vcpu:

    assert!(Vcpu::run_on_thread_local(|_| ()).is_err());
run_on_thread_local was a KVM-era helper that accepted a closure and ran it against the TLS-stored Vcpu pointer. It was never ported to the HVF Vcpu. The simplified version keeps the meaningful assertions — that you can't double-init TLS, that you can reset it, and that you can't reset it twice — while using the methods that actually exist (init_thread_local_data / reset_thread_local_data).
```

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,2 @@
+[target.aarch64-apple-darwin]
+runner = ".cargo/macos-test-runner.sh"
--- a/.cargo/macos-test-runner.sh
+++ b/.cargo/macos-test-runner.sh
@@ -1,0 +1,14 @@
+#!/bin/sh
+# Cargo target runner for macOS (aarch64-apple-darwin).
+# Codesigns the test/run binary with the Hypervisor.framework entitlement
+# before executing it, so that HVF-based tests can run without a developer
+# certificate.
+set -eu
+
+BINARY="$1"
+shift
+
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+REPO_ROOT="$(cd "$SCRIPT_DIR/.." && pwd)"
+codesign --entitlements "$REPO_ROOT/hvf-entitlements.plist" --force -s - "$BINARY"
+exec "$BINARY" "$@"
--- a/.github/workflows/cross-compilation.yml
+++ b/.github/workflows/cross-compilation.yml
@@ -1,0 +1,58 @@
+name: Cross-compilation
+on: [pull_request]
+
+jobs:
+  cross-compile-macos:
+    name: Cross-compilation (macOS)
+    runs-on: macos-latest
+    env:
+      DYLD_LIBRARY_PATH: /Library/Developer/CommandLineTools/usr/lib/
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Setup build environment
+        uses: ./.github/actions/setup-build-env
+
+      - name: Install cross-compilation dependencies
+        run: |
+          brew install lld xz
+
+      - name: Build libkrun on macOS (with Linux init)
+        run: make
+
+      - name: Build FreeBSD init on macOS
+        run: make BUILD_BSD_INIT=1 -- init/init-freebsd
+
+  cross-compile-linux:
+    name: Cross-compilation (Linux x86_64 -> FreeBSD)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Setup build environment
+        uses: ./.github/actions/setup-build-env
+
+      - name: Install cross-compilation dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y --no-install-recommends lld
+
+      - name: Build FreeBSD init on Linux
+        run: make BUILD_BSD_INIT=1 -- init/init-freebsd
+
+  cross-compile-linux-aarch64:
+    name: Cross-compilation (Linux aarch64 -> FreeBSD)
+    runs-on: ubuntu-24.04-arm
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Setup build environment
+        uses: ./.github/actions/setup-build-env
+
+      - name: Install cross-compilation dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y --no-install-recommends lld
+
+      - name: Build FreeBSD init on Linux aarch64
+        run: make BUILD_BSD_INIT=1 -- init/init-freebsd
--- a/Makefile
+++ b/Makefile
@@ -120,7 +120,7 @@ $(AWS_NITRO_INIT_BINARY): $(AWS_NITRO_INIT_SRC)
 	$(CC) -O2 -static -s -Wall $(AWS_NITRO_INIT_LD_FLAGS) -o $@ $(AWS_NITRO_INIT_SRC) $(AWS_NITRO_INIT_LD_FLAGS)
 
 ifeq ($(OS),Darwin)
-# If SYSROOT_BSD is not set and we're on macOS, generate sysroot automatically
+# macOS -> FreeBSD cross-compilation
 ifeq ($(SYSROOT_BSD),)
     SYSROOT_BSD = $(FREEBSD_ROOTFS_DIR)
     SYSROOT_BSD_TARGET = $(FREEBSD_ROOTFS_DIR)/.sysroot_ready
@@ -129,6 +129,16 @@ else
 endif
     # Cross-compile on macOS with the LLVM linker (brew install lld)
     CC_BSD=$(CLANG) -target $(ARCH)-unknown-freebsd -fuse-ld=lld -stdlib=libc++ -Wl,-strip-debug --sysroot $(SYSROOT_BSD)
+else ifeq ($(OS),Linux)
+# Linux -> FreeBSD cross-compilation
+ifeq ($(SYSROOT_BSD),)
+    SYSROOT_BSD = $(FREEBSD_ROOTFS_DIR)
+    SYSROOT_BSD_TARGET = $(FREEBSD_ROOTFS_DIR)/.sysroot_ready
+else
+    SYSROOT_BSD_TARGET =
+endif
+    # Cross-compile on Linux with clang
+    CC_BSD=$(CLANG) -target $(ARCH)-unknown-freebsd -fuse-ld=lld -Wl,-strip-debug --sysroot $(SYSROOT_BSD)
 else
     # Build on FreeBSD host
     CC_BSD=$(CC)
--- a/Makefile
+++ b/Makefile
@@ -150,7 +150,7 @@ endif
 ifeq ($(BUILD_BSD_INIT),1)
 INIT_BINARY_BSD = init/init-freebsd
 $(INIT_BINARY_BSD): $(INIT_SRC) $(SYSROOT_BSD_TARGET)
-	$(CC_BSD) -std=c23 -O2 -static -Wall -lutil -o $@ $(INIT_SRC)
+	$(CC_BSD) -std=c23 -O2 -static -Wall -o $@ $(INIT_SRC) -lutil
 endif
 
 # Sysroot preparation rules for cross-compilation on macOS
--- a/Makefile
+++ b/Makefile
@@ -20,6 +20,8 @@ AWS_NITRO_INIT_SRC = \
 
 AWS_NITRO_INIT_LD_FLAGS = -larchive -lnsm
 
+INIT_SRC = init/init.c
+
 ifeq ($(SEV),1)
     VARIANT = -sev
     FEATURE_FLAGS := --features amd-sev
@@ -148,7 +150,7 @@ endif
 ifeq ($(BUILD_BSD_INIT),1)
 INIT_BINARY_BSD = init/init-freebsd
 $(INIT_BINARY_BSD): $(INIT_SRC) $(SYSROOT_BSD_TARGET)
-	$(CC_BSD) -std=c23 -O2 -static -Wall $(INIT_DEFS) -lutil -o $@ $(INIT_SRC) $(INIT_DEFS)
+	$(CC_BSD) -std=c23 -O2 -static -Wall -lutil -o $@ $(INIT_SRC)
 endif
 
 # Sysroot preparation rules for cross-compilation on macOS
@@ -189,10 +191,12 @@ $(FREEBSD_ROOTFS_DIR)/.sysroot_ready: $(FREEBSD_BASE_TXZ)
 	@cd $(FREEBSD_ROOTFS_DIR) && tar xJf base.txz 2>/dev/null || true
 	@touch $@
 
+BSD_ARCH=$(subst x86_64,amd64,$(subst aarch64,arm64,$(ARCH)))
+
 $(FREEBSD_BASE_TXZ):
-	@echo "Downloading FreeBSD $(FREEBSD_VERSION) base for $(ARCH)..."
+	@echo "Downloading FreeBSD $(FREEBSD_VERSION) base for $(BSD_ARCH)..."
 	@mkdir -p $(FREEBSD_ROOTFS_DIR)
-	@curl -fL -o $@ https://download.freebsd.org/releases/$(ARCH)/$(FREEBSD_VERSION)/base.txz
+	@curl -fL -o $@ https://download.freebsd.org/releases/$(BSD_ARCH)/$(FREEBSD_VERSION)/base.txz
 
 clean-sysroot:
 	rm -rf $(ROOTFS_DIR)
--- a/src/libkrun/build.rs
+++ b/src/libkrun/build.rs
@@ -10,6 +10,4 @@ fn main() {
         std::env::var("CARGO_PKG_VERSION_MAJOR").unwrap(), std::env::var("CARGO_PKG_VERSION_MAJOR").unwrap(),
         std::env::var("CARGO_PKG_VERSION_MAJOR").unwrap(), std::env::var("CARGO_PKG_VERSION_MINOR").unwrap()
     );
-    #[cfg(target_os = "macos")]
-    println!("cargo:rustc-link-lib=framework=Hypervisor");
 }
--- a/src/vmm/build.rs
+++ b/src/vmm/build.rs
@@ -9,4 +9,8 @@ fn main() {
         println!("cargo:rustc-env=KRUN_EDK2_BINARY_PATH={edk2_binary_path}");
         println!("cargo:rerun-if-env-changed=KRUN_EDK2_BINARY_PATH");
     }
+
+    if std::env::var("CARGO_CFG_TARGET_OS").as_deref() == Ok("macos") {
+        println!("cargo:rustc-link-lib=framework=Hypervisor");
+    }
 }
--- a/src/vmm/src/macos/vstate.rs
+++ b/src/vmm/src/macos/vstate.rs
@@ -10,7 +10,6 @@ use std::collections::HashMap;
 use std::fmt::{Display, Formatter};
 use std::io;
 use std::result;
-#[cfg(not(test))]
 use std::sync::Arc;
 use std::thread;
 use std::time::Duration;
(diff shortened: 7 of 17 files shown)


.cargo/config.toml
nohajc, line 1: The `runner` configuration lets me run unit tests from VS Code editor.

.cargo/macos-test-runner.sh
```diff
@@ -0,0 +1,26 @@
...
+cat > "$ENTITLEMENTS" << 'EOF'
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+    <key>com.apple.security.hypervisor</key>
+    <true/>
+</dict>
+</plist>
+EOF
```
mtjhrc: The `tests/run.sh` already codesigns the executable. It builds it first (not sure if that is a problem for you) and runs it, works on both Linux and macOS hosts. 
I basically always run the tests via `make test` personally.

I'm fine with making this more flexible - can we somehow unify the workflow instead of adding the code signing in a different shell-script? 

You still need the run.sh to build the guest side of the tests, so this alone only codesigns the host part of the tests (presumably after being rebuilt through cargo?) but the guest part stays stale? Doesn't seem great...

nohajc: I'll think about how to unify these. The thing is, when I open the project in VS Code on macOS, the Rust IDE extension calls `cargo check` automatically and `cargo test` through the buttons generated for each test function. So it needs to be somehow transparent.

Anyway, it's not so important, just a quality of life improvement. We can even split the PR, I mostly care about the first two commits now. The editor / test build tweaks can be solved separately.

nohajc: I should also mention I was only trying to make unit tests work. Not the whole test suite. So running cargo tests selectively is not the same as `make test` I believe. The test-runner is only meant for the former.

mtjhrc: Are you talking about the libkrun project/workspace (unit tests) or tests workspace (tests subdirectory) though? 

Btw not sure about VS Code, but for me on Zed (which also uses rust-analyzer) I have to open the tests separately otherwise I don't get code completions (they are a separate workspace), I can enable both of them, but then it gets confused about the features...

mtjhrc: > I should also mention I was only trying to make unit tests work

Ahh, makes sense then, this can stay as is then.

nohajc: Yeah, I just answered your question at the same time. 😀

I only wanted rust-analyzer in VS Code to stop shouting at me. Interestingly, some of the unit tests seemed broken (copy pasted from the Linux version).

.github/workflows/cross-compilation.yml
```diff
@@ -0,0 +1,41 @@
+name: Cross-compilation
+on: [pull_request]
+
+jobs:
+  cross-compile-macos:
+    name: Cross-compilation (macOS)
+    runs-on: macos-latest
+    env:
+      DYLD_LIBRARY_PATH: /Library/Developer/CommandLineTools/usr/lib/
```
nohajc, line 9: @slp I had issues building the macOS version which looked like this:
https://github.com/slp/homebrew-krun/issues/14

Setting `DYLD_LIBRARY_PATH` fixed it for the GitHub action. Maybe the `krun_display` patch for homebrew is not needed.

.cargo/macos-test-runner.sh
```diff
@@ -0,0 +1,26 @@
...
+
+codesign --entitlements "$ENTITLEMENTS" --force -s - "$BINARY"
```
mtjhrc: You don't need the temp file, you can just do:
```suggestion
codesign --entitlements /dev/stdin --force -s - "$BINARY" << 'EOF'
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
    <key>com.apple.security.hypervisor</key>
    <true/>
</dict>
</plist>
EOF
```

nirs: Why the plist is not in the source?

.cargo/macos-test-runner.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+# Cargo target runner for macOS (aarch64-apple-darwin).
+# Codesigns the test/run binary with the Hypervisor.framework entitlement
+# before executing it, so that HVF-based tests can run without a developer
+# certificate.
```
nirs, line 5: How does it work locally on a developer machine?

codesign should be part of the build on macOS, no extra code should be needed to run tests locally on in CI.

nohajc: You always need to sign the executable that links to libkrun. Signing the library itself is not good enough. When you set this script as `runner` in `.cargo/config.toml`, `cargo test` will automatically use it as a wrapper before running your unit tests. There is at least one test which needs the Hypervisor entitlement.

Since `rust-analyzer` IDE extension expects to be able to call `cargo test` directly, this is the way to do it.

nirs: So this is CI only build step?

.cargo/macos-test-runner.sh
nohajc: @nirs There is already https://github.com/containers/libkrun/blob/main/tests/run.sh#L50 and a similar plist here: https://github.com/containers/libkrun/blob/main/examples/chroot_vm.entitlements

Could be reused. The established pattern was to have one for each executable artefact, it seems.

.cargo/macos-test-runner.sh
nohajc: Sorry for the confusion. Not exactly.
- The first commit is CI related, introducing a GitHub action that verifies cross-compilation. Doesn't run any tests.
- Second commit fixes issues with Makefile that I found when testing the CI steps.
- Third commit is a bit unrelated. It fixes issues which become apparent when you open the libkrun workspace in VS Code on macOS. I was unable to build and run unit tests from the IDE. Some of them were outdated and the codesigning for `cargo test` didn't work.

nohajc: @nirs I think the bottom line is `make test` worked but `cargo test` (just unit tests, not integration) didn't. But my new CI workflow is build only.

nirs: I mean the runner that need codesign - if this is needed both locally and in CI, it is better o have the entitlement.plist in the source and have trivial signing code that runs in make test.

nohajc: Sure. `make test` and `cargo test` each sign a different executable but the entitlement is always the same.

@mtjhrc I think the file is better than repeating the same heredoc.

mtjhrc: > I think the file is better than repeating the same heredoc

Makes sense, inline strings would make sense if they were different, but practically only thing we need is `com.apple.security.hypervisor`

nohajc: Done.

mtjhrc:
Please try to make sure the commit messages do what the title and commit msg text says.

Like `github actions: add cross-compilation workflow` commit also introduces `Linux -> FreeBSD` cross-compilation.
I would expect 2 commits:
`Makefile: Add Linux -> FreeBSD init cross-compilation support`
`CI: Add init cross-compilation workflow` 

`build fixes for macOS` is very vague and there is no commit message text besides titile...

src/vmm/build.rs
```diff
@@ -9,4 +9,8 @@ fn main() {
         println!("cargo:rustc-env=KRUN_EDK2_BINARY_PATH={edk2_binary_path}");
         println!("cargo:rerun-if-env-changed=KRUN_EDK2_BINARY_PATH");
     }
+
+    if std::env::var("CARGO_CFG_TARGET_OS").as_deref() == Ok("macos") {
+        println!("cargo:rustc-link-lib=framework=Hypervisor");
+    }
```
mtjhrc, line 15: `libkrun` crate already links this. Do we need this in both places?
Possibly this is the correct place and it can be removed from `libkrun` crate?

nohajc: Good point, I will test this. I had the linker error when trying to run cargo test on the vmm crate so you're probably right.

nohajc:
@mtjhrc Split commits and improved messages.

src/vmm/build.rs
nohajc: Still works after removing from libkrun crate.

src/libkrun/build.rs
```diff
@@ -10,6 +10,4 @@ fn main() {
         std::env::var("CARGO_PKG_VERSION_MAJOR").unwrap(), std::env::var("CARGO_PKG_VERSION_MAJOR").unwrap(),
         std::env::var("CARGO_PKG_VERSION_MAJOR").unwrap(), std::env::var("CARGO_PKG_VERSION_MINOR").unwrap()
     );
-    #[cfg(target_os = "macos")]
-    println!("cargo:rustc-link-lib=framework=Hypervisor");
```
mtjhrc, line 14: Please also split the Hypervisor linking move into its own commit. It's a fundamental change to how these crates are linked, and bundling it with the test fixes hides it a little bit. If this accidentally breaks something, a separate commit makes it obvious what happened, and it helps us review it properly.

(Having all the unit test fixes together is totally ok  - those tests were fully broken already)

mtjhrc (changes requested):

mtjhrc:
There are still `cargo test` failures. Also I need to specify `CC_LINUX` env var otherwise `cargo test` doesn't even run.
Do you have that, or how are you using it? 

You should probably also disable these failing tests: 
```
test legacy::i8042::tests::test_i8042_kbd ... FAILED
test legacy::i8042::tests::test_i8042_read_write_and_event ... FAILED
```

<details>
<summary><code>$ CC_LINUX=... cargo test</code></summary>

```
$ CC_LINUX="/usr/bin/clang -target aarch64-linux-gnu -fuse-ld=lld --sysroot $(pwd)/linux-sysroot -B$(pwd)/linux-sysroot/usr/lib/gcc/aarch64-linux-gnu/12 -L$(pwd)/linux-sysroot/usr/lib/gcc/aarch64-linux-gnu/12 -Wno-c23-extensions" cargo test
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.04s
     Running unittests src/lib.rs (target/debug/deps/arch-7e7fa7be3e6801ce)
/Users/mhrica/Dev/libkrun/c/main/target/debug/deps/arch-7e7fa7be3e6801ce: replacing existing signature

running 2 tests
test aarch64::tests::test_regions_gt_1024gb ... ok
test aarch64::tests::test_regions_lt_1024gb ... ok

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running unittests src/lib.rs (target/debug/deps/arch_gen-f9e1e70fc02b4a7c)
/Users/mhrica/Dev/libkrun/c/main/target/debug/deps/arch_gen-f9e1e70fc02b4a7c: replacing existing signature

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running unittests src/lib.rs (target/debug/deps/aws_nitro-642edc16ad499ce2)
/Users/mhrica/Dev/libkrun/c/main/target/debug/deps/aws_nitro-642edc16ad499ce2: replacing existing signature

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running unittests src/lib.rs (target/debug/deps/cpuid-7d75ebe5abbc11cd)
/Users/mhrica/Dev/libkrun/c/main/target/debug/deps/cpuid-7d75ebe5abbc11cd: replacing existing signature

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running unittests src/lib.rs (target/debug/deps/devices-ad2d2385014d2cb8)
/Users/mhrica/Dev/libkrun/c/main/target/debug/deps/devices-ad2d2385014d2cb8: replacing existing signature

running 51 tests
test legacy::i8042::tests::test_i8042_commands ... ok
test legacy::rtc_pl031::tests::test_rtc_read_write_and_event ... ok
test bus::tests::busrange_cmp_and_clone ... ok
test bus::tests::bus_read_write ... ok
test bus::tests::bus_read_write_values ... ok
test bus::tests::test_display_error ... ok
test legacy::i8042::tests::test_i8042_buffer ... ok
test bus::tests::bus_insert ... ok
test virtio::console::port_queue_mapping::test::test_queue_idx_to_port_id_ok ... ok
test virtio::console::port_queue_mapping::test::test_port_id_to_queue_idx ... ok
test virtio::console::port_queue_mapping::test::test_queue_idx_to_port_id_panic_rx_control - should panic ... ok
test legacy::i8042::tests::test_i8042_kbd ... FAILED
test legacy::i8042::tests::test_i8042_read_write_and_event ... FAILED
test virtio::console::port_queue_mapping::test::test_queue_idx_to_port_id_panic_tx_control - should panic ... ok
test virtio::descriptor_utils::tests::reader_test_simple_chain ... ok
test virtio::descriptor_utils::tests::reader_test_incompatible_chain ... ok
test virtio::descriptor_utils::tests::read_full ... ok
test virtio::descriptor_utils::tests::reader_writer_shared_chain ... ok
test virtio::descriptor_utils::tests::reader_writer_shattered_object ... ok
test virtio::descriptor_utils::tests::split_beginning ... ok
test virtio::descriptor_utils::tests::reader_unexpected_eof ... ok
test virtio::descriptor_utils::tests::split_end ... ok
test virtio::descriptor_utils::tests::split_border ... ok
test virtio::descriptor_utils::tests::split_outofbounds ... ok
test virtio::descriptor_utils::tests::split_middle ... ok
test virtio::descriptor_utils::tests::writer_test_incompatible_chain ... ok
test virtio::descriptor_utils::tests::write_full ... ok
test virtio::fs::multikey::test::get ... ok
test virtio::descriptor_utils::tests::writer_test_simple_chain ... ok
test virtio::fs::multikey::test::remove ... ok
test virtio::fs::multikey::test::update_alt_key ... ok
test virtio::fs::multikey::test::update_both_keys_alt ... ok
test virtio::fs::multikey::test::update_both_keys_main ... ok
test virtio::fs::multikey::test::update_main_key ... ok
test virtio::fs::multikey::test::update_value ... ok
test virtio::fs::read_only::tests::read_only_open_flags_allow_append ... ok
test virtio::fs::read_only::tests::read_only_open_flags_reject_truncate ... ok
test virtio::fs::read_only::tests::read_only_open_flags_reject_write_access ... ok
test virtio::mmio::tests::test_ack_features_by_page ... ok
test virtio::mmio::tests::test_bus_device_activate ... ok
test virtio::mmio::tests::test_bus_device_read ... ok
test virtio::mmio::tests::test_get_acked_features ... ok
test virtio::mmio::tests::test_bus_device_reset ... ok
test virtio::mmio::tests::test_get_avail_features ... ok
test virtio::mmio::tests::test_new ... ok
test virtio::mmio::tests::test_set_acked_features ... ok
test virtio::queue::tests::test_add_used ... ok
test virtio::queue::tests::test_checked_new_descriptor_chain ... ok
test virtio::queue::tests::test_queue_processing ... ok
test virtio::queue::tests::test_queue_validation ... ok
test virtio::mmio::tests::test_bus_device_write ... ok

failures:

---- legacy::i8042::tests::test_i8042_kbd stdout ----

thread 'legacy::i8042::tests::test_i8042_kbd' (4502008) panicked at src/devices/src/legacy/i8042.rs:416:13:
assertion failed: i8042.kbd_interrupt_evt.read().unwrap() > 1

---- legacy::i8042::tests::test_i8042_read_write_and_event stdout ----

thread 'legacy::i8042::tests::test_i8042_read_write_and_event' (4502009) panicked at src/devices/src/legacy/i8042.rs:333:9:
assertion `left == right` failed
  left: 1
 right: 2


failures:
    legacy::i8042::tests::test_i8042_kbd
    legacy::i8042::tests::test_i8042_read_write_and_event

test result: FAILED. 49 passed; 2 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.13s

error: test failed, to rerun pass `-p devices --lib`
# Last command returned 101
```
</details>

src/libkrun/build.rs
nohajc: Makes sense. Done.

nohajc:
@mtjhrc Yeah, currently I have `CC_LINUX` in my vscode settings. I suppose this could be improved overall but that's a task for another day.

Sorry, I think I actually only tried to run `mmio` and `vstate` tests. At first, I was mostly focused on build failures.

nohajc:
I'd suggest we also add `cargo test` to CI in another PR.

nohajc:
As for those two particular `legacy::i8042` tests, not sure if I should just disable them. They don't seem macOS specific and I have no idea why they're failing. Maybe it would deserve more investigation?

Claude suggests some assertions in those tests are correct on Linux but not macOS.

IMHO if the primary goal here was to fix the build, we could address this properly in another PR. I already feel I should have made my PR more focused in the first place.

mtjhrc:
> As for those two particular legacy::i8042 tests, not sure if I should just disable them. They don't seem macOS specific and I have no idea why they're failing. Maybe it would deserve more investigation?

I wasn't sure, but I thought that's an `x86_64` thing, I checked the code and yes, the `i8042` isn't used on aarch64. Only thing that uses it is `PortIODeviceManager` which is behind `#[cfg(target_arch = "x86_64")]`. (And we don't support `x86_64` macOS...)

So the whole `i8042` module and it's tests should be skipped on macOS.

nohajc:
> So the whole i8042 module and it's tests should be skipped on macOS.

Thanks, I'll disable them for aarch64 then (which includes Linux).

src/vmm/src/macos/vstate.rs
```diff
@@ -601,308 +600,110 @@ enum VcpuEmulation {
...
-            // Make sure the Vcpu is out of KVM_RUN.
-            self.send_event(VcpuEvent::Pause).unwrap();
-            // Close the original channel so that the Vcpu thread errors and goes to exit state.
-            let (event_sender, _event_receiver) = unbounded();
-            self.event_sender = event_sender;
-            // Wait for the Vcpu thread to finish execution
-            self.vcpu_thread.take().unwrap().join().unwrap();
-        }
-    }
+    use arch::aarch64::layout::DRAM_MEM_START_EFI;
```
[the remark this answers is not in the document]
mtjhrc: we don't support x86_64 macOS => file is only compiled on macOS => `#[cfg(target_arch = "aarch64")]` is always true

nohajc:
Code review suggestions addressed. I still have three more tests failing:

```
failures:
    event_manager::tests::test_modify
    event_manager::tests::test_register
    event_manager::tests::test_unregister
```

And actually there's more because `cargo` short-circuits whenever any module's tests fail:

```
failures:
    rutabaga_gralloc::gralloc::tests::create_render_target
    rutabaga_gralloc::gralloc::tests::create_video_buffer
    rutabaga_gralloc::gralloc::tests::export_and_map
```

This is turning into a bit of a rabbit hole. :) And not really what I set out to solve in the beginning.

**UPDATE:** Actually managed to fix everything but please review the event_manager tests because the suggested fix was AI generated. The last issue I have running `cargo test` on macOS is that I need to raise `ulimit -n` to about 1024. The default 256 is too low.

Makefile
```diff
@@ -179,10 +191,12 @@ $(FREEBSD_ROOTFS_DIR)/.sysroot_ready: $(FREEBSD_BASE_TXZ)
 	@cd $(FREEBSD_ROOTFS_DIR) && tar xJf base.txz 2>/dev/null || true
 	@touch $@
 
+BSD_ARCH=$(subst x86_64,amd64,$(ARCH))
```
mtjhrc: We also need to normalize the aarch64/arm64 name too (otherwise the url constructed from this would not be correct on aarch64 Linux):
```suggestion
BSD_ARCH=$(subst x86_64,amd64,$(subst aarch64,arm64,$(ARCH)))
```

nohajc: Thanks. Yeah, this is kind of messy. I forgot `uname` on Mac returns something else than Linux.

nohajc: Which reminds me I only added x86_64 Linux to the GitHub workflow... otherwise I would have caught this. Let me fix that too.

mtjhrc:
> This is turning into a bit of a rabbit hole. :) And not really what I set out to solve in the beginning.

Huge thanks though! The macOS unit tests have been completely unusable, this gives us an opportunity to properly start maintaining them.

> Actually managed to fix everything but please review the event_manager tests because the suggested fix was AI generated. 

Looks fine for now  - the sematics of the emulated EventFd are kind of broken but the actual non-test code is smart enough to avoid the problems, I plan to look into refactoring EventFd stuff in libkrun 2.0 so there's a separate explicit read and write end.

> The last issue I have running cargo test on macOS is that I need to raise ulimit -n to about 1024. The default 256 is too low.

Yes this fine and generally recommended when using libkrun - we consume a lot of file descriptors. Btw by default macOS gives you a lower limit over SSH (256 fds) than e.g. if you open a terminal in the GUI...

nohajc:
> Huge thanks though!

No problem! This project is fun. :)

Makefile
nohajc: Both done.

nohajc:
Example of successful cross-compilation workflow runs:
https://github.com/nohajc/libkrun/actions/runs/24682731130/job/72183942282

nohajc:
Ok, I think it's done. Any other thoughts @mtjhrc? :)

mtjhrc (approved):
LGTM, this is great, fixing the unit tests on macOS has been long overdue, thanks!

nohajc:
Some of them which don't require starting a VM could be even run in CI.

PR #637, opened 2026-04-15
Add `KRUN_FS_ROOT_TAG` define and document `krun_set_root` alternative

d-e-s-o:
Pull request #623 added support for read-only mounts. It seems only consequent to also provide users with the ability to set read-only state for the root file system.
Introduce `krun_set_root2()`, which extends `krun_set_root()` with a read_only parameter to allow exposing the root filesystem as read-only to the guest. The original `krun_set_root()` now delegates to the new function with `read_only=false`, preserving the previous behavior.

### DIFF
--- a/include/libkrun.h
+++ b/include/libkrun.h
@@ -98,7 +98,16 @@ int32_t krun_free_ctx(uint32_t ctx_id);
 int32_t krun_set_vm_config(uint32_t ctx_id, uint8_t num_vcpus, uint32_t ram_mib);
 
 /**
+ * The virtiofs tag used for the root filesystem. Can be used with krun_add_virtiofs*
+ * for more control over root filesystem parameters (e.g. read-only, DAX window size).
+ */
+#define KRUN_FS_ROOT_TAG "/dev/root"
+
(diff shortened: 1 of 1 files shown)


mtjhrc:
I'm not really a fan of adding this right now, users can already do `krun_add_virtiofs3(ctx, "/dev/root", "/path/to/root", /*shm_size*/0, /*read-only*/ true)`.

Only problem is user has to know we use "/dev/root" in the default cmdline arguments for the kernel. Only addition I think we should consider is  `#define KRUN_FS_ROOT_TAG  "/dev/root"` in the header.

For libkrun 2.0 I think we should make the user always pass in the virtiofs tag to boot from explicitly, that is more obvious to the user and flexible.

mtjhrc:
Also we could also add an explicit comment to `krun_set_root` to tell users to use `krun_add_virtiofs*` instead to specify more parameters for the root virtiofs.

d-e-s-o:
Sure, done.

mtjhrc (approved):
LGTM, thanks!

PR #640, opened 2026-04-17
init/aws-nitro: Continue if module already loaded

tylerfanelli:
If a module is already loaded, then the initramfs can continue on loading other modules. Although returning an error, this does not require failure of the initramfs.


dorindabassey (approved):

jakecorrenti:
@tylerfanelli can you rebase. it should fix the CI failures

Issue #634, opened 2026-04-13
Better C API + native Rust API


mtjhrc:
For 2.0 we want to add a native Rust API, but we still plan to support users using a C API. 

The C API will have breaking changes too - we want to namely don't having any implicit devices and providing more clear errors.

We don't want 2 entry points for every call in libkrun - C API should be a layer built on top of Rust API.  The new C API and Rust API should be reasonably similar which is why I think we need to think about this together.

mtjhrc:
Here's a rough API I have so far, if anyone thinks this is going in the wrong direction:

Using Rust you con configure libkrun as this:
```Rust
let rootfs = FsDevice::new("/dev/root", new_root).context("create rootfs")?;

let mut console_builder = ConsoleDevice::builder();
let payload = Init::builder(&rootfs, &mut console_builder)
    .exec(guest_cmd, &guest_args)
    .context("exec")?
    .workdir("/")
    .context("workdir")?
    .env(&["HOME=/root", "TERM=xterm-256color"])
    .context("env")?
    .build()
    .context("build payload")?;
let console = console_builder.build().context("build console")?;

let balloon = BalloonDevice::new().context("create balloon")?;
let rng = RngDevice::new().context("create rng")?;

let mut devices = MmioDeviceManager::new();
devices.add(rootfs);
devices.add(console);
devices.add(balloon);
devices.add(rng);

let mut vmm = VmmBuilder::new()
    .vcpus(2)
    .context("vcpus")?
    .ram_mib(512)
    .context("ram")?
    .payload(payload)
    .devices(devices)
    .build()
    .context("build vmm")?;
vmm.run();
```
This API Rust API can actually be called 2 different ways just by changing dependency in Cargo.toml from `libkrun` to `libkrun-via-cdylib`:
 - natively linked statically as a rust crate
 - dynamcially linked via C abi - this acts as Rust bindings to a C   

corresponding C side looks like:
```C
#define KRUN_CALL(call)                                                        \
    do {                                                                       \
        KrunError _err = (call);                                               \
        if (_err.code) {                                                       \
            fprintf(stderr, "%s failed: %s\n", #call,                          \
                    krun_error_message(&_err));                                \
            krun_error_free(&_err);                                            \
            return 1;                                                          \
        }                                                                      \
    } while (0)

KRUN_CALL(krun_init_log(KRUN_LOG_TARGET_DEFAULT, KRUN_LOG_LEVEL_WARN,
[...]

PR #642, opened 2026-04-17
Add vhost-user device support (RNG, sound, vsock, CAN, console, input, rtc)

dorindabassey:
Implement vhost-user protocol support in libkrun, enabling out-of-process virtio device backends using rust-vmm/vhost-device.
  ### Key Changes                                                                                         
                  
  - **Core vhost-user infrastructure**: Generic `VhostUserDevice` wrapper, file-backed memory (memfd), config space support
  - **Device support**: RNG, sound, vsock, CAN, console, rtc and input devices                               
  - **API**: `krun_set_vhost_user_device()` with public constants for each device type                   
  - **Examples**: Integration in `chroot_vm` for all supported devices

All devices tested and working with corresponding vhost-device backends.

### DIFF
--- a/include/libkrun.h
+++ b/include/libkrun.h
@@ -753,6 +753,7 @@ int32_t krun_set_snd_device(uint32_t ctx_id, bool enable);
  */
 #define KRUN_VIRTIO_DEVICE_CONSOLE 3
 #define KRUN_VIRTIO_DEVICE_RNG 4
+#define KRUN_VIRTIO_DEVICE_RTC 17
 #define KRUN_VIRTIO_DEVICE_INPUT 18
 #define KRUN_VIRTIO_DEVICE_VSOCK 19
 #define KRUN_VIRTIO_DEVICE_SND 25
@@ -772,6 +773,13 @@ int32_t krun_set_snd_device(uint32_t ctx_id, bool enable);
  */
 #define KRUN_VHOST_USER_RNG_NUM_QUEUES 1
 #define KRUN_VHOST_USER_RNG_QUEUE_SIZES ((uint16_t[]){256})
+
+/**
+ * Vhost-user RTC device default queue configuration.
+ * RTC device uses 2 queues: requestq (idx 0), alarmq (idx 1).
+ */
+#define KRUN_VHOST_USER_RTC_NUM_QUEUES 2
+#define KRUN_VHOST_USER_RTC_QUEUE_SIZES ((uint16_t[]){1024, 1024})
 
 /**
  * Vhost-user input device default queue configuration.
--- a/include/libkrun.h
+++ b/include/libkrun.h
@@ -753,6 +753,7 @@ int32_t krun_set_snd_device(uint32_t ctx_id, bool enable);
  */
 #define KRUN_VIRTIO_DEVICE_CONSOLE 3
 #define KRUN_VIRTIO_DEVICE_RNG 4
+#define KRUN_VIRTIO_DEVICE_INPUT 18
 #define KRUN_VIRTIO_DEVICE_VSOCK 19
 #define KRUN_VIRTIO_DEVICE_SND 25
 #define KRUN_VIRTIO_DEVICE_CAN 36
@@ -771,6 +772,13 @@ int32_t krun_set_snd_device(uint32_t ctx_id, bool enable);
  */
 #define KRUN_VHOST_USER_RNG_NUM_QUEUES 1
 #define KRUN_VHOST_USER_RNG_QUEUE_SIZES ((uint16_t[]){256})
+
+/**
+ * Vhost-user input device default queue configuration.
+ * Input device uses 2 queues: eventq (idx 0), statusq (idx 1).
+ */
+#define KRUN_VHOST_USER_INPUT_NUM_QUEUES 2
+#define KRUN_VHOST_USER_INPUT_QUEUE_SIZES ((uint16_t[]){1024, 1024})
 
 /**
  * Vhost-user sound device default queue configuration.
--- a/include/libkrun.h
+++ b/include/libkrun.h
@@ -751,10 +751,19 @@ int32_t krun_set_snd_device(uint32_t ctx_id, bool enable);
  * Vhost-user device types.
  * These correspond to virtio device type IDs for devices.
  */
+#define KRUN_VIRTIO_DEVICE_CONSOLE 3
 #define KRUN_VIRTIO_DEVICE_RNG 4
 #define KRUN_VIRTIO_DEVICE_VSOCK 19
 #define KRUN_VIRTIO_DEVICE_SND 25
 #define KRUN_VIRTIO_DEVICE_CAN 36
+
+/**
+ * Vhost-user console device default queue configuration.
+ * Console device uses 4 queues for multiport support:
+ * receiveq (idx 0), transmitq (idx 1), control receiveq (idx 2), control transmitq (idx 3).
+ */
+#define KRUN_VHOST_USER_CONSOLE_NUM_QUEUES 4
+#define KRUN_VHOST_USER_CONSOLE_QUEUE_SIZES ((uint16_t[]){128, 128, 64, 64})
 
 /**
  * Vhost-user RNG device default queue configuration.
--- a/include/libkrun.h
+++ b/include/libkrun.h
@@ -748,6 +748,76 @@ int krun_add_input_device_fd(uint32_t ctx_id, int input_fd);
 int32_t krun_set_snd_device(uint32_t ctx_id, bool enable);
 
 /**
+ * Vhost-user device types.
+ * These correspond to virtio device type IDs for devices.
+ */
+#define KRUN_VIRTIO_DEVICE_RNG 4
+#define KRUN_VIRTIO_DEVICE_SND 25
+#define KRUN_VIRTIO_DEVICE_CAN 36
+
+/**
+ * Vhost-user RNG device default queue configuration.
+ * Use these when you want explicit defaults instead of auto-detection.
+ */
+#define KRUN_VHOST_USER_RNG_NUM_QUEUES 1
+#define KRUN_VHOST_USER_RNG_QUEUE_SIZES ((uint16_t[]){256})
+
+/**
+ * Add a vhost-user device to the VM.
+ *
+ * This function adds a vhost-user device by connecting to an external
+ * backend process (e.g., vhost-device-rng, vhost-device-snd). The backend
+ * must be running and listening on the specified socket before starting the VM.
+ *
+ * This API is designed for devices like RNG, sound, and CAN.
+ *
+ * Arguments:
+ *  "ctx_id"       - the configuration context ID.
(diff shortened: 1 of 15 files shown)


src/vmm/src/builder.rs
```diff
@@ -461,6 +467,14 @@ impl Display for StartMicrovmError {
                     "Cannot initialize a MMIO Snd Device or add a device to the MMIO Bus. {err_msg}"
                 )
             }
+            RegisterVhostUserDevice(ref err) => {
+                let mut err_msg = format!("{err}");
```
bilelmoussaoui: Given that mmio::Error likely implements Display, you could use to_string() here instead of going through format! no?

src/vmm/src/builder.rs
```diff
@@ -1331,9 +1367,79 @@ fn load_payload(
                     return Err(StartMicrovmError::MissingKernelConfig);
                 };
 
-            let kernel_region = unsafe {
-                MmapRegion::build_raw(kernel_host_addr as *mut u8, kernel_size, 0, 0)
-                    .map_err(StartMicrovmError::InvalidKernelBundle)?
+            #[cfg(all(feature = "vhost-user", target_os = "linux"))]
```
bilelmoussaoui, line 1369: I guess you could have put this cfg above a scope defined with `{}` so you don't need to repeat it multiple times?

src/vmm/src/builder.rs
```diff
@@ -1502,10 +1608,74 @@ pub fn create_guest_memory(
             .map_err(StartMicrovmError::ShmCreate)?;
     }
 
+    // For vhost-user devices, we need file-backed memory so the backend can mmap it
+    #[cfg(all(feature = "vhost-user", target_os = "linux"))]
```
bilelmoussaoui, line 1605: Same here

src/vmm/src/builder.rs
```diff
@@ -1502,10 +1608,74 @@ pub fn create_guest_memory(
...
+            debug!(
+                "Creating file-backed memory for vhost-user (regions: {})",
+                arch_mem_regions.len()
+            );
+            // Create file-backed memory regions using memfd
+            let regions_with_files: Vec<_> = arch_mem_regions
+                .iter()
+                .map(|(addr, size)| {
+                    debug!(
+                        "  Creating memfd for region: addr=0x{:x}, size=0x{:x}",
```
bilelmoussaoui:
```suggestion
                        "Creating memfd for region: addr=0x{:x}, size=0x{:x}",
```

src/vmm/src/builder.rs
```diff
@@ -1502,10 +1608,74 @@ pub fn create_guest_memory(
...
+                        addr.0, size
+                    );
+                    // SAFETY: memfd_create is called with a valid null-terminated C string and valid flags.
+                    // File descriptor ownership is transferred to File::from_raw_fd below.
+                    let memfd = unsafe {
+                        let fd = libc::memfd_create(
+                            b"guest_mem\0".as_ptr() as *const libc::c_char,
```
bilelmoussaoui: You can use c"" nowadays no?

src/vmm/src/builder.rs
```diff
@@ -1502,10 +1608,74 @@ pub fn create_guest_memory(
...
+                        }
+                        if libc::ftruncate(fd, *size as i64) < 0 {
+                            error!(
+                                "Failed to ftruncate memfd: {:?}",
+                                io::Error::last_os_error()
+                            );
+                            libc::close(fd);
+                            return Err(io::Error::last_os_error());
+                        }
+                        debug!("  Created memfd with fd={}", fd);
```
bilelmoussaoui:
```suggestion
                        debug!("Created memfd with fd={}", fd);
```

src/devices/src/virtio/vhost_user/device.rs
```diff
@@ -0,0 +1,591 @@
...
+    /// * `device_type` - Virtio device type ID
+    /// * `device_name` - Human-readable device name for logging
+    /// * `num_queues` - Number of queues (0 = query backend via MQ protocol)
+    /// * `queue_sizes` - Size for each queue (empty = use default 256)
+    ///
+    /// # Returns
+    ///
+    /// A new VhostUserDevice or an error if connection fails.
+    pub fn new(
+        socket_path: &str,
```
bilelmoussaoui: an actual Path maybe?

src/devices/src/virtio/vhost_user/device.rs
```diff
@@ -0,0 +1,591 @@
...
+
+        let default_size = queue_sizes.last().copied().unwrap_or(256);
+        let queue_configs: Vec<_> = (0..actual_num_queues)
+            .map(|i| {
+                let size = queue_sizes.get(i).copied().unwrap_or(default_size);
+                QueueConfig::new(size)
+            })
+            .collect();
+
+        Ok(VhostUserDevice {
```
bilelmoussaoui:
```suggestion
        Ok(Self {
```

bilelmoussaoui (commented):
various comments

src/vmm/src/builder.rs
dorindabassey: The cfg repetition is intentional to ensure both branches compile cleanly.

dorindabassey:
Thanks for the reviews @bilelmoussaoui! 
Since this PR builds on several other pending PRs, I'll address all the feedback here in #642 to avoid rebase churn.

include/libkrun.h
```diff
@@ -747,6 +747,123 @@ int krun_add_input_device_fd(uint32_t ctx_id, int input_fd);
...
+ *
+ * Returns:
+ *  Zero on success or a negative error number on failure.
+ *  -EINVAL  - Invalid parameters (device_type, socket_path, etc.)
+ *  -ENOENT  - Context doesn't exist
+ *  -ENOTSUP - vhost-user support not compiled in
+ */
+int32_t krun_add_vhost_user_device(uint32_t ctx_id,
+                                   uint32_t device_type,
+                                   const char *socket_path,
```
ggoodman, line 862: Does this need to be a _path_ at the API level? What if we already have a file descriptor?

mtjhrc: Fair, we should probably support that, but this is not really an issue blocking the current PR really, this is getting merged into main which is now libkrun 2.x branch. Every public API function is about to get replaced.  #634

src/utils/Cargo.toml
```diff
@@ -12,11 +12,11 @@ bitflags = "1.2.0"
 libc = ">=0.2.85"
 log = "0.4.0"
 nix = "0.30.1"
-vmm-sys-util = "0.14"
+vmm-sys-util = "0.15"
```
mtjhrc: Btw why was this upgraded? The commit message doesn't mention it, though I have no objections to the bump itself.

It seems this is why the CI is failing. This upgraded the dependency here but didn't regenerate the Cargo.lock in the gui-vm example, which also uses this crate. [^1]

[^1]: Maybe it shouldn't use it since is not public API but it does... On the other hand the gui-vm is like a test/reference implementation and not something we actively publish, so it's fine.

mtjhrc: Actaully the version bump also broke the TDX it seems:
```
error[E0308]: mismatched types
   --> src/vmm/src/linux/tee/inteltdx.rs:28:33
    |
 28 |         launcher.init_vm(&caps, cpuid).map_err(Error::InitVm)?;
    |                  -------        ^^^^^ expected `FamStructWrapper<kvm_cpuid2>`, found a different `FamStructWrapper<kvm_cpuid2>`
    |                  |
    |                  arguments to this method are incorrect
    |
note: there are multiple different versions of crate `vmm_sys_util` in the dependency graph
   --> /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/vmm-sys-util-0.14.0/src/fam.rs:164:1
    |
164 | pub struct FamStructWrapper<T: Default + FamStruct> {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ this is the expected type
    |
```

dorindabassey: This issue was introduced in commit https://github.com/containers/libkrun/pull/588/changes/8b5f1fa48168c118f7d77c4a4eb9b522661da755 from PR #588, which upgraded to vm-memory 0.17.
There's a dependency incompatibility between the vhost crate and TDX requirements.
- vhost 0.14 requires vm-memory 0.16 + vmm-sys-util 0.14 (TDX works here) which means reverting the commit https://github.com/containers/libkrun/pull/588/changes/8b5f1fa48168c118f7d77c4a4eb9b522661da755 
- vhost 0.15/0.16 requires vm-memory 0.17 + vmm-sys-util 0.15 (correct vm-memory but breaks TDX because TDX requires vmm-sys-util 0.14)

To resolve this, we need either:
  1. Update TDX dependencies to work with vmm-sys-util 0.15, or                                          
  2. Temporarily pin to vm-memory 0.16 until upstream dependencies align.

dorindabassey: fixed it by adding vmm-sys-util 0.15 as optional dependency for vhost-user feature, and converted EventFd types when calling `set_vring_kick`/`set_vring_call` so we don't have to wait for upstream dep updates.

mtjhrc (dismissed):
LGTM, thanks! 

Code LGTM. I've only tested Sound and Vsock specifically now (and NET before which is not in the list of "officially" supported devices yet).

src/devices/src/virtio/vhost_user/device.rs
```diff
@@ -0,0 +1,558 @@
...
+
+        // Unregister activate_evt as it's only needed once
+        event_manager
+            .unregister(self.activate_evt.as_raw_fd())
+            .unwrap_or_else(|e| {
+                error!(
+                    "{}: failed to unregister activate event: {e:?}",
+                    self.device_name
+                );
+            });
```
mtjhrc, line 561: This is not really correct, you will miss registering the file descriptors on second activate() call after reset. But It still wouldn't work even if you removed the highlighted portion since the `VHOST_USER_PROTOCOL_F_RESET_DEVICE` is not implemented - the implemented `fn reset()` only works as a shutdown but not as an actual reset. 

Reset is a general problem in other libkrun devices too and is half-borked everywhere, there are also some unfinished things in our mmio transport layer implementation, so we should probably address that more generally first before implementing the vhost reset here. [^1]  [^2] [^3]

No change needed for now, I just wanted to point out the limitation which is not obvious. 

[^1]: https://github.com/containers/libkrun/pull/631#issuecomment-4237384774
[^2]: https://github.com/mtjhrc/libkrun/blob/e12b9b3780ffa8df9f3e1797b217d13453479167/src/devices/src/virtio/mmio.rs#L295-L296
[^3]: https://github.com/mtjhrc/libkrun/blob/e12b9b3780ffa8df9f3e1797b217d13453479167/src/devices/src/virtio/mmio.rs#L352-L353

slp (approved):
vhost-user support let's gooooo!

thanks @dorindabassey

ggoodman:
Do you imagine extending this to vhost-user-fs devices?

PR #644, opened 2026-04-20
virtio/gpu: mmap dma-buf blob fds into the guest shm region

aford173:
resource_map_blob() in the virgl_resource_map2/Linux variant only mmaps SHM (memfd) blob fds into the guest's virtio shm region directly. DMABUF blob fds are delegated to Rutabaga::resource_map -> virgl_renderer_resource_map2, which is designed for virglrenderer-allocated GPU memory and has no machinery to mmap an external dma-buf fd into a host VA range. For dma-bufs that are not virglrenderer-managed (V4L2 capture buffers exported via VIDIOC_EXPBUF, dma-bufs forwarded to the guest by a host PipeWire daemon, etc.) it silently no-ops and the guest's blob is left backed by zero pages.

The user-visible symptom is camera capture across the muvm PipeWire bridge delivering all-zero frames even though every layer reports success: SCM_RIGHTS forwards the dma-buf fd, libkrun classifies it correctly as DMABUF, the guest issues DRM_IOCTL_VIRTGPU_RESOURCE_CREATE_BLOB which succeeds, and multifilesink writes plausible-sized files that turn out to be 0x00 bytes from offset 0 to EOF.

A V4L2 EXPBUF dma-buf wraps regular kernel pages allocated via VIDIOC_REQBUFS/V4L2_MEMORY_MMAP and is CPU-mappable. It can be mmap'd into the guest's pre-reserved shm region with MAP_FIXED exactly the same way SHM (memfd) blobs are. Extending the existing SHM branch to also accept DMABUF gives the guest live pages and fixes camera capture without changes to the bridge, PipeWire, or the v4l2 userspace stack.

If the dma-buf is GPU-only and not CPU-mappable, mmap returns MAP_FAILED and we surface ErrUnspec, no worse than the prior silent no-op. The handle_type is now included in the error log to help diagnose mismatches.

The fd lifetime is fine: Rutabaga::export_blob returns a try_clone() of the descriptor, and the mapping persists past the temporary's drop because the original fd is retained in the resource's persistent handle slot.

Requires: https://github.com/AsahiLinux/muvm/pull/223

### DIFF
--- a/src/devices/src/virtio/gpu/virtio_gpu.rs
+++ b/src/devices/src/virtio/gpu/virtio_gpu.rs
@@ -20,6 +20,8 @@ use krun_display::{
 use libc::c_void;
 #[cfg(target_os = "macos")]
 use rutabaga_gfx::RUTABAGA_MEM_HANDLE_TYPE_APPLE;
+#[cfg(all(feature = "virgl_resource_map2", target_os = "linux"))]
+use rutabaga_gfx::RUTABAGA_MEM_HANDLE_TYPE_DMABUF;
 #[cfg(all(not(feature = "virgl_resource_map2"), target_os = "linux"))]
 use rutabaga_gfx::RUTABAGA_MEM_HANDLE_TYPE_OPAQUE_FD;
 #[cfg(all(feature = "virgl_resource_map2", target_os = "linux"))]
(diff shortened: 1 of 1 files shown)


slp:
The change looks good to me, but can't get it to work here after enabling the `virgl_renderer_map2` feature. Does it require any other changes apart from https://github.com/AsahiLinux/muvm/pull/223 ?

aford173:
@slp - From what I can tell,I only needed the libkrun and muvm patches to stream video from guest to host via Pipewire. I setup pipewire to run as a user service and the user that runs muvm is added to the groups for  seat,tty,video,render,audio,pipewire, and input. I only have one camera configured with pipewire, so it uses that one camera by default.  My host runs Xwayland.

I run "muvm -- bash" to get my virtual machine, then I run gstreamer to capture video inside the virtual machine

gst-launch0-1.0 pipewiresrc ! videoconvert ! ximagesink

The video streams from my laptop's built-in USB Camera to a window.

I haven't tried updating libkrunfw, but the version I have checked out is from December 2025.

aford173:
@slp - Is there anything you want me to do or change?

slp:
> @slp - Is there anything you want me to do or change?

I couldn't get it to work on Asahi Linux, but the change itself looks good to me and won't cause any regressions, so let's get it merged.

slp (approved):

aford173:
Thanks!

valpackett:
> I couldn't get it to work on Asahi Linux

I haven't been using `resource_map2` at all (and the non-map2 variant has already had the correct `export.handle_type != RUTABAGA_MEM_HANDLE_TYPE_OPAQUE_FD` check), and I also don't have PipeWire working properly on my aarch64 machine (i.e. camera input is all black, audio output is all silent). PipeWire wasn't even using dmabufs though… I'll try to finally get around to debugging that

Issue #632, opened 2026-04-13
Rewrite init in Rust


mtjhrc:
Currently our custom init is written in C, I believe it it would ease further development to rewrite it in Rust.

Unanswered questions we need to investigate:
- do we still keep statically linking it? - probably yes
  - is musl libc good enough for us? 
  - alternatively is there *really* no way to link against glibc statically?
  - alternatively can we just go no-std, no-libc route via https://github.com/bytecodealliance/rustix 
    - similar to nix crate but can make raw syscalls 
    - this should also shrink the init binary size, not sure if we care that much though
- how is Rust support for FreeBSD guests?
   - will we make sure upstream libkrun init supports FreeBSD?

mtjhrc:
Assigned to @jakecorrenti since I believe he started working on this

nohajc:
I can answer the FreeBSD part. The aarch64 variant is a tier 3 target which makes it more difficult to cross-compile.

https://doc.rust-lang.org/nightly/rustc/platform-support.html#tier-3

Currently, you need Rust nightly to make it somewhat reasonable.

Example: https://github.com/nohajc/anylinuxfs/blob/main/build-app.sh#L39-L45

nohajc:
Just want to mention there is a language with arguably better cross-compilation story, excellent C interop and producing lean binaries. Zig.

Not sure if the goal here is to just replace C with something easier to write or have everything in one language.

Anyway, I was actually considering Zig for the guest part of FreeBSD tests so I can avoid the Rust nightly issue. Although to be fair, Zig hasn't reached 1.0 yet, so it might not be considered stable either.

mtjhrc:
> Not sure if the goal here is to just replace C with something easier to write or have everything in one language.

For me "something easier to write" is the main goal, "have everything in one language" is also nice, but more like a secondary goal. Using something like Zig introduces a huge dependency for us, which I'd like to avoid. I've never used Zig, but it looked cool (less safety oriented than Rust though...)

nohajc:
It's somewhere between C and Rust safety-wise. I don't have much experience with it myself. It just crossed my mind.

It's probably still suitable for early adopters only. If you're looking for stable language editions there's no such thing currently.

Although there are projects which use it as a C cross-toolchain instead of gcc or llvm. It can serve as a really smart wrapper for clang/llvm. It has some pretty nice tricks up its sleeve like bundling all existing libcs for any platform you can think of. So you can do simpler builds without any sysroot for example.

[...]

PR #645, opened 2026-04-21
Rework workspace crates to get ready to publish them

slp:
Publishing libkrun to crates.io requires to be able to publish also every crate in the workspace. This requires the following changes, addressed in this PR:

- Rename crates to avoid potential collisions in crates.io
- Start managing crate versions.
- Add missing fields to each crate's Cargo.toml

We're setting up versioning in two different formats:

- 0.1.0-$LIBKRUN_VERSION: For crates that are only intended to be consumed by libkrun, we're leaving the 0.1.0 and adding the main crate version. Only the latter will be updated.

- 0.1.0: For display and input, we're giving them an independent version, and will be updated on its own pace accoding to SemVer.


src/devices/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "krun-devices"
-version = "0.1.0"
+version = "1.17.3"
```
mtjhrc: So this technically implies the API of the internal crates is stable (which it really isn't). Instead can we use something like `0.1.0-1.17.3`, where `0.1.0` is an unstable API version, and  `1.17.3` is the libkrun version this is for.

Longer term for libkrun 2.x I would eventually want to make things like a VirtioDevice trait public, but crucially IMO this has to be versioned totally independently from libkrun "core". Suppose libkrun-2.0 could depend on krun-devices 2.0, but libkrun 2.1, could start depending on devices 3.0.0 etc. (Or also start the version numbers with `0` - I'm not sure). So only users who implement custom devices have to follow frequent API changes (but they opted into it), users who just use builtin devices can just depend on libkrun 2.x not doing any breaking changes.

nohajc: If you wanted to avoid publishing the internal crates altogether, would it be feasible to refactor them into modules? Afaik that's the only way around it.

src/arch/Cargo.toml
```diff
@@ -15,14 +18,14 @@ libc = ">=0.2.39"
 vm-memory = { version = "0.17", features = ["backend-mmap"] }
 vmm-sys-util = "0.14"
 
-arch_gen = { path = "../arch_gen" }
-smbios = { path = "../smbios" }
-utils = { path = "../utils" }
+arch_gen = { package = "krun-arch-gen", version = "1.17.3", path = "../arch_gen" }
```
mtjhrc:
```suggestion
arch_gen = { package = "krun-arch-gen", version = "=1.17.3", path = "../arch_gen" }
```

I think everywhere the version requirement should be prefixed with `=` everywhere - we don't have any forward compatibility. Using just `version = "1.17.3"` implies  `"^1.17.3"` I think this means means: "version >=1.17.3 && version < 2.0.0"

src/devices/Cargo.toml
mtjhrc: > If you wanted to avoid publishing the internal crates altogether, would it be feasible to refactor them into modules? Afaik that's the only way around it.

Yes but problem is this effects the compilation speed I think and doesn't do incremental builds.  Each `crate` is a compilation unit.

src/arch/Cargo.toml
slp: Agreed, thanks.

src/devices/Cargo.toml
slp: TBH, I wasn't contemplating the possibility of anyone beyond `libkrun` using those crates. But if there are plans to do so, we can use your suggested `0.1.0-1.17.3`.

mtjhrc: Btw the `krun-display` and `krun-input` were meant to be consumed externally from the beginning, I wouldn't say they are fully stable though but for the rest of 1.x they might as well be. 
They provide Rust bindings to our C API for providing GPU displays and input devices, `gui_vm` example already depends on them.

slp: @mtjhrc in that case, I'd be inclined to use `0.1.0`, without the `-0.17.3`, so we remember those must be versioned independently.

jakecorrenti: I'm more in favor of `0.1.0` and dropping the suffix. Do we really care what version of libkrun the internal crate is for if it's not stable?

slp: > I'm more in favor of `0.1.0` and dropping the suffix. Do we really care what version of libkrun the internal crate is for if it's not stable?

If we don't add the suffix, we'd need to manually bump the version to an arbitrary number on every crate changed between libkrun releases, since we can't publish crate updates with the same version number.

jakecorrenti (dismissed):

src/arch/Cargo.toml
slp: We're now clamping dependencies for crates with version format `0.1.0-$LIBKRUN_VERSION`, and we leave regular SemVer rules to apply for the rest (right now, display and input).

d-e-s-o:
Very cool, thanks for working on this. One question: have you thought through how things will work with the `init.c` file required by the `devices` crate? This may require some moving around of code and making sure it's included in published packages or may even be blocked on https://github.com/containers/libkrun/issues/632, I fear. I may be missing something, though.

ggoodman:
Might that be helped with #593?

d-e-s-o:
> Might that be helped with #593?

Good point. I'd think that may help as well.

slp:
@d-e-s-o @ggoodman For 1.x, I was expecting consumers of the library from crates.io to export `KRUN_INIT_BINARY_PATH` pointing to their own `init`. For 2.0 we'll have #593 plus a new init written in Rust. Do you see a problem with the 1.x approach?

ggoodman:
Sounds plausible. The reason I proposed the PR is because that's the DX I prefer so I think I'm totally biased 😅.

mtjhrc:
I mean the env var isn't great, but not sure if now is the time to change it.

> Sounds plausible. The reason I proposed the PR is because that's the DX I prefer so I think I'm totally biased 😅.

Yeah but what we really want is no built-in init by default, no? But that also means introducing a built-time option to disable the baked in init. (or make it a default enabled feature I suppose?)
And then have a separate rust crate `krun-init` (like you have) which gives you the bytes you can pass into `krun_set_init` (or similar flow... ) explicitly. 

I mean this is the right direction, not sure how much of that we want to include in 1.18 though, In principle I am not against it, especially if we also want to publish the crates it might be nice to include this. On the other hand 1.18 was supposed to be released last week and we have a growing queue of changes waiting for 2.0.

d-e-s-o:
> Do you see a problem with the 1.x approach?

I was just worried about the UX when one attempts to consume the library as a `crates.io` dependency and it just won't work by default. That being said, personally I am fine with anything that moves us closer to enabling consumption from `crates.io` (which this change is helping with), so I am fine with tackling issues step by step.

slp:
I think this one gets us into a better (but not ideal) state for potential crates.io users. Then, for 2.0, we'll have the #593 and a Rust init, which will make everything smoother.

mtjhrc (approved):
LGTM, thanks!

PR #284, opened 2025-03-21
feat: add support for PVH boot protocol

dsseng:
WIP support for PVH boot, enabling e.g. smolBSD to boot.

Parts copied from:
- https://github.com/firecracker-microvm/firecracker/commit/a52890bac2a0928e8c9ea0abd0223145fbe37641
- https://github.com/firecracker-microvm/firecracker/commit/20b50ce11ed45d99e514f3eda025c185188cd15d
- https://github.com/firecracker-microvm/firecracker/commit/6e87d70b64fd99bac095f01f5f468b0cc9f278d5

PR #646, opened 2026-04-21
Add support for PVH boot protocol; FreeBSD VM tests

nohajc:
Adds three new integration test cases:
1. Boot FreeBSD VM and print OK
2. Test FreeBSD TCP client
3. Test FreeBSD TCP server

Adds PVH boot protocol support (based on #284) to enable FreeBSD boot on x86_64.

Direct kernel boot works
- on macOS and Linux aarch64 with the raw boot image and generic kernel configuration
- on Linux x86_64 with ELF PVH image and the Firecracker kernel configuration

Gvproxy with vfkit enabled on Linux via https://github.com/containers/gvisor-tap-vsock/commit/c09fb7d0a9e08260d238066c3c4311498d15311e

### DIFF
--- a/src/arch/src/x86_64/mod.rs
+++ b/src/arch/src/x86_64/mod.rs
@@ -21,7 +21,9 @@ use crate::x86_64::layout::{EBDA_START, FIRST_ADDR_PAST_32BITS, MMIO_MEM_START};
 #[cfg(feature = "tee")]
 use crate::x86_64::layout::{FIRMWARE_SIZE, FIRMWARE_START};
 use crate::{ArchMemoryInfo, InitrdConfig};
-use arch_gen::x86::bootparam::{boot_params, E820_RAM, E820_RESERVED};
+#[cfg(not(feature = "tee"))]
+use arch_gen::x86::bootparam::E820_RESERVED;
+use arch_gen::x86::bootparam::{boot_params, E820_RAM};
 use vm_memory::Bytes;
 use vm_memory::{Address, ByteValued, GuestAddress, GuestMemoryMmap};
 use vmm_sys_util::align_upwards;
@@ -358,12 +360,7 @@ fn configure_pvh(
 }
 
 #[cfg(not(feature = "tee"))]
-fn add_memmap_entry(
-    memmap: &mut Vec<hvm_memmap_table_entry>,
-    addr: u64,
-    size: u64,
-    mem_type: u32,
-) {
+fn add_memmap_entry(memmap: &mut Vec<hvm_memmap_table_entry>, addr: u64, size: u64, mem_type: u32) {
     memmap.push(hvm_memmap_table_entry {
         addr,
         size,
@@ -536,21 +533,48 @@ mod tests {
         let (arch_mem_info, arch_mem_regions) =
             arch_memory_regions(mem_size, Some(KERNEL_LOAD_ADDR), KERNEL_SIZE, 0, None);
         let gm = GuestMemoryMmap::from_ranges(&arch_mem_regions).unwrap();
-        configure_system(&gm, &arch_mem_info, GuestAddress(0), 0, &None, no_vcpus, false).unwrap();
+        configure_system(
+            &gm,
+            &arch_mem_info,
+            GuestAddress(0),
+            0,
+            &None,
+            no_vcpus,
+            false,
+        )
+        .unwrap();
 
         // Now assigning some memory that is equal to the start of the 32bit memory hole.
         let mem_size = 3328 << 20;
         let (arch_mem_info, arch_mem_regions) =
             arch_memory_regions(mem_size, Some(KERNEL_LOAD_ADDR), KERNEL_SIZE, 0, None);
         let gm = GuestMemoryMmap::from_ranges(&arch_mem_regions).unwrap();
-        configure_system(&gm, &arch_mem_info, GuestAddress(0), 0, &None, no_vcpus, false).unwrap();
+        configure_system(
+            &gm,
+            &arch_mem_info,
+            GuestAddress(0),
+            0,
+            &None,
+            no_vcpus,
+            false,
+        )
+        .unwrap();
 
         // Now assigning some memory that falls after the 32bit memory hole.
         let mem_size = 3330 << 20;
         let (arch_mem_info, arch_mem_regions) =
             arch_memory_regions(mem_size, Some(KERNEL_LOAD_ADDR), KERNEL_SIZE, 0, None);
         let gm = GuestMemoryMmap::from_ranges(&arch_mem_regions).unwrap();
-        configure_system(&gm, &arch_mem_info, GuestAddress(0), 0, &None, no_vcpus, false).unwrap();
+        configure_system(
+            &gm,
+            &arch_mem_info,
+            GuestAddress(0),
+            0,
+            &None,
+            no_vcpus,
+            false,
+        )
+        .unwrap();
     }
 
     #[test]
--- a/src/arch/src/x86_64/mod.rs
+++ b/src/arch/src/x86_64/mod.rs
@@ -21,10 +21,17 @@ use crate::x86_64::layout::{EBDA_START, FIRST_ADDR_PAST_32BITS, MMIO_MEM_START};
 #[cfg(feature = "tee")]
 use crate::x86_64::layout::{FIRMWARE_SIZE, FIRMWARE_START};
 use crate::{ArchMemoryInfo, InitrdConfig};
-use arch_gen::x86::bootparam::{boot_params, E820_RAM};
+use arch_gen::x86::bootparam::{boot_params, E820_RAM, E820_RESERVED};
 use vm_memory::Bytes;
 use vm_memory::{Address, ByteValued, GuestAddress, GuestMemoryMmap};
 use vmm_sys_util::align_upwards;
+
+#[cfg(not(feature = "tee"))]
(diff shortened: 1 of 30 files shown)


nohajc:
Ready for review.

nohajc:
Gonna make adjustments based on the new contribution guidelines. Mainly need to fix "co-authored by" for AI assistants.

tests/test_cases/src/test_freebsd_gvproxy_tcp_guest_connect.rs
```diff
@@ -66,7 +66,10 @@ mod host {
             if output_str == "OK\n" {
                 TestOutcome::Pass
             } else {
-                TestOutcome::Fail(format!("expected exactly {:?}, got {:?}", "OK\n", output_str))
+                TestOutcome::Fail(format!(
+                    "expected exactly {:?}, got {:?}",
+                    "OK\n", output_str
+                ))
```
mtjhrc, line 72: The file `test_freebsd_gvproxy_tcp_guest_connect.rs` was introduced as part of this PR. Please squash this formatting fix into the appropriate commit `tests: add FreeBSD guest support` in this case.

mtjhrc:
Thanks for working on this!  I'll take a more detailed look later. Also note, that I changed the PR title, because this not only adds the tests but implements/revives the PVH boot support.

nohajc:
Makes sense. PVH is the enabler for the full test suite to cover all  architectures but I wasn't sure at first if I can make it work.

tests/test_cases/src/test_freebsd_gvproxy_tcp_guest_connect.rs
nohajc: Done.

nohajc:
Rebased, fixed issues, updated gvproxy to 0.8.9, running all gvproxy tests in the CI workflow.

src/vmm/src/builder.rs
```diff
@@ -1505,15 +1530,21 @@ fn load_payload(
...
+            })
         }
-        Payload::Firmware => Ok((guest_mem, GuestAddress(arch::RESET_VECTOR), None, None, false)),
+        Payload::Firmware => Ok(LoadedPayload {
+            guest_mem,
+            entry_addr: GuestAddress(arch::RESET_VECTOR),
+            initrd_config: None,
+            kernel_cmdline: None,
+            pvh: false,
+        }),
```
mtjhrc, line 1547: Please avoid making commits like this, as far as I can tell this just fixes bad formatting in the first commit in the PR. 

(having a separate fmt/clippy commit is fine when fixing something that is already merged into main, otherwise please squash fixups like this).

tests/test_cases/src/tcp_tester.rs
```diff
@@ -71,9 +71,27 @@ impl TcpTester {
...
+                Ok(()) => {
+                    assert_eq!(&buf[..], b"ping!");
+                    break stream;
+                }
+                Err(_) => {
+                    drop(stream);
+                    thread::sleep(Duration::from_millis(200));
+                }
+            }
+        };
```
mtjhrc, line 100: This loop needs to have an upper bound on how many retries to do, otherwise this can spin forever instead of the test using this failing.

tests/test_cases/src/tcp_tester.rs
```diff
@@ -71,9 +71,27 @@ impl TcpTester {
...
 
     pub fn run_client(&self) {
-        let mut stream = connect(self.server_ip, self.port);
-        set_timeouts(&mut stream);
-        expect_msg(&mut stream, b"ping!");
+        // Retry connect + first read until the proxy/guest bridge is fully wired up.
+        // gvproxy (and libkrun's TSI) accept the host-side connection immediately,
+        // but the upstream dial to the guest only succeeds once the guest has
+        // reached listen()/accept(). Until then the host either reads zero bytes
+        // (timeout/WouldBlock) or sees an EOF from a forced close. Once the first
```
mtjhrc, line 78: Hmm, "timeout/wouldblock" makes sense to me, but I feel like seeing an EOF/forced close is unexpected. Did the connect from proxy timeout or something? I am a little worried this might be masking a bug somewhere in the system and the test just passes.  

If the problem is the upstream dial `connect` times out in the proxy while guest is booting, we can keep this test fix for now, but we should find a better way to sync this. Like only start connecting to guest once the guest is booted up? (but this needs a side channel for the sync, virtio-console would be nice for that but that is problematic on freebsd)

tests/test_cases/src/test_virtiofs_misc.rs
```diff
@@ -205,3 +208,7 @@ mod guest {
         }
     }
 }
+
+// Stub impl so the guest-agent binary compiles on non-Linux targets.
+#[cfg(all(feature = "guest", not(target_os = "linux")))]
+impl crate::Test for TestVirtioFsMisc {}
```
mtjhrc: I am not in favor of these stubs, instead we could just not compile this whole file in the guest. 

You can do this by doing:
```
#[cfg(any(feature = "host", target_os = "linux"))]
mod test_virtiofs_misc;
```

and in the testcase list just do:
```
#[cfg(any(feature = "host", target_os = "linux"))]
TestCase::new("virtiofs-misc", Box::new(TestVirtioFsMisc)),
```

tests/test_cases/src/test_vsock_guest_connect.rs
```diff
@@ -111,3 +119,7 @@ mod guest {
         }
     }
 }
+
+// Stub impl so the guest-agent binary compiles on non-Linux targets (vsock test won't be dispatched there).
+#[cfg(all(feature = "guest", not(target_os = "linux")))]
+impl crate::Test for TestVsockGuestConnect {}
```
mtjhrc: Same here as the other comment, I think it would be easier to skip compiling the whole file (which would also make it so we don't need all the `#[allow(dead_code)]`...

mtjhrc (changes requested):
Sorry for taking a while to review this. The PVH code LGTM (though I don't have much intricate knowledge about that area, it helps the code is tested :-) )  

I left some comments regarding improving the test code. 

Thanks!

src/vmm/src/builder.rs
nohajc: Missed this one, thanks.

tests/test_cases/src/tcp_tester.rs
nohajc: I did see unexpected EOF from the proxy if connected too early. It didn't help to retry the read but retrying connect worked. IMO better than the random sleep that was used before. I tried to go with it at first but it was too unreliable.

src/vmm/src/builder.rs
nohajc: Squashed.

tests/test_cases/src/test_vsock_guest_connect.rs
nohajc: Resolved.

tests/test_cases/src/test_virtiofs_misc.rs
nohajc: Done.

tests/test_cases/src/tcp_tester.rs
nohajc: Fixed.

tests/test_cases/src/common_freebsd.rs
```diff
@@ -0,0 +1,249 @@
...
+
+/// Normalize serial-console line endings for FreeBSD output assertions.
+///
+/// FreeBSD's serial console emits CRLF (`\r\n`); strip the `\r` so that
+/// test `check()` overrides can compare against plain `\n`-terminated strings.
+pub fn normalize_serial_output(bytes: Vec<u8>) -> String {
+    String::from_utf8_lossy(&bytes)
+        .replace("\r\n", "\n")
+        .replace('\r', "\n")
+}
```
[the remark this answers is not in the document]
nohajc: `String::from_utf8_lossy` is used in other tests too. I don't think we care about UTF errors.

src/arch/src/x86_64/mod.rs
```diff
@@ -253,6 +266,116 @@ pub fn configure_system(
...
+        );
+        if last_addr > first_addr_past_32bits {
+            add_memmap_entry(
+                &mut memmap,
+                first_addr_past_32bits.raw_value(),
+                last_addr.unchecked_offset_from(first_addr_past_32bits) + 1,
+                E820_RAM,
+            );
+        }
+    }
```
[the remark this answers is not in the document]
nohajc: Not entirely sure what's correct but the same memmap pattern was already used before:
https://github.com/nohajc/libkrun/blob/main/src/arch/src/x86_64/mod.rs#L299-L328

src/arch/src/x86_64/mod.rs
nohajc, line 424: The same offset math here...

nohajc, line 443: ... and here.

nohajc:
@mtjhrc Reviewed Gemini's findings... I don't think there's anything actionable.

mtjhrc:
I'm not sure - I think Gemini might be right, but in that case this is pre-exisiting bug... We we can address that in another PR, let's get this merged already - this will be quite helpful to have with the upcoming init rewrite.

I ran the test on my Mac - it works.

LGTM, just rebase this please.

nohajc:
Done.

mtjhrc (approved):

PR #647, opened 2026-04-24
init: don't fail if a fs is already mounted

slp:
This allows us to use init.krun while booting from an initramfs.

### DIFF
--- a/init/init.c
+++ b/init/init.c
@@ -520,19 +520,22 @@ static int mount_filesystems()
     }
 
     if (mount("proc", "/proc", "proc",
-              MS_NODEV | MS_NOEXEC | MS_NOSUID | MS_RELATIME, NULL) < 0) {
+              MS_NODEV | MS_NOEXEC | MS_NOSUID | MS_RELATIME, NULL) < 0 &&
+        errno != EBUSY) {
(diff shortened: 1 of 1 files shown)


init/init.c
mtjhrc, line 562: Why not this one too?

slp: This one didn't fail with an initramfs, but let's check EBUSY too for consistency.

init/init.c
```diff
@@ -520,19 +520,22 @@ static int mount_filesystems()
     }
 
     if (mount("proc", "/proc", "proc",
-              MS_NODEV | MS_NOEXEC | MS_NOSUID | MS_RELATIME, NULL) < 0) {
+              MS_NODEV | MS_NOEXEC | MS_NOSUID | MS_RELATIME, NULL) < 0 &&
+        errno != EBUSY) {
```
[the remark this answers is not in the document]
slp: SEV code is going to get dropped from `init/init.c`, and the existing implementation doesn't support booting from a generic initramfs either. NAK.

dorindabassey (approved):

mtjhrc (approved):
LGTM

PR #651, opened 2026-04-24
Remove EFI feature, bundled edk2 and documentation mentioning it

mtjhrc:
Users are expected to use `krun_set_firmware` or equivalent API we will provide in the future.


slp (approved):
Finally, let's do this, yay!

PR #654, opened 2026-04-28
Remove builtin virtio-snd device

mtjhrc:
The builtin virtio-snd device is not very actively maintained and has been planned for removal in 2.0. For libkrun 2.0, user are expected to either use `vhost-user-snd` using the newly added vhost-user support or pipewire passthrough via virtio-gpu cross-domain. 

This removes:
- The virtio-snd device implementation (src/devices/src/virtio/snd/)
- The 'snd' feature flag from all crates
- The pipewire (pw) dependency from devices
- The krun_set_snd_device() C API and KRUN_FEATURE_SND constant
- CI references to the snd feature and libpipewire-0.3-dev


slp (approved):
Nice, let's drop some weight.

PR #656, opened 2026-04-29
virtio/vsock: use device's peer_buf_alloc

slp:
In commit c0e42fb0e0 ("vsock/virtio: cap TX credit to local buffer size") the kernel stopped honoring our peer_buf_alloc value, capping it to its own.

Use the kernel's peer_buf_alloc instead of CONN_TX_BUF_SIZE as a hint of when we need to send a credit update.

### DIFF
--- a/src/devices/src/virtio/vsock/tsi_stream.rs
+++ b/src/devices/src/virtio/vsock/tsi_stream.rs
@@ -613,9 +613,7 @@ impl Proxy for TsiStreamProxy {
             -libc::EINVAL
         };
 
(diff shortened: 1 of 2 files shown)


dorindabassey (approved):
LGTM

mtjhrc:
Btw shouldn't this be tagged `cherry-pick-1.18` ?

slp:
> Btw shouldn't this be tagged `cherry-pick-1.18` ?

Indeed, thanks for pointing it out!

PR #658, opened 2026-04-30
tsi_dgram: fix sendto_addr binding for AF_INET6 sockets

nohajc:
I noticed DNS was broken in my alpine VM whenever I added an IPv6 nameserver to /etc/resolv.conf. This change fixes the issue.

### DIFF
--- a/src/devices/src/virtio/vsock/tsi_dgram.rs
+++ b/src/devices/src/virtio/vsock/tsi_dgram.rs
@@ -1,5 +1,5 @@
 use std::collections::HashMap;
-use std::net::{Ipv4Addr, SocketAddrV4};
+use std::net::{Ipv4Addr, Ipv6Addr, SocketAddrV4, SocketAddrV6};
 use std::num::Wrapping;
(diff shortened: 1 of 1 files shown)


mtjhrc:
Thanks for the fix! But actually, I just remembered I was going to fix this but I totally forgot about it: https://github.com/containers/libkrun/pull/472/changes/71d5270dc41927284060b7866bd714f6877c900b  
Hmm, my version seems to also handle the unix case, let me re-test that case and maybe open a PR.

 I was thinking recently about figuring out how to merge the test - I had trouble setting up a minimal test environment (user namespaces) without running libkrun in podman via crun... But I totally forgot I also fixed the TSI code

nohajc:
Sure, we can use your fix.

Anyway, it seems my users hit this issue couple times already but it wasn't evident whether it was IPv6 related. I only realized this time...

mtjhrc:
Closing in favor of #659 which is a bit more robust and handles the unix socket case too. Thanks for bringing this up though!

PR #659, opened 2026-04-30
vsock/tsi_dgram: Make sendto_addr bind to correct socket adress

mtjhrc:
We need to bind to the correct socket types (IPv6, Unix) instead of only IPv4. This fixes UDP and unix dgram tests hanging when waiting for reply.

This supersedes #658 by also fixing an issue for unix datagram sockets over TSI.


dorindabassey (approved):
LGTM

PR #660, opened 2026-04-30
windows: Fix polly and utils crates to compile on Windows

lstocchi:
This is the second PR for porting libkrun on Windows. Nothing fancy and relatively small. 
I'm just going back through the code I wrote for my tests over the last few weeks. Cleaning it up and pushing it with a bit more organization. Next would be the WHP crate. 

As @slp confirmed I'm going to push to main. 
As you can see the first two commits are the same as https://github.com/containers/libkrun/pull/609 (edit 5/5/26: updated epoll to support the change made in https://github.com/containers/libkrun/pull/612) because the main do not contain them. If you prefer we can merge them separately and rebase this PR. 

Overall, it's just a bunch of instructions to gate unix-related functions which won't be used on Windows and replace the misssing one with Windows code.

### DIFF
--- a/src/utils/src/windows/epoll.rs
+++ b/src/utils/src/windows/epoll.rs
@@ -1,0 +1,710 @@
+//! Epoll-like I/O event polling for Windows, backed by I/O Completion Ports.
+//!
+//! Uses an IOCP as the central event multiplexer with
+//! `NtAssociateWaitCompletionPacket` to bridge waitable kernel handles (like
+//! Windows Event objects used by `EventFd`) into the completion port.  This
+//! gives true O(1) wake-up with no handle-count limitations.
+//!
+//! ## How it works
+//!
+//! 1. [`Epoll::new`] creates an I/O Completion Port.
+//! 2. [`Epoll::ctl`] with [`ControlOperation::Add`] heap-allocates a [`Watch`]
+//!    struct, creates a *Wait Completion Packet* (WCP) via the NT native API,
+//!    and associates the caller's waitable handle with the IOCP.  The raw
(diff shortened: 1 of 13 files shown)


src/utils/src/windows/epoll.rs
```diff
@@ -0,0 +1,700 @@
...
+        let iocp_handle = self.iocp.handle;
+
+        let capacity = max_events.min(events.len());
+        if capacity == 0 {
+            return Ok(0);
+        }
+
+        const BATCH_SIZE: usize = 256;
+        let batch_limit = capacity.min(BATCH_SIZE);
+        let mut entries = [MaybeUninit::<OVERLAPPED_ENTRY>::uninit(); BATCH_SIZE];
```
mtjhrc: FYI in "#612 Remove unecessary heap allocations when processing Epoll events" I'm making wait() take `&mut self`, that way you can have `entries` array here as a `Vec` inside the struct (similar to macOS).

lstocchi: Updated. Thanks for the info!!

slp:
LGTM, thanks @lstocchi. Do you think the unit tests for Windows could be run in a GitHub runner (at least, the ones not requiring WHPX).

slp (approved):

lstocchi:
@slp yes, we should definitely add it. The problem is that we cannot use cargo test right now bc of the compilation issue with vmm-rust deps. Maybe we could target the specific crates that are windows compatible? E.g. cargo test -p krun-polly ....
BTW i'm gonna set the workflow up asap

PR #661, opened 2026-04-30
fix: drop unsafe host env vars when envp is NULL

ShivanshVij:
## Summary
This PR fixes a silent boot hang when `krun_set_exec`/`krun_set_env` is called with `envp = NULL`:

```c
const char *const envp[] = { NULL };
krun_set_exec(ctx, exec, argv, envp);   // explicit empty array - works
krun_set_exec(ctx, exec, argv, NULL);   // documented as "inherit host env" - silently hangs the boot
```

Every in-tree example, demo, and test happens to use the first form, so the bug never fired in CI even though the second form is documented behavior that external callers are supposed to be able to use.

The host-env-inheritance fallback (which triggers when `envp = NULL`) currently dumps environment variables onto the kernel cmdline with naive serialization (literal quotes, no whitespace or length filtering), which silently corrupt tokenization or pushes past the guest kernel's `COMMAND_LINE_SIZE`. `KRUN_INIT` gets truncated, init falls back to /bin/sh, boot hangs.

### Fixes
- New filter restricts inherited vars to printable ASCII without whitespace and drops individual vars that wouldn't fit a 1024-byte budget. The bogus literal quotes around values are also gone (init's `getenv("HOME")` no longer returns `"/root"`).
- Adds an `exec-null-envp` integration test as a regression guard, plus unit tests for the filter. Runner now counts `TestOutcome::Timeout` as a failure so a future hangs of this shape will cause CI to fail. 

## For Reviewers

I think that this PR has very few (if any) unintended side effects, except for this change: https://github.com/containers/libkrun/pull/661/changes#diff-30fb34448238a2b61c86dc96c1e58a7afa4af2fee6085b829d28464ad1e263ebR378

We're modifying the testing harness to fail on timeouts (it wasn't previously doing this). While I believe this change is correct (regardless of just for this PR), I did want to point it out as it may have unintended side effects. I've made sure all the tests pass locally.

### DIFF
--- a/src/libkrun/src/lib.rs
+++ b/src/libkrun/src/lib.rs
@@ -1303,6 +1303,45 @@ pub unsafe extern "C" fn krun_set_workdir(ctx_id: u32, c_workdir_path: *const c_
     KRUN_SUCCESS
 }
 
+/// Soft cap on the size of the env section inherited from the host process when
+/// the caller passes `envp = NULL` to `krun_set_exec`/`krun_set_env`. Linux
+/// genreally accepts only the first 2048 bytes so setting this value to 1024 leaves
+/// room for default options (KRUN_INIT, KRUN_RLIMITS, etc.).
+const MAX_INHERITED_ENV_BYTES: usize = 1024;
+
+/// Best-effort inheritance of the host process's environment for the guest's
+/// kernel cmdline. Vars whose key or value contain anything outside printable
+/// ASCII (excluding space) are dropped, since the Linux kernel cmdline parser
+/// tokenizes at whitespace and `Cmdline::valid_str` rejects non-printable
+/// chars. Each var is also dropped individually if appending it would push
+/// the total over `MAX_INHERITED_ENV_BYTES`, so a single oversized var (e.g.
+/// LS_COLORS) doesn't take out smaller vars iterated after it.
+fn inherit_host_env() -> String {
+    serialize_env(env::vars())
+}
+
+fn serialize_env<I: IntoIterator<Item = (String, String)>>(vars: I) -> String {
+    let mut buf = String::new();
+    for (k, v) in vars {
+        if k.is_empty()
+            || k.contains('=')
+            || !k.chars().all(|c| c.is_ascii_graphic())
+        {
+            continue;
+        }
+        if !v.chars().all(|c| c.is_ascii_graphic()) {
+            continue;
+        }
+
+        let entry = format!(" {k}={v}");
+        if buf.len() + entry.len() > MAX_INHERITED_ENV_BYTES {
+            continue;
+        }
+        buf.push_str(&entry);
+    }
+    buf
+}
+
 unsafe fn collapse_str_array(array: &[*const c_char]) -> Result<String, std::str::Utf8Error> {
     let mut strvec = Vec::new();
 
@@ -1318,7 +1357,6 @@ unsafe fn collapse_str_array(array: &[*const c_char]) -> Result<String, std::str
     Ok(strvec.join(" "))
 }
 
-#[allow(clippy::format_collect)]
 #[allow(clippy::missing_safety_doc)]
 #[no_mangle]
 pub unsafe extern "C" fn krun_set_exec(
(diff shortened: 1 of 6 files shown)


mtjhrc:
Thanks for the fix!

> We're modifying the testing harness to fail on timeouts (it wasn't previously doing this). While I believe this change is correct (regardless of just for this PR), I did want to point it out as it may have unintended side effects. I've made sure all the tests pass locally.

Yes that is a bug I noticed recently too, thanks for fixing it!

Please split the test change (especially the test framework changes) into separate commit(s). This makes `git log` way  more clear what happened and helps if we want to cherry pick changes into another branch. Also please write a proper commit message (something like the PR description). Also note we usually prefix the commit title with the subsystem/area it changes e.g. `tests: count timed-out tests as failed`

Btw we will likely be removing the env/arg passing mechanism via kernel cmdline for libkrun 2.0 (main branch now). We should fix this in the 1.18.x release (keep the PR open for main branch though, we will cherry pick it into 1.18). 

The fix itself looks good!

src/libkrun/src/lib.rs
```diff
@@ -1303,6 +1303,45 @@ pub unsafe extern "C" fn krun_set_workdir(ctx_id: u32, c_workdir_path: *const c_
     KRUN_SUCCESS
 }
 
+/// Soft cap on the size of the env section inherited from the host process when
+/// the caller passes `envp = NULL` to `krun_set_exec`/`krun_set_env`. Linux
+/// genreally accepts only the first 2048 bytes so setting this value to 1024 leaves
+/// room for default options (KRUN_INIT, KRUN_RLIMITS, etc.).
+const MAX_INHERITED_ENV_BYTES: usize = 1024;
```
dorindabassey, line 1310: about this constant, cmdline size limits vary for different archs, for example `riscv64` max size is `1024` bytes, `aarch64` is `2048` and so on, should `MAX_INHERITED_ENV_BYTES` be architecture-aware? Something like this if you're trying to leave room for default options:
```suggestion
#[cfg(target_arch = "riscv64")]
const MAX_INHERITED_ENV_BYTES: usize = 256;

#[cfg(target_arch = "aarch64")]
const MAX_INHERITED_ENV_BYTES: usize = 1024;

#[cfg(target_arch = "x86_64")]
const MAX_INHERITED_ENV_BYTES: usize = 8192;
```
Or is the 1024 limit intentional?

nit: s/genreally/generally

src/libkrun/src/lib.rs
```diff
@@ -1303,6 +1303,45 @@ pub unsafe extern "C" fn krun_set_workdir(ctx_id: u32, c_workdir_path: *const c_
     KRUN_SUCCESS
 }
 
+/// Soft cap on the size of the env section inherited from the host process when
+/// the caller passes `envp = NULL` to `krun_set_exec`/`krun_set_env`. Linux
+/// genreally accepts only the first 2048 bytes so setting this value to 1024 leaves
```
dorindabassey, line 1308: here it's not `Linux that has a 2048 limit` - the limit depends on the architecture

dorindabassey (commented):
Thanks for this PR, just a few comments.

src/libkrun/src/lib.rs
mtjhrc: @dorindabassey Keep in mind the `MAX_INHERITED_ENV_BYTES` has to be smaller than the kernel cmdline size though (because there are also other arguments to the kernel unrelated to this...).  Where did you get these values btw?

Also TBH, I wouldn't worry too much about this, we can never do this fully correctly (as the comment this PR added mentions it's best effort). In the long term we should stop using this to pass the arguments. That said its possible to "optimize" this to pick a better value. I'm just not sure It's good use of time to worry about this, this argument passing mechanism is inherently broken.

src/libkrun/src/lib.rs
mtjhrc: Yeah that is misleading, even if we keep a single limit (not platform specific), this comments needs to be updated.

PR #668, opened 2026-05-06
devices/fs: fix set_creds not switching credentials when CAP_SET{U,G}ID is available

nikromen:
When the VMM process runs as non-root but retains CAP_SETUID/CAP_SETGID (e.g. in a user namespace with --userns=keep-id), set_creds() skipped the credential switch for uid/gid 0 because the `uid == 0` check came before the capability check. This caused all guest root file operations to execute with the VMM's actual uid/gid, leading to EPERM on writes to root-owned directories.

Reorder the conditions so that when we have the capability to switch credentials, we always do so... including for uid/gid 0.

Reproducer (on fedora-43 x86):

podman run --rm --privileged --runtime krun --userns=keep-id fedora:latest dnf install -y htop

Fails with "Cannot open log file: (13) - Permission denied [/var/log/dnf5.log]". Any guest root file operation returns EPERM.

### DIFF
--- a/src/devices/src/virtio/fs/linux/passthrough.rs
+++ b/src/devices/src/virtio/fs/linux/passthrough.rs
@@ -836,26 +836,33 @@ impl PassthroughFs {
         gid: libc::gid_t,
     ) -> io::Result<(Option<ScopedUid>, Option<ScopedGid>)> {
         // Change the gid first, since once we change the uid we lose the capability to change the gid.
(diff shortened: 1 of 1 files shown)


jakecorrenti (commented):
Please make sure that you sign your commits. You can fix it by doing `git commit --amend -s`

slp:
Closing it as stale.

PR #670, opened 2026-05-07
Port the init binary code to Rust

### LINKED ISSUE #632: Rewrite init in Rust

jakecorrenti:
This PR ports the init binary code to Rust. It acts like any of the other crates that we have within the project.

To run the examples or with Podman, you would build the project as usual: `make BLK=1 NET=1 && sudo make BLK=1 NET=1 install` and continue with business as usual.

In my testing I've found the init binary to still remain small:
```
libkrun/init port-init ≡
❯ ll ../target/release/krun-init
.rwxr-xr-x@ 657k jcorrent  4 Jun 11:16 -I ../target/release/krun-init
```

NOTE: the AWS Nitro init binary is still written in C and will be ported next.

Fixes: #632


src/init_blob/build.rs
mtjhrc, line 1: Maybe this would also be a good opportunity to move this build.rs away from devices crate too.

Not sure what should it be called, maybe `init-blob`? I'm thinking it should literally be a crate that has 1 public constant (which is the init binary) and this `build.rs`. 

For now devices crate can depend on this `init-blob` as usual, but I plan to change that.  I may end up stacking multiple PRs on top of this which (depending on how long it will take to merge this), which need this to be a separate crate[^1], so it would really simplify the rebases for me. 

[^1]: I want to make the VMM crate depend on this `init-blob` and not the fs device itself (fs device will just receive a list of virtual files in constructor) this is in preparation for the 2.0 Rust API.

ggoodman: Sounds like #593 ;-)

jakecorrenti:
@nohajc if you could give this a try on FreeBSD that would be awesome :)

nohajc:
@jakecorrenti I'll see what I can do. :)

init/src/config.rs
```diff
@@ -0,0 +1,212 @@
...
+#[derive(Default)]
+pub struct Config {
+    pub argv: Option<Vec<String>>,
+    pub workdir: Option<String>,
+    #[cfg(target_os = "linux")]
+    pub tmpfs: Option<String>,
+}
+
+pub fn load(#[cfg(target_os = "linux")] is_mount_point: impl Fn(&str) -> bool) -> Config {
+    let path = env::var("KRUN_CONFIG").unwrap_or_else(|_| CONFIG_FILE_PATH.to_string());
```
nohajc: Already found an issue when I noticed in the github action log that `ISO_CONFIG_PATH` is unused. It needs to be used instead of `CONFIG_FILE_PATH` on FreeBSD.

nohajc (dismissed):
Currently, passing the command-line on FreeBSD is broken. See my comment.

This also reveals an issue in integration tests because you can see they TIMEOUT but the test run is marked as successful anyway. Not sure why...

init/src/config.rs
jakecorrenti: ~seems like things are running ok so far and added the fix you suggested~ timed out again, but not sure what's going on. i don't have a FreeBSD machine so i will have to figure it out

nohajc: I can try to play with it more tomorrow. Anyway, you should be able to run the full integration test including FreeBSD guest cases locally on either macOS or Linux. It downloads the sysroot and all other dependencies automatically.

nohajc: FreeBSD tests are skipped by default but they should run once init-freebsd is first compiled.

tests/test_cases/src/test_freebsd_gvproxy_tcp_guest_connect.rs
```diff
@@ -44,6 +44,10 @@ mod host {
             }
         }
 
+        fn timeout_secs(&self) -> u64 {
+            30
+        }
```
nohajc: I don't really buy Claude's explanation. The tests were consistently passing within the original time limit with C init.

jakecorrenti: I don't buy it either but I'm ruling it out

jakecorrenti: seems to be fixed. looks like it was a compilation issue

init/src/exec.rs
```diff
@@ -0,0 +1,134 @@
...
+        Ok(ForkResult::Parent { child }) => {
+            let code = loop {
+                match wait::waitpid(None, None) {
+                    Ok(WaitStatus::Exited(pid, c)) if pid == child => break c,
+                    Ok(WaitStatus::Signaled(pid, sig, _)) if pid == child => {
+                        break sig as i32 + 128;
+                    }
+                    _ => continue,
+                }
+            };
```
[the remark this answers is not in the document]
jakecorrenti: The EINTR case is already internally handled by nix

init/src/main.rs
```diff
@@ -0,0 +1,111 @@
...
+    env::apply_env();
+    env::apply_hostname();
+    env::apply_rlimits();
+
+    if let Some(workdir) = std::env::var("KRUN_WORKDIR").ok().or(cfg.workdir)
+        && let Err(e) = nix::unistd::chdir(workdir.as_str())
+    {
+        eprintln!("chdir to {workdir} failed: {e}");
+        std::process::exit(125);
+    }
```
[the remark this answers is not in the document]
jakecorrenti: NAK this is stable in the 2024 edition of rust.

init/src/exec.rs
```diff
@@ -0,0 +1,135 @@
...
+fn exec_workload(argv: &[String]) -> ! {
+    #[cfg(target_os = "linux")]
+    setup_redirects();
+    #[cfg(target_os = "freebsd")]
+    crate::freebsd::open_console();
+
+    let c_argv: Vec<CString> = argv
+        .iter()
+        .map(|s| CString::new(s.as_str()).expect("argv contains null byte"))
+        .collect();
```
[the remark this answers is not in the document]
jakecorrenti: NAK. argv can't be NULL in practice. this would fall under "No error handling for impossible scenarios" in the AGENTS.md

init/src/dhcp.rs
```diff
@@ -0,0 +1,626 @@
...
+
+    let addr_bytes = addr.to_ne_bytes();
+    add_rtattr(&mut buf, &mut msg_len, libc::IFA_LOCAL, &addr_bytes);
+    add_rtattr(&mut buf, &mut msg_len, libc::IFA_ADDRESS, &addr_bytes);
+
+    nl_send(nl_sock, &buf[..msg_len])?;
+    // Receive ACK
+    let recv_len = nl_recv(nl_sock, &mut buf)?;
+    nl_check_ack(&buf, recv_len, "mod_addr4")
+}
```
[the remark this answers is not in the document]
jakecorrenti: NAK we use u32 because that's what the DHCP packet contains.

init/src/dhcp.rs
```diff
@@ -0,0 +1,626 @@
...
+        &mut msg_len,
+        libc::RTA_GATEWAY,
+        &gateway.to_ne_bytes(),
+    );
+
+    nl_send(nl_sock, &buf[..msg_len])?;
+    // Receive ACK
+    let recv_len = nl_recv(nl_sock, &mut buf)?;
+    nl_check_ack(&buf, recv_len, "mod_route4")
+}
```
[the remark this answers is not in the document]
jakecorrenti: NAK we use u32 because that's what the DHCP packet contains.

init/src/dhcp.rs
```diff
@@ -0,0 +1,626 @@
...
+    let prefix_len = u32::from_be(netmask).leading_ones() as u8;
+
+    mod_addr4(nl_sock, iface_index, libc::RTM_NEWADDR, addr, prefix_len)
+        .context("add address from DHCP")?;
+    mod_route4(nl_sock, iface_index, libc::RTM_NEWROUTE, router)
+        .context("add default route from DHCP")?;
+    let _ = set_mtu(nl_sock, iface_index, mtu as u32);
+
+    Ok(())
+}
```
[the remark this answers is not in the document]
jakecorrenti: NAK we use u32 because that's what the DHCP packet contains.

init/src/dhcp.rs
```diff
@@ -0,0 +1,626 @@
...
+            let offered_addr = u32::from_ne_bytes(response[16..20].try_into().unwrap());
+            let server_addr = from.map(|a| a.ip().octets()).unwrap_or([0; 4]);
+
+            pkt.options = [0; DHCP_OPTIONS_SIZE];
+            let mut opts = DhcpOptionsWriter::new(&mut pkt.options);
+            opts.push(53, &[3]); // Request
+            opts.push(50, &offered_addr.to_ne_bytes()); // Requested IP
+            opts.push(54, &server_addr); // Server ID
+            opts.finish();
+
```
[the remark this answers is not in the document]
jakecorrenti: NAK. offered_addr is read as 4 bytes from the DHCP response and immediately written back as 4 bytes into a DHCP option. Abstracting into an Ipv4Addr adds no value

init/src/env.rs
```diff
@@ -0,0 +1,256 @@
...
+    // Set netmask to 255.255.255.0.
+    let mut sin: libc::sockaddr_in = unsafe { mem::zeroed() };
+    sin.sin_family = libc::AF_INET as libc::sa_family_t;
+    sin.sin_addr.s_addr = u32::from_ne_bytes(Ipv4Addr::new(255, 255, 255, 0).octets());
+    unsafe {
+        ifr.ifr_ifru.ifru_netmask = *(&sin as *const libc::sockaddr_in as *const libc::sockaddr);
+        if libc::ioctl(sock.as_raw_fd(), libc::SIOCSIFNETMASK as _, &ifr) < 0 {
+            eprintln!("Warning: dummy interface mask failed");
+        }
+    }
```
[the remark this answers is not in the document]
jakecorrenti: NAK. they produce identical results.

init/src/freebsd.rs
```diff
@@ -0,0 +1,170 @@
...
+    // stderr always mirrors stdout.
+    unsafe { libc::dup2(libc::STDOUT_FILENO, libc::STDERR_FILENO) };
+    // Close the log fd if it wasn't already consumed as stdout.
+    if out_raw >= 0 && out_raw != libc::STDOUT_FILENO {
+        drop(log_fd);
+    } else {
+        // Either no log fd, or it landed on STDOUT_FILENO — don't double-close.
+        std::mem::forget(log_fd);
+    }
+}
```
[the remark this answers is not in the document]
jakecorrenti: the suggested code would result in a use-after-free bug. The comment is valid, but the suggested change is not.

src/arch/Cargo.toml
```diff
@@ -24,9 +24,9 @@ utils = { package = "krun-utils", version = "=0.1.0-1.18.0", path = "../utils" }
 smbios = { package = "krun-smbios", version = "=0.1.0-1.18.0", path = "../smbios" }
 
 [target.'cfg(target_os = "linux")'.dependencies]
-kvm-bindings = { version = "0.12", features = ["fam-wrappers"] }
-kvm-ioctls = "0.22"
-tdx = { version = "0.1.0", optional = true }
+kvm-bindings = { version = "0.14", features = ["fam-wrappers"] }
```
slp, line 27: Does this PR really require bumping these versions? It's not wrong, and definitely something we'll need to do eventually, but feels out of scope and makes a little harder to review.

jakecorrenti: I remember running into issues where the versions of those crates prevented compilation due to version conflicts. I'll go back and try again to see if it was some leftover cached state causing problems

jakecorrenti: Ah, ok. I think I see what's going on. The Cargo.lock on the main branch vm-memory is pinned to v0.17.1 but when we add the init crate to the workspace it requires the lockfile to be regenerated. When the lockfile gets regenerated, v0.17 resolves to 0.17.2 (the newer version) which then breaks other deps (the ones that I also bumped)

slp (approved):
Good stuff here, thanks @jakecorrenti 

Let's get it merged it early in the release cycle so we can test it thoroughly.

PR #671, opened 2026-05-07
rutabaga_gfx/cross_domain: fix audio glitches by dropping write lock

aford173:
The previous write() held item_state.lock() across the write_volatile call that performs the actual syscall, serializing every cross-domain CMD_WRITE behind any other operation on the items table — including add_item from process_receive, which fires for every new fd received via SCM_RIGHTS as guests open additional channels.

For per-period eventfd wakeups (e.g. PipeWire audio streams signaling host playback every audio period), this means the write completes only after any in-flight item_state operation finishes. Under stream-create churn — many guest applications opening streams concurrently, each delivering new fds via SCM_RIGHTS that hit add_item under the same lock — the wait can exceed the audio period budget and produce missed-deadline glitches at the host's audio output.

This change shrinks the critical section to a brief fd dup() under the lock, performs the syscall lock-free, and only re-acquires the lock if hang_up indicates the item should be removed. In the common case (hang_up == 0, e.g. repeated eventfd wakeups for an active stream) the table is no longer touched per write, eliminating both the contention and the previous "remove + conditional re-insert" churn.

Behavioral changes vs the old code:
- Common case (hang_up == 0): item stays in the table; we hand out a dup'd fd for the write. Net behavior identical, lock hold time bounded by dup().
- hang_up == 1: item removed in a separate phase after the write, instead of "removed unconditionally then re-inserted on hang_up == 0". Same observed end state.
- Concurrent writes to the same id (no longer serialized): the kernel guarantees atomicity for eventfd writes (8 bytes) and pipe writes <= PIPE_BUF, which are the only two CrossDomainItem variants this branch handles. Each caller dup's its own fd and writes through it independently.

Verified with a synthetic reproducer: a sustained 1 kHz sine playing through a guest PipeWire stream while ~200 short-lived guest streams open and close concurrently (each issuing SCM_RIGHTS for new eventfds, contending with the sustained stream's per-period writes for item_state). Capturing the host sink monitor and counting sample-to-sample deltas exceeding a clean-sine threshold, the stress workload produced ~10 distinct glitch bursts in an 8-second capture before this change, and zero across five consecutive runs after.

### DIFF
--- a/src/rutabaga_gfx/src/cross_domain/mod.rs
+++ b/src/rutabaga_gfx/src/cross_domain/mod.rs
@@ -965,56 +965,47 @@ impl CrossDomainContext {
     }
 
     fn write(&self, cmd_write: &CrossDomainReadWrite, opaque_data: &[u8]) -> RutabagaResult<()> {
-        let mut items = self.item_state.lock().unwrap();
-
-        // Most of the time, hang-up and writing will be paired.  In lieu of this, remove the
-        // item rather than getting a reference.  In case of an error, there's not much to do
-        // besides reporting it.
-        let item = items
-            .table
-            .remove(&cmd_write.identifier)
-            .ok_or(RutabagaError::InvalidCrossDomainItemId)?;
-
         let len: usize = cmd_write.opaque_data_size.try_into()?;
-        match item {
-            CrossDomainItem::WaylandWritePipe(file) => {
-                if len != 0 {
-                    write_volatile(&file, opaque_data)?;
-                }
 
-                if cmd_write.hang_up == 0 {
-                    items.table.insert(
-                        cmd_write.identifier,
-                        CrossDomainItem::WaylandWritePipe(file),
-                    );
+        // Phase 1 (under lock): look up the item and dup() its fd. We release
+        // the lock before the actual write syscall so unrelated operations on
+        // item_state aren't serialized behind it. The previous design held
+        // the lock across write_volatile, which was the dominant contention
+        // source for per-period eventfd writes from active audio streams
+        // competing with concurrent stream-create churn (process_receive's
+        // add_item path also takes this lock). Cloning the fd is a cheap
+        // dup() syscall; the actual write — which can be longer for pipe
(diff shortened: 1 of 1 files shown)


valpackett:
Woah, that's quite the torture test scenario, did you ever actually experience stream-create churn in realistic regular desktop usage? :eyes: 

I agree that current code is locking too much but I'm wary of the `dup()`, turning `write()` into `dup()+write()+close()` is quite some overhead. If only this were all async code, the forwarding of each fd would've just been an independent task… sigh.

Maybe we could use a refcount instead? Would only add a pointer chase for derefing the `Rc` stored in the table, better than two context-switches churning kernel state.

---

BTW, keep in mind that for libkrun 2.0 the plan is to switch to upstream rutabaga (#560) so changes should be proposed there as well (or ideally, first)

aford173:
@valpackett - I never saw it in real life, but I was playing with Claude.  My end-game is run multiple instances of muvm on the same machine and have them share an audio pipeline, so I asked it to help craft a worst-case scenario.

slp:
> I agree that current code is locking too much but I'm wary of the `dup()`, turning `write()` into `dup()+write()+close()` is quite some overhead. If only this were all async code, the forwarding of each fd would've just been an independent task… sigh.

Async runtime thread management isn't free either.

slp (approved):
LGTM, thanks.

aford173:
@valpackett  / @slp - For future patches, I can run them through the rutabaga tree, but i was curious to know if the plan is to build libkrun with Rutabaga as a shared library or copy the rutabaga code into libkrun like it is now?

slp:
> @valpackett / @slp - For future patches, I can run them through the rutabaga tree, but i was curious to know if the plan is to build libkrun with Rutabaga as a shared library or copy the rutabaga code into libkrun like it is now?

The plan is to eventually consume https://github.com/magma-gpu/rutabaga_gfx as a crate. The main roadblock is still macOS support, but there's still time to get the fixed before 2.0

PR #673, opened 2026-05-11
Replace hardcoded init.krun with generic virtual file overlay

mtjhrc:
This PR replaces the hardcoded init.krun handling in the virtiofs passthrough backends with a generic virtual-files overlay (`AugmentFs`).

This introduces 2 new filesystem trait implementations:
- `AugmentFs<T>`, a wrapper that intercepts FUSE operations for virtual inodes - synthetic read-only files/directories backed by static data. It also handles our custom ioctls
- `NullFs`, a minimal FileSystem impl with just an empty root directory — used when no host directory is needed

The `init.krun` is registered as just a virtual file from the API layer. As a bonus you can even inject the `.krun_config.json` as a virtual file. 

Reimplemented `krun_set_root_disk_remount()` via `NullFs`+`AugmentFs` https://github.com/containers/libkrun/pull/551#issuecomment-3916446393

The public API is still mostly compatible. There are minor differences like `init.krun` dissapears after it has been looked up once. 

API breaking changes - applying `krun_disable_implicit_init()` and other `disable_implicit_*` will be applied by default in a follow up PR.

The init binary is now in its own init-blob crate. The direction for #634 (2.0 API) is to invert the dependency: init-blob would depend on libkrun's overlay APIs to inject itself, rather than libkrun depending on a specific init.

This supersedes #593 by @ggoodman, which tackled the same problem of decoupling init from the fs backends. This PR takes that idea further by removing awareness of init from the filesystem layer entirely - it's just another virtual file. #593 also introduced InitPolicy startup validation - how that fits into the 2.0 API (#634) with different payload types is still an open question.

### Known limitations / future work:
- Virtual inodes don't appear in readdir (pre-existing — init.krun was also lookup-only)
- The EXPORT_FD ioctl (GPU cross-domain shared memory) remains in passthrough for now
- No DAX (setupmapping) for virtual files on macOS (pre-existing — init.krun never had DAX on macOS either)
- DAX setupmapping/removemapping layering: the overlay creates mappings but the inner passthrough tears them down (works correctly but is architecturally messy)


ggoodman:
No comments on the code but I really love the direction!

jakecorrenti:
@mtjhrc do you want this to merge before #670 or after?

ggoodman:
As a thought, while you're implementing this, @mtjhrc, is there a reasonable opportunity to generalize this to supporting arbitrary virtual files (at least in a ro capacity)?

I actually have a use-case where I want a `procfs`-like filesystem provides by the host. The contents of this filesystem would all be programmatically-backed. Making it r/w sounds like a whole other beast but wanted to seed the idea just in case it could become a logical extension of what you're already hacking on.

mtjhrc:
> supporting arbitrary virtual files (at least in a ro capacity)?

Yes I thought about that, it would be pretty cool. But I would really want to first land some Rust API. And then we can just easily have a `VirtualFile` trait.  I don't think writable files would be that much harder, listing directories looked more tricky though (making the virtual files appear in `ls` mixed along real files).

mtjhrc:
@jakecorrenti 
>  @mtjhrc do you want this to merge before https://github.com/containers/libkrun/pull/670 or after?

Actually, I think I changed my mind we can merge this before #670, but if we do it after it doesn't matter either way.

jakecorrenti:
> @jakecorrenti
> 
> > @mtjhrc do you want this to merge before #670 or after?
> 
> Actually, I think I changed my mind we can merge this before #670, but if we do it after it doesn't matter either way.

I already did an initial rebase on top of your commits so let's just do this one first.

src/libkrun/src/lib.rs
```diff
@@ -2425,6 +2469,163 @@ pub unsafe extern "C" fn krun_set_root_disk_remount(
...
+    c_fs_tag: *const c_char,
+    c_path: *const c_char,
+    data: *const u8,
+    data_len: size_t,
+    mode: u32,
+    one_shot: bool,
+) -> i32 {
+    if c_fs_tag.is_null() || c_path.is_null() || data.is_null() || data_len == 0 {
+        return -libc::EINVAL;
+    }
```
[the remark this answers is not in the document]
mtjhrc: Makes sense, fixed

src/init-blob/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "init-blob"
+version = "0.1.0"
```
jakecorrenti: Should this follow the same versioning scheme as the other crates? i.e. `0.1.0-1.18.0`?

jakecorrenti:
### Code review

Found 5 issues:

1. `NullFs::lookup` returns `ENOSYS` instead of `ENOENT` — when a guest looks up any name not in the virtual overlay on a NullFs-backed filesystem, `AugmentFs::lookup` falls through to `NullFs::lookup`, which uses the default `FileSystem::lookup` returning `ENOSYS`. The Linux FUSE client treats `ENOSYS` on `FUSE_LOOKUP` as a protocol error rather than "file not found", causing unexpected failures for any non-virtual name lookup.

https://github.com/containers/libkrun/blob/e627484cd3f5bc7c4803093fee79cc707002363d/src/devices/src/virtio/fs/null_fs.rs#L20-L46

2. `aws-nitro` builds fail to compile — `init_virtual_entry()` is defined with `#[cfg(not(any(feature = "tee", feature = "aws-nitro")))]`, but all three call sites (`krun_set_root`, `krun_add_virtiofs3`, `krun_set_root_disk_remount`) are only gated with `#[cfg(not(feature = "tee"))]`. An `aws-nitro` build compiles all three callers but not the callee, causing a guaranteed link error.

https://github.com/containers/libkrun/blob/e627484cd3f5bc7c4803093fee79cc707002363d/src/libkrun/src/lib.rs#L94-L98

Call sites at [L620](https://github.com/containers/libkrun/blob/e627484cd3f5bc7c4803093fee79cc707002363d/src/libkrun/src/lib.rs#L619-L622), [L698](https://github.com/containers/libkrun/blob/e627484cd3f5bc7c4803093fee79cc707002363d/src/libkrun/src/lib.rs#L697-L700), [L2421](https://github.com/containers/libkrun/blob/e627484cd3f5bc7c4803093fee79cc707002363d/src/libkrun/src/lib.rs#L2420-L2422).

3. `AugmentFs::opendir` delegates unconditionally to the inner filesystem — it does not check whether the inode is virtual before calling `inner.opendir()`. When the inner is `PassthroughFs` or `PassthroughFsRo`, an `opendir` call on a virtual directory inode (registered via `krun_fs_add_overlay_dir`) returns `EBADF` because the inner filesystem has never seen that inode. The same applies to `releasedir`.

https://github.com/containers/libkrun/blob/e627484cd3f5bc7c4803093fee79cc707002363d/src/devices/src/virtio/fs/augment_fs.rs#L500-L509

4. `Fs::set_export_table` panics when `passthrough_cfg` is `None` — a NullFs device sets `passthrough_cfg: None`. If the user also enables a GPU device, the VMM calls `set_export_table` on all `Fs` devices and panics on the NullFs one via `.expect("export_table requires a passthrough filesystem")`.

https://github.com/containers/libkrun/blob/e627484cd3f5bc7c4803093fee79cc707002363d/src/devices/src/virtio/fs/device.rs#L104-L117

5. `VIRTIO_IOC_REMOVE_ROOT_DIR_REQ` (0x7603) is removed from `PassthroughFs::ioctl` but not handled in `AugmentFs::ioctl`, which only handles `VIRTIO_IOC_EXIT_CODE_REQ`. The `init/init.c` still issues this ioctl (to clean up the temporary root directory) when `krun_set_root_disk_remount` is used, and now receives `EOPNOTSUPP`, causing a spurious "Error removing temporary root directory: Operation not supported" error on every boot.

https://github.com/containers/libkrun/blob/e627484cd3f5bc7c4803093fee79cc707002363d/src/devices/src/virtio/fs/augment_fs.rs#L720-L747

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

jakecorrenti:
Ok, well the agent created the comment without my explicit approval... sorry about that.

src/init-blob/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "init-blob"
+version = "0.1.0-1.18.0"
```
jakecorrenti: This would be `0.1.0-1.18.1` now that the bugfix release came out.

src/init-blob/Cargo.toml
mtjhrc: Fixed

mtjhrc:
@jakecorrenti 
1. NullFs::lookup ENOSYS → ENOENT — Fixed. 
3. aws-nitro build failure — Fixed (hopefully?). Gated all virtiofs API functions, imports, and FsDeviceConfig::virtual_entries with not(aws-nitro).
4. AugmentFs::opendir — Intentionally out of scope. Virtual directory listing is not implemented — this is noted in the PR description under "Known limitations" (pre-existing: init.krun was also lookup-only). lookup+getattr work, which is sufficient for the current use cases: mount points and nested virtual file access (the integration test verifies reading /etc/nested/deep.txt through two levels of virtual dirs). Directory listing is future work.
5. set_export_table panic — Fixed. Replaced .expect() with let Some(...) else that skips NullFs devices gracefully — they don't participate in cross-domain fd export.
6. REMOVE_ROOT_DIR ioctl — Fixed. Removed the KRUN_REMOVE_ROOT_DIR_IOCTL call and #define from init.c entirely. With NullFs there's no host directory to clean up, so the ioctl is dead code.

src/devices/src/virtio/fs/linux/passthrough.rs
```diff
@@ -993,25 +986,7 @@ impl FileSystem for PassthroughFs {
...
-                generation: 0,
-                attr: st,
-                attr_flags: 0,
-                attr_timeout: self.cfg.attr_timeout,
-                entry_timeout: self.cfg.entry_timeout,
-            })
-        } else {
-            self.do_lookup(parent, name)
-        }
+        self.do_lookup(parent, name)
```
jakecorrenti: Has `do_lookup()` become an unnecessary abstraction at this point? It seems like we might be able to just have `lookup()` do what `do_lookup()` is doing and just `%s/do_lookup/lookup`?

src/devices/src/virtio/fs/augment_fs.rs
```diff
@@ -0,0 +1,737 @@
...
+
+use super::filesystem::{
+    Context, DirEntry, Entry, Extensions, FileSystem, FsOptions, GetxattrReply, ListxattrReply,
+    OpenOptions, SetattrValid, ZeroCopyReader, ZeroCopyWriter,
+};
+use super::fuse;
+use super::inode_alloc::InodeAllocator;
+use super::virtual_entry::{VirtualDirEntry, VirtualEntry, VirtualEntryContent, VIRTUAL_BLKSIZE};
+use crate::virtio::bindings;
+use crate::virtio::linux_errno;
```
jakecorrenti, line 34: non-blocking nit:

The import style here isn't consistent. You put the `std::sync` imports on multiple lines, but you use a single import for all the `super::filesystem` structs. It doesn't have to happen now, but maybe a TODO to make it more consistent would be good?

include/libkrun.h
```diff
@@ -1153,6 +1153,7 @@ int32_t krun_get_max_vcpus(void);
 */
 int32_t krun_split_irqchip(uint32_t ctx_id, bool enable);
 
+
```
jakecorrenti, line 1162: nit: random whitespace

include/libkrun.h
```diff
@@ -1180,6 +1180,56 @@ int32_t krun_disable_implicit_console(uint32_t ctx_id);
...
+ * Add a virtual overlay file to a virtiofs device.
+ *
+ * The file is backed entirely by host memory (no host file). The data
+ * pointer is NOT copied — the caller must keep the memory valid for the
+ * full VM lifetime.
+ *
+ * "path" may contain '/' to place the file inside a virtual directory
+ * previously created with krun_fs_add_overlay_dir (e.g. "etc/hostname").
+ * All intermediate directories must already exist; -ENOENT is returned
+ * if a component is missing, -ENOTDIR if a component is not a directory.
```
jakecorrenti, line 1219: I think the specific return values could also be mentioned in the `Returns` section? Sort of like how `krun_add_vhost_user_device` does it

include/libkrun.h
```diff
@@ -1180,6 +1180,56 @@ int32_t krun_disable_implicit_console(uint32_t ctx_id);
...
+                                 size_t data_len, uint32_t mode, bool one_shot);
+
+/**
+ * Add a virtual overlay directory to a virtiofs device.
+ *
+ * The directory is empty and read-only, useful as a mount point.
+ *
+ * "path" may contain '/' to nest inside an existing virtual directory
+ * (e.g. "usr/lib"). All intermediate directories must already exist;
+ * -ENOENT is returned if a component is missing, -ENOTDIR if a component
```
jakecorrenti, line 1247: Same comment about the errors here.

src/libkrun/src/lib.rs
```diff
@@ -19,8 +19,6 @@ use devices::virtio::fs::virtual_entry::{VirtualDirEntry, VirtualEntry, VirtualE
 use libc::{c_char, c_int, size_t};
 use once_cell::sync::Lazy;
 use polly::event_manager::EventManager;
-#[cfg(all(feature = "blk", not(feature = "tee")))]
-use rand::distr::{Alphanumeric, SampleString};
```
jakecorrenti, line 21: I think you can remove the `rand` dependency from the src/libkrun/Cargo.toml now?

jakecorrenti (changes requested):
Only blocking change here is the init-blob package version. the others aren't blockers and are nits

src/devices/src/virtio/fs/augment_fs.rs
mtjhrc: Hmm actually I think after #686 this would have failed CI - cargo fmt (2024 edition) would want format this nicely.

jakecorrenti:
Just the comment about the package version in `src/init-blob/Cargo.toml` and this LGTM

jakecorrenti (approved):
LGTM!

lstocchi:
Just gave a look as this affects the windows part as well. LGTM. Interesting stuff!! 🚀

jakecorrenti:
Merging so that we can get the init binary and API work forward

d-e-s-o:
Is there any chance we could get https://github.com/containers/libkrun/pull/673/changes/0127980f04489ed3e0a5c5f2aa4e784aec884a33 into the [`stable-1.18.x` branch](https://github.com/containers/libkrun/tree/stable-1.18.x) (and thereby a published release)? The change eliminates the need for setting `KRUN_INIT_BINARY_PATH` when consuming the `crates.io`-published version and would be tremendously useful to have, as it unblocks publishing of dependent crates themselves to the registry. I could also look into creating a slimmed down variant if need be. cc @slp

(may also be good to add a few lines to how branching works in the repository to the README or CONTRIBUTING.md)

slp:
> Is there any chance we could get [0127980](https://github.com/containers/libkrun/commit/0127980f04489ed3e0a5c5f2aa4e784aec884a33) into the [`stable-1.18.x` branch](https://github.com/containers/libkrun/tree/stable-1.18.x) (and thereby a published release)? The change eliminates the need for setting `KRUN_INIT_BINARY_PATH` when consuming the `crates.io`-published version and would be tremendously useful to have, as it unblocks publishing of dependent crates themselves to the registry. I could also look into creating a slimmed down variant if need be. cc @slp

Sounds doable.

PR #676, opened 2026-05-13
Add AGENTS.md agent file

jakecorrenti:
Adds an AGENTS.md file to give agents context about libkrun. This
includes project structure and coding standards.

Create a .claude/ directory that points the Claude Code agent to the AGENTS.md file since
that file isn't supported.

This was initially generated by running /init inside of Claude Code. I took some sections from
https://github.com/multica-ai/andrej-karpathy-skills/blob/main/CLAUDE.md?plain=1 and
https://github.com/openai/openai-agents-python/blob/f20aa40f56ec374841dfe6e4f3d6118c8e32fb2b/AGENTS.md.

This should be a living document that we adjust and modify as time goes on to better improve the agentic workflow.

I would love to get some discussion going about a couple things:
- What would people like to add
- What would people like to remove
- How is the formatting
- How do we handle multiple agent files? Claude doesn't support AGENTS.md, but we could have one AGENTS.md that we symlink to from the other files.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,141 @@
+# AGENTS.md
+
+This file provides guidance to Agents when working with code in this repository.
+
+## What this project is
+
+**libkrun** is a Rust dynamic library that provides virtualization-based process isolation via KVM (Linux) and HVF (macOS). It exposes a C API (`include/libkrun.h`) that lets callers run a process inside a lightweight VM with configurable vCPUs, RAM, virtio devices, and an embedded init binary.
+
+## Build commands
+
+All builds go through the Makefile, which handles feature flags, platform detection, and sysroot management for cross-compilation. Direct `cargo build` skips that plumbing.
+
+```bash
+# Release build (minimal — no optional devices)
+make
+
+# Release build with common optional devices
+make BLK=1 NET=1 GPU=1 SND=1 INPUT=1
+
+# Debug build
+make debug
+
+# TEE variants (mutually exclusive with each other and GPU/SND/INPUT)
+make SEV=1        # AMD SEV — produces libkrun-sev.so
+make TDX=1        # Intel TDX — produces libkrun-tdx.so
+
+# Other optional features
+make VHOST_USER=1
+make VIRGL_RESOURCE_MAP2=1
+
+# Install to /usr/local (or PREFIX=...)
+make install
+make PREFIX=$HOME/.local install
+```
+
+The Makefile exports `CC_LINUX` for Rust build scripts that compile C code targeting Linux. On macOS it auto-downloads a Debian sysroot; on Linux it uses the host toolchain.
+
+## Lint and format
+
+Clippy is run with `-D warnings` — zero warnings are allowed. The CI checks several feature combinations; you should too when touching device or feature-gated code:
+
+```bash
+# Required before every PR — same as CI
+cargo clippy --locked -- -D warnings
+cargo clippy --locked --features amd-sev -- -D warnings
+cargo clippy --locked --features tdx -- -D warnings
+cargo clippy --locked --features net,blk,gpu,snd,input -- -D warnings
+
+# Format check
+cargo fmt -- --check
+```
+
+## Tests
+
+```bash
+# Unit tests (requires init/init to exist and KVM access on Linux)
+touch init/init
+cargo test
+
+# Integration tests (builds and installs the library first)
+make test
+make test TEST=test_name       # run a single integration test
+make test BLK=1                # integration tests with blk feature
+```
+
+Integration tests live in `tests/` as a separate Cargo workspace. They require the library to be installed to a local prefix (`test-prefix/`) which `make test` handles automatically.
+
+## Crate architecture
+
+The workspace (`Cargo.toml`) contains these crates under `src/`:
+
+| Crate | Role |
+|---|---|
+| **libkrun** | C API surface. Manages `KrunContext` instances (one per VM), translates API calls into VMM configuration, and drives startup via `krun_start_enter()`. |
+| **vmm** (krun-vmm) | VMM core: builds and runs the VM, owns the vCPU threads, wires devices to memory and IRQs. |
+| **devices** (krun-devices) | All virtio device implementations: console, block, fs (virtiofs passthrough + read-only wrapper), net, gpu, sound, input, vsock. |
+| **arch** (krun-arch) | Platform-specific VM setup: GDT/IDT (x86_64), FDT (aarch64/riscv), boot protocol. |
+| **kernel** | Loads the kernel image and sets up the boot parameters passed to the VMM. |
+| **hvf** | macOS HVF hypervisor bindings (Linux uses kvm-ioctls directly in vmm). |
+| **cpuid** | x86_64 CPUID leaf manipulation for vCPU feature exposure. |
+| **rutabaga_gfx** | Wraps virglrenderer for virtio-gpu. |
+| **display** / **input** | Host-side display and input backends (used by the gpu feature). |
+| **utils**, **polly**, **smbios**, **arch_gen** | Shared utilities, event loop, SMBIOS table generation, architecture codegen. |
+| **aws_nitro** | AWS Nitro Enclave support. |
+
+### How a VM starts
+
+1. Caller invokes `krun_create_ctx()` → allocates a `KrunContext`
+2. Caller configures it (vCPUs, RAM, disks, network, exec path, etc.)
+3. `krun_start_enter()` calls into **vmm**, which:
+   - Loads the kernel via **kernel**
+   - Instantiates virtio devices from **devices**
+   - Starts vCPU threads (KVM ioctls on Linux, HVF on macOS)
+   - The guest boots, the C init binary runs as PID 1, reads `.krun_config.json` from the virtiofs overlay, and execs the workload
+
+### The init binary (`init/init.c`)
+
+The guest PID-1 is a statically-linked C binary (`init/init.c`) compiled by `src/devices/build.rs` via `CC_LINUX`. The compiled binary path is set in the `KRUN_INIT_BINARY_PATH` env var at build time and embedded into the devices crate via `include_bytes!`. The passthrough fs backend exposes it as a virtual read-only file named `init.krun` (inode defined in `src/devices/src/virtio/fs/linux/passthrough.rs`) — the real host filesystem never sees it.
+
+AWS Nitro uses a separate C init (`init/aws-nitro/`) built by the Makefile when `AWS_NITRO=1`.
+
+### Feature flags
+
+Features are additive and controlled at the `libkrun` crate level. Each device feature (`blk`, `net`, `gpu`, `snd`, `input`) enables the corresponding code in both `devices` and `vmm`. The TEE variants (`amd-sev`, `tdx`) imply `blk` + `tee` and affect the soname of the output library.
+
+## Pull Request expectations
+- New tests are added when necessary.
+- Documentation is added or updated when necessary.
+- Linting and formatting has been done.
+- All of the tests pass.
+- The commit structure follows what's described in [Commit structure](### Commit structure).
+
+## Code Quality
+
+### Commit structure
+
+- Format: `<subsystem>: <title>` — e.g., `virtio/blk: add print_text() function`
+- Commits must be self-contained, compile, and pass tests independently
+- Sign all commits with `git commit -s` (DCO requirement)
+- Agent attribution format: `<Agent-tool>: <model-name>` - e.g., `Claude Code: sonnet-4.6`, `Cursor: codex-5.3`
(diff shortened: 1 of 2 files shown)


jakecorrenti:
I've also been using the [Superpowers](https://claude.com/plugins/superpowers) Claude plugin which has been useful.

I also want to experiment with Personas to see if that has any workflow improvement as well. If anyone has any other cool plugins/configurations I'd be interested to hear about those here as well.

CLAUDE.md
```diff
@@ -0,0 +1,194 @@
...
+- "Refactor X" -> "Ensure tests pass before and after"
+
+For multi-step tasks, state a brief plan:
+```
+1. [Step] -> verify: [check]
+2. [Step] -> verify: [check]
+3. [Step] -> verify: [check]
+```
+
+Strong success criteria that let you loop independently. Weak criteria ("make it work") require constant clarification.
```
mtjhrc: This genuinely looks interesting, I definitely have to try this!

However, I think this is an example of what I was getting at in my email regarding scope. To me, a committed project-level CLAUDE.md or AGENTS.md should focus strictly on the codebase: architecture overviews, coding conventions, how to run tests, etc. In other words I think this file should mostly be applicable for humans working on libkrun too.

Personal agent workflows, prompt loops, and tool preferences like this feel like they belong in a global, user-level CLAUDE.md - much like personal shell aliases in a dotfiles repo. It's super valuable to share, but probably shouldn't be hardcoded into the upstream project. (And Claude should still pick it up perfectly fine from your global config).

CLAUDE.md
```diff
@@ -0,0 +1,194 @@
...
+- **Queues**: Don't create empty queues - pass existing ones through. Empty queues -> nothing received.
+- **Files**: Don't create dummy files for `include_bytes!()`. Ask the user how to build/obtain them.
+
+**Principle**: When something is wired into the system, propagate it. Don't create disconnected instances. If a build fails due to missing dependencies, ask the user rather than stubbing.
+
+## Platform support
+
+- **Linux x86_64 / aarch64**: primary targets, full feature support
+- **macOS aarch64**: supported (HVF), GPU works
+- **FreeBSD**: experimental cross-compilation target (`make BUILD_BSD_INIT=1` for the FreeBSD init)
```
mtjhrc: This is a bit misleading. We need to differentiate between host platforms libkrun runs on (Linux, macOS) and guests libkrun supports (Linux, FreeBSD, possibly others but they are not tested).

CLAUDE.md
```diff
@@ -0,0 +1,194 @@
...
+
+The Makefile exports `CC_LINUX` for Rust build scripts that compile C code targeting Linux. On macOS it auto-downloads a Debian sysroot; on Linux it uses the host toolchain.
+
+## Lint and format
+
+Clippy is run with `-D warnings` — zero warnings are allowed. The CI checks several feature combinations; you should too when touching device or feature-gated code:
+
+```bash
+# Required before every PR — same as CI
+touch init/init   # build scripts require this file to exist
```
mtjhrc: This kind of conflicts with this section:

```
**Files**: Don't create dummy files for `include_bytes!()`. Ask the user how to build/obtain them.
```

I had a problem with Claude doing touch `touch init/init` after `make clean` and then getting **very** confused why the **integration** tests don't work anymore . So that is why added that section. I don't currently have `touch init/init` mentioned at all, I just let it include the real init.

AGENTS.md
```diff
@@ -0,0 +1,194 @@
...
+
+### The init binary (`init/init.c`)
+
+The guest PID-1 is a statically-linked C binary (`init/init.c`) compiled by `src/devices/build.rs` via `CC_LINUX`. The compiled binary path is set in the `KRUN_INIT_BINARY_PATH` env var at build time and embedded into the devices crate via `include_bytes!`. The passthrough fs backend exposes it as a virtual read-only file named `init.krun` (inode defined in `src/devices/src/virtio/fs/linux/passthrough.rs`) — the real host filesystem never sees it.
+
+AWS Nitro uses a separate C init (`init/aws-nitro/`) built by the Makefile when `AWS_NITRO=1`.
+
+### The virtiofs read-only wrapper
+
+`src/devices/src/virtio/fs/read_only.rs` wraps the passthrough backend and rejects all mutating FUSE operations with `EROFS`. When adding new `FileSystem` trait methods, explicitly handle them in this wrapper — unoverridden methods fall back to `ENOSYS`, which is safe but may not give the right error semantics.
```
mtjhrc: This is perhaps fine for now, but it seems kind of random in this file. Maybe put we need to have like a docs directory and just list topics here pointing to several markdown files. Wdyt? 

Also this specific issue, it seems like it should be addressed in the code (via comment or refactoring it).

CLAUDE.md
```diff
@@ -0,0 +1,194 @@
...
+- Commit messages should be concise and written in the imperative mood. Small, focused commits are preferred.
+
+### Think before coding
+Don't assume. Don't hide confusion. Surface tradeoffs.
+
+Before implementing:
+- State your assumptoins explicitely. If uncertain, ask.
+- If multiple interpretatoins exist, present them - don't pick silently.
+- If a simpler approach exists, say so. Push back when warranted.
+- If something is unclear, stop. Name what's confusing. Ask.
```
mtjhrc: Same as the other comment, this looks useful, but doesn't look like something that should go into the project's `CLAUDE.md`/`AGENTS.md` IMO.

mtjhrc (changes requested):
The project overview and information about running tests is definitely something we should have in a commited  AGENTS.md and looks useful! 

 I am not a fan of committing some the general Claude "workflow" sections here however.

Also imagine someone want to override them, because they don't like them for some reason or it doesn't work well with their LLM, then it will become super annoying for them to not be able to use the project provided AGENTS.md at all and having to configure something to ignore them.  

> How do we handle multiple agent files? Claude doesn't support AGENTS.md, but we could have one AGENTS.md that we symlink to from the other files.

I would strongly prefer `AGENTS.md` and point CLAUDE.md to it.  I think you can just put `@AGENTS.md` into a CLAUDE.md file and it would work.

CLAUDE.md
```diff
@@ -0,0 +1,194 @@
...
+- No abstractions for single-use code.
+- No "flexibility" or "configurability" that wasn't requested.
+- No error handling for impossible scenarios.
+- If you write 200 lines and it could be 50, rewrite it.
+- avoid checking for empty input when calling a function if the function already handles the base case well (eg. empty input is noop)
+
+Ask yourself: "Would a senior engineer say this is overcomplicated?" If yes, simplify.
+
+### Surgical changes
+Touch only what you must. Clean up only your own mess.
```
jakecorrenti: @mtjhrc where do we stand on these sections as well? These seem like personal workflow sections as well.

CLAUDE.md
jakecorrenti: Yep good catch, thanks!

AGENTS.md
jakecorrenti: Yeah I think that documenting broader topics that may not be intuitive is a good idea.

I'll remove the section and add a comment in another PR

CLAUDE.md
jakecorrenti: Fixed! Thanks.

CLAUDE.md
mtjhrc: Yeah I suppose we should mostly remove this. Though perhaps these could fall under coding guidelines:
`- No error handling for impossible scenarios.`
`- avoid checking for empty input when calling a function if the function already handles the base case well (eg. empty input is noop)`

jakecorrenti: Added.

AGENTS.md
```diff
@@ -0,0 +1,151 @@
...
+
+### Rust coding style
+- Use `use` imports instead of inline full paths. One level of qualitifation is fine when it clarifies what something is (e.g., `log::trace!`, `fs::read_to_string()`), but don't use longer paths like `std::process::Command::new(...)` or `crate::foo::bar::baz()` -- import with `use` instead.
+
+Creating stubs to make code compile causes silent runtime failures:
+- **EventFd**: Don't create new `EventFd::new()` - propagate existing ones (via `Arc<EventFd>` or `.try_clone()`). New `EventFd`s are never signaled -> deadlock.
+- **Queues**: Don't create empty queues - pass existing ones through. Empty queues -> nothing received.
+- **Files**: Don't create dummy files for `include_bytes!()`. Ask the user how to build/obtain them.
+
+**Principle**: When something is wired into the system, propagate it. Don't create disconnected instances. If a build fails due to missing dependencies, ask the user rather than stubbing.
```
mtjhrc: So this section comes from my CLAUDE.md, but I am not sure if we should include it fully verbatim in the committed version. 
  
This is also perhaps bit of workflow thing too. Though it's also a bit specific to things I found Claude (at least in the past) struggle with in libkrun. But if you test the code you submit you, you would notice your change doesn't actually work and just deadlocks immediately.

The more I think about this I realize there is sort of a tradeoff here. If we put a lot of these guidelines here it could be annoying for people that have their own guidelines like this. However it would be really beneficial for people with default agents configuration especially if they are inexperienced or don't notice the issues produced by the agent - I can see this improving the quality of random drive-by PRs.

Maybe a middle-ground could be to keep this section but shorten it?

mtjhrc: @jakecorrenti Have you noticed Claude or other LLMs also struggle with this? If not maybe we should hold off on this for now?

jakecorrenti: I don't think I've run into that personally. The one thing I've been running into a lot is the agent basically liking to hear itself talk. i.e. 
```rust
fn foo() {
    bar();
}

fn bar() {
   foobar();
}

fn foobar() {
    println!("Hello, world!");
}
```

lstocchi:
@jakecorrenti have you read this [best practice doc](https://gitlab.cee.redhat.com/global-engineering/wg-agentic-sdlc/-/blob/main/best-practices/repo-scaffolding/README.md)? There are some interesting tips related to the AGENTS file.

jakecorrenti:
> @jakecorrenti have you read this [best practice doc](https://gitlab.cee.redhat.com/global-engineering/wg-agentic-sdlc/-/blob/main/best-practices/repo-scaffolding/README.md)? There are some interesting tips related to the AGENTS file.

I hadn't seen that before. Looking at it now. Thanks!

AGENTS.md
mtjhrc: > I don't think I've run into that personally. The one thing I've been running into a lot is the agent basically liking to hear itself talk.

Not sure what you meant there exactly.

Regarding this section I would be in favor of dropping it (It's common sense really, nothing special about libkrun).

mtjhrc:
Also @jakecorrenti you should probably update the commit message, it doesn't reflect the `CLAUDE.md` -> `AGENTS.md` change.

AGENTS.md
jakecorrenti: I dropped it.

What I meant in the comment above is the only persistent issue I had (before creating the CLAUDE.md) was the agent would go like three levels down the call stack to call a function that could have just been called directly.

jakecorrenti:
@mtjhrc PTAL when you get a chance

AGENTS.md
```diff
@@ -0,0 +1,142 @@
...
+- Use `use` imports instead of inline full paths. One level of qualitifation is fine when it clarifies what something is (e.g., `log::trace!`, `fs::read_to_string()`), but don't use longer paths like `std::process::Command::new(...)` or `crate::foo::bar::baz()` -- import with `use` instead.
+
+## Platform support
+
+### Host platforms (where libkrun runs as a library)
+
+- **Linux x86_64 / aarch64**: primary targets, full feature support (KVM)
+- **Linux riscv64**: experimental (KVM)
+- **macOS aarch64**: supported (HVF), GPU works
+- **FreeBSD**: experimental cross-compilation target (`make BUILD_BSD_INIT=1`)
```
mtjhrc: The title of this section says "(where libkrun runs as a library)", but we don't run on FreeBSD. I would suggest just removing this line, I feel like this would cause more confusion than it solves.

AGENTS.md
```diff
@@ -0,0 +1,142 @@
...
+- The commit structure follows what's described in [Commit structure](### Commit structure).
+
+## Code Quality
+
+### Commit structure
+
+- Format: `<subsystem>: <title>` — e.g., `virtio/blk: add print_text() function`
+- Commits must be self-contained, compile, and pass tests independently
+- Sign all commits with `git commit -s` (DCO requirement)
+- Agent attribution: `Claude Code: <model-name>`
```
mtjhrc: So this says `Claude Code` rather explicitly, this may confuse other agents that people use to say they are "Claude Code".

jakecorrenti: Adjusted a bit, let me know how that looks.

AGENTS.md
jakecorrenti: Removed. Thanks.

mtjhrc (approved):
LGTM now!

PR #677, opened 2026-05-13
README: add security email section

jakecorrenti:
Add a section telling users to report security issues to libkrun-security@redhat.com instead of the GH issue tracker or Matrix.

The email is configured such that only the core maintainers are able to see messages and mailing list members. Outside parties can **only** send messages.


mtjhrc (approved):
LGTM, thanks for setting this up!

PR #679, opened 2026-05-14
init: fix unescape_string() not advancing past escape characters

### LINKED ISSUE #678: quoting issues for environment variables passed from podman
I am seeing quoting issues for environment variables where newlines and quotes `"` are getting mangled somehow. See the output below.

```
[bound] -bash-5.3$ rpm -q podman crun crun-krun libkrun libkrunfw
podman-5.8.2-1.fc44.x86_64
crun-1.27.1-1.fc44.x86_64
crun-krun-1.27.1-1.fc44.x86_64
libkrun-1.17.4-1.fc44.x86_64
libkrunfw-5.3.0-1.fc44.x86_64
[bound] -bash-5.3$ podman pull --quiet quay.io/fedora/fedora:44
a01ae4eb9c82a895c1e10f58ff011a8dcfca2b68b88d44979871a73d40184f85
[bound] -bash-5.3$ 
[bound] -bash-5.3$ export JSON_VAR='
{
    "var": "value",
    "var2": "value2",
}'
[bound] -bash-5.3$ podman run -it --rm -e JSON_VAR quay.io/fedora/fedora:44 echo "$JSON_VAR"

{
    "var": "value",
    "var2": "value2",
}
[bound] -bash-5.3$ 
[bound] -bash-5.3$ podman run -it --rm --runtime=krun -e JSON_VAR quay.io/fedora/fedora:44 echo "$JSON_VAR"

n{
n    ""var"": ""value"",
n    ""var2"": ""value2"",
n}
```

dustymabe:
The unescape_string() function in init.c, which handles JSON escape sequences when parsing environment variables from .krun_config.json, had a bug where the pointer 'val' was not advanced past the escape character after processing it.

When encountering a two-character JSON escape sequence like \n or \", the switch statement pre-increments val to point at the escape character (e.g., 'n' or '"') and writes the unescaped byte to the output. However, it never advances val past that character. On the next loop iteration, the character is not a backslash, so it gets copied again as a literal character.

This causes:
 - \n (JSON-escaped newline) to produce a newline followed by a literal 'n'
 - \" (JSON-escaped double quote) to produce two double quotes

For example, an environment variable set to a JSON string like:

  {\"key\": \"value\"}

would be rendered inside the krun VM as:

  {""key"": ""value""}

Fix this by adding val++ after writing the unescaped character in each case of the switch statement. The 'u' (unicode) case already handles its own pointer arithmetic and is not affected.

Fixes: https://github.com/containers/libkrun/issues/678

### DIFF
--- a/init/init.c
+++ b/init/init.c
@@ -641,27 +641,35 @@ static void unescape_string(char *string, int len)
         switch (*++val) {
         case 'n':
             string[i++] = '\n';
+            val++;
             break;
         case 't':
             string[i++] = '\t';
+            val++;
             break;
         case 'r':
             string[i++] = '\r';
+            val++;
             break;
         case 'b':
             string[i++] = '\b';
+            val++;
             break;
         case 'f':
             string[i++] = '\f';
+            val++;
             break;
         case '\\':
             string[i++] = '\\';
+            val++;
             break;
         case '\"':
             string[i++] = '\"';
+            val++;
             break;
         case '/':
             string[i++] = '/';
+            val++;
             break;
         case 'u': {
             const char *unescaped = "?";


dustymabe:
Please add quite a bit of scrutiny here. I definitely don't know this code base. 

In my local test this fixes at least the problem from #678:

```
[dustymabe@hattop ~]$ export JSON_VAR='
{
    "var": "value",
    "var2": "value2",
}'
[dustymabe@hattop ~]$ podman run -it --rm -e JSON_VAR quay.io/fedora/fedora:44 echo "$JSON_VAR"
Trying to pull quay.io/fedora/fedora:44...
Getting image source signatures
Copying blob dfa92e469035 done   | 
Copying config a01ae4eb9c done   | 
Writing manifest to image destination

{
    "var": "value",
    "var2": "value2",
}
[dustymabe@hattop ~]$ podman run -it --rm --runtime=krun -e JSON_VAR quay.io/fedora/fedora:44 echo "$JSON_VAR"

{
    "var": "value",
    "var2": "value2",
}
```

slp (approved):
LGTM, thanks @dustymabe !

PR #680, opened 2026-05-14
devices/vsock: support multi-descriptor TX chains

slp:
from_tx_virtq_head assumed a fixed 2-descriptor layout (header + one data descriptor), which breaks with newer kernels that may combine header and data in a single descriptor (Linux 6.2+) or split data across multiple descriptors.

Handle three cases: single combined descriptor (zero-copy), classic two-descriptor (zero-copy, unchanged), and multi-descriptor data (copied into an owned contiguous buffer). The RX path already handled the combined case; this brings the TX path to parity and beyond.

### DIFF
--- a/src/devices/src/virtio/vsock/packet.rs
+++ b/src/devices/src/virtio/vsock/packet.rs
@@ -191,6 +191,7 @@ pub struct VsockPacket {
     hdr: *mut u8,
     buf: Option<*mut u8>,
     buf_size: usize,
+    owned_buf: Option<Vec<u8>>,
 }
 
 fn get_host_address<T: GuestMemory>(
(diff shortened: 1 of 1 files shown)


mtjhrc:
Just a nitpick, FYI the `Assisted-by: <anthropic/claude-opus-4.6>` tag doesn't match the `CONTRIBUTING.md` format. 


`anthropic` is not a name of a coding agent harness, (did you mean `Claude Code` or perhaps `OpenCode`?) 
Also according to the merged `CONTRIBUTING.md` the format should be `Assisted-by: Claude Code:claude-opus-4.6` I suppose. (yeah the space is weird, but it is technically called "Claude Code" with a space...)

Maybe I'll try creating a CI check for this...

nohajc:
@mtjhrc I was amending one of my commits to use the prescribed format and there's actually an issue that sometimes you might not know what the model was.

For example Copilot only puts the harness name to its own commits and it's possible to use model "auto". I'm not sure if you can always retrieve information about the exact model used in that case.

lstocchi:
> Maybe I'll try creating a CI check for this...

My .02 cents ... I wonder if we should be a bit elastic around the naming and avoid adding a deep CI check.
The RH guidelines ([here](https://source.redhat.com/projects_and_programs/ai/wiki/code_assistants_guidelines_for_responsible_use_of_ai_code_assistants#6-mark-code-with-substantial-ai-generated-portions-)) just mention we should use the labels `Assisted-by`/`Generated-by` but then there is no a strict rule how the value must be formatted. I see in some projects there is a preference to use `tool (model)` (e.g. `Assisted-by: OpenCode (Claude Opus 4.6)`), in others just `model (context)` (e.g. `Claude Opus 4.6 (1M context)`). Also, as said by @nohajc you may not know exactly the model you used. In some cases i saw `Co-authored-by: Claude <noreply@anthropic.com>` without any model mention.

Idk if you received other direction but AFAIK for statistical measurements they're just detecting the `Assisted-by` label.

slp:
> Just a nitpick, FYI the `Assisted-by: <anthropic/claude-opus-4.6>` tag doesn't match the `CONTRIBUTING.md` format.
> 
> `anthropic` is not a name of a coding agent harness, (did you mean `Claude Code` or perhaps `OpenCode`?) Also according to the merged `CONTRIBUTING.md` the format should be `Assisted-by: Claude Code:claude-opus-4.6` I suppose. (yeah the space is weird, but it is technically called "Claude Code" with a space...)

Fixed, thanks!

src/devices/src/virtio/vsock/packet.rs
```diff
@@ -346,11 +420,15 @@ impl VsockPacket {
...
-            unsafe { std::slice::from_raw_parts_mut(ptr, self.buf_size) }
-        })
+        if let Some(ref mut owned) = self.owned_buf {
+            Some(owned.as_mut_slice())
+        } else {
+            self.buf.map(|ptr| {
+                // This is safe since bound checks have already been performed when creating the
+                // packet from the virtq descriptor.
+                unsafe { std::slice::from_raw_parts_mut(ptr, self.buf_size) }
+            })
```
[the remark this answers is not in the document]
slp: This is just a temporary solution for 1.x, it will be replaced with zero-copy abstractions for 2.0.

slp:
@mtjhrc PTAAL, I'd like to have this one in 1.18.1.

src/devices/src/virtio/vsock/packet.rs
```diff
@@ -237,25 +239,92 @@ impl VsockPacket {
...
+                let src = get_host_address(desc.mem, desc.addr, desc.len as usize)
+                    .map_err(VsockError::GuestMemoryMmap)?;
+                owned_buf.extend_from_slice(unsafe {
+                    std::slice::from_raw_parts(src as *const u8, desc.len as usize)
+                });
+            }
+            next = desc.next_descriptor();
+        }
+
+        if (owned_buf.len() as u32) < pkt.len() {
```
mtjhrc: There is not really a bug here currently as far as I can tell, but the other way around is safer:
```suggestion
        if owned_buf.len() < (pkt.len() as usize) {
```

Btw we should eventually enable `clippy::cast_possible_truncation` IMO.

src/devices/src/virtio/vsock/packet.rs
```diff
@@ -237,25 +239,92 @@ impl VsockPacket {
             return Err(VsockError::InvalidPktLen(pkt.len()));
         }
```
mtjhrc: This is pre-exisitng but the newly added code continues to use this broken pattern, `pkt.len()` technically reads length from guest controlled memory, so there are many TOCTOU bugs possible here (though the repetetive reads are very likely optimized out by the compiler, since this doesn't use volatile reads).

src/devices/src/virtio/vsock/packet.rs
slp: Fixed, thanks!

src/devices/src/virtio/vsock/packet.rs
slp: Good point, let's address it now.

mtjhrc (approved):
LGTM, thanks for coming up with the temp fix!  Tested this by dropping the libkrunfw workaround patches.

PR #683, opened 2026-05-18
examples/vsock-latency: reproducer for first-connection vsock delay

smoser:
This was opened to share a reproducer for https://github.com/containers/libkrun/issues/684


smoser:
#806 landed, so this is not needed.

PR #687, opened 2026-05-19
Expand accepted Assisted-by formats and add CI enforcement

### LINKED ISSUE #681: CONTRIBUTING.md: `Assisted-by:` form, "auto" model selection
Let's move the discussion from #680 here.

> My .02 cents ... I wonder if we should be a bit elastic around the naming and avoid adding a deep CI check. The RH guidelines just mention we should use the labels Assisted-by/Generated-by, but there is no strict rule for how the value must be formatted.

The Assisted-by: format I picked was inspired by the kernel. I don't have a strong preference for the exact format or perhaps also allowing `Generated-by`, but if we are going to keep adding it, I would prefer to have a somewhat uniform and defined format(s).

Maybe we don't strictly need it in the CI, but I don't see it being that problematic if we do. The Signed-off-by trailer format is mandated, code formatting is mandated - I don't see why the format of this can't be standardized too. Btw please note that `CONTRIBUTING.md` currently specifies we SHOULD (not MUST) use Assisted-by in that form.

I also noticed people using  `/` as a separator, e.g. `AGENT NAME / MODEL NAME`. We could also use/allow that form since it reads a bit better, especially with spaces.

> @mtjhrc I was amending one of my commits to use the prescribed format and there's actually an issue that sometimes you might not know what the model was.

If the model name is unknown, we can just use `auto` or `unknown`; that seems fine to me. @nohajc

--- 
### My thoughts on Co-authored-by:

>  In some cases i saw Co-authored-by: Claude <noreply@anthropic.com> without any model mention.

@lstocchi
`Co-authored-by: Claude <noreply@anthropic.com>` is the default in Claude Code. Personally, I dislike it. Saying something is `Co-authored-by` implies that the co-author owns copyright to portions of the work. However, the general legal consensus is that LLMs do not own copyright to the generated work (either the user does, or no one does; of course this may differ between jurisdictions and how much user input was given).

Because of that, using `Co-authored-by: Claude <noreply@anthropic.com>` implies the LLM is an author and therefore owns copyright, which contradicts the common copyright assumptions above. There is also a comment under the RH guideline expressing a similar concern.

Also, IMHO, one of the main reasons they use the `Co-authored-by` trailer is that GitHub prominently displays authorship information, making it effectively free advertising for the vendor.

mtjhrc:
Previously only `AGENT:MODEL` was documented as valid. This relaxes
the format to also accept `AGENT / MODEL` and `AGENT (MODEL)`, and
introduces `Generated-by` for fully AI-generated code. `auto`/`unknown`
are now explicitly allowed when the model name isn't known.

A new CI check is added to enforce the policy on every PR:
- Bans `Co-authored-by` for known AI tools (e.g. Claude, Copilot)
- Rejects malformed `Assisted-by` / `Generated-by` trailers
- Rejects other well known auto-generated AI attribution lines (e.g. the
  "Generated with [Claude Code]" footer Claude Code appends by default)

Fixes: #681


jakecorrenti (approved):

.github/scripts/check-ai-trailers.py
```diff
@@ -0,0 +1,257 @@
...
+# AGENT:MODEL | AGENT: MODEL | AGENT / MODEL | AGENT (MODEL)
+ASSISTED_BY_RE = re.compile(
+    r"^(?P<key>Assisted-by|Generated-by):\s*"
+    r"(?P<agent>[A-Za-z][A-Za-z0-9 -]*[A-Za-z0-9])"
+    r"(?:"
+    r":\s*(?P<model_colon>.+)"        # AGENT: MODEL or AGENT:MODEL
+    r"| / (?P<model_slash>.+)"        # AGENT / MODEL
+    r"| \((?P<model_paren>.+)\)"      # AGENT (MODEL)
+    r")$"
+)
```
lstocchi, line 41: I see for the `co-authored-by` label there is an insensitive check. Should that be applied for the `Assisted/Generated` too?

lstocchi:
LGTM 👍

PR #690, opened 2026-05-21
Tsi: handle protocol field to make DGRAM+ICMP ping work 

mtjhrc:
Three things were missing for ping to work through TSI:

1. Host used to create UDP sockets instead of ICMP — now reads the protocol field from the guest - needs https://github.com/containers/libkrunfw/pull/127

2. Guest used to reject SOCK_DGRAM+ICMP with EACCES — init now sets ping_group_range

3. macOS host includes the IP header in ICMP recv — we need to strip it before forwarding to guest

NOTE: This only works with modern ping implementations using DGRAM+ICMP "ping" sockets  ( works on modern Fedora, Ubuntu). Busybox ping uses SOCK_RAW which TSI doesn't hijack.

### DIFF
--- a/src/devices/src/virtio/vsock/tsi_dgram.rs
+++ b/src/devices/src/virtio/vsock/tsi_dgram.rs
@@ -38,6 +38,8 @@ pub struct TsiDgramProxy {
     sendto_addr: Option<SockaddrStorage>,
     listening: bool,
     family: AddressFamily,
+    #[cfg_attr(not(target_os = "macos"), allow(dead_code))]
+    protocol: u16,
     mem: GuestMemoryMmap,
     queue: Arc<Mutex<VirtQueue>>,
     rxq: Arc<Mutex<MuxerRxQ>>,
@@ -116,6 +118,7 @@ impl TsiDgramProxy {
             sendto_addr: None,
             listening: false,
             family,
+            protocol,
             mem,
             queue,
             rxq,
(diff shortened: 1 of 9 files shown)


src/devices/src/virtio/vsock/tsi_dgram.rs
```diff
@@ -170,11 +183,31 @@ impl TsiDgramProxy {
...
+                    // macOS DGRAM ICMP sockets include the IP header in
+                    // recv, unlike Linux which strips it. Strip the IP
+                    // header (variable length, from the IHL field) so the
+                    // guest sees the same format as a Linux ping socket.
+                    // buf is the guest's RX virtqueue descriptor — writable.
+                    #[cfg(target_os = "macos")]
+                    if matches!(
+                        self.protocol as _,
+                        libc::IPPROTO_ICMP | libc::IPPROTO_ICMPV6
+                    ) && cnt >= 20
```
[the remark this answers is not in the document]
mtjhrc: fixed

tests/test_cases/src/test_tsi_ping.rs
```diff
@@ -0,0 +1,240 @@
...
+    use crate::{Test, TestOutcome, TestSetup};
+    use crate::{krun_call, krun_call_u32};
+    use krun_sys::*;
+    use std::ffi::CString;
+    use std::net::{Ipv4Addr, Ipv6Addr};
+    use std::os::unix::io::AsRawFd;
+    use std::process::Command;
+
+    const CONTAINERFILE: &str = "\
+FROM fedora:44
```
mz-pdm: Is it necessary to specify the exact Fedora version?  Would `fedora:latest` perhaps work the same or better (never obsoletes, detecting problems with newer guests)?

tests/test_cases/src/test_tsi_ping.rs
```diff
@@ -0,0 +1,240 @@
...
+            String::from_utf8(output.stdout).ok()?
+        };
+
+        for line in text.lines() {
+            for token in line.split_whitespace() {
+                let addr_str = token.split('/').next().unwrap_or(token);
+                if let Ok(addr) = addr_str.parse::<Ipv4Addr>()
+                    && !addr.is_loopback()
+                    && !addr.is_unspecified()
+                {
```
mz-pdm: Should the inner look break even if any of the `addr.is_*` conditions is true?  If the loop continues after hitting a valid (although unwanted) address, it may pick another IP-address-like field from the line, such as netmask.

tests/test_cases/src/test_tsi_ping.rs
```diff
@@ -0,0 +1,240 @@
...
+            }
+            Ok(())
+        }
+
+        fn check(self: Box<Self>, stdout: Vec<u8>, _test_setup: TestSetup) -> TestOutcome {
+            let output = String::from_utf8(stdout).unwrap_or_default();
+            if output.starts_with("SKIP:") {
+                TestOutcome::Skip(match self.version {
+                    IpVersion::V4 => "no pingable IPv4 address found on host",
+                    IpVersion::V6 => "no pingable IPv6 address found on host",
```
mz-pdm: Could the message from `output` be reused?  In order to provide a correct message in case there are more skip reasons in future.

mz-pdm (commented):
Looks good to me overall, AFAICT.  Just a few minor comments on the tests.

PR #694, opened 2026-05-27
Implement TDShim support

jakecorrenti:
## Background

libkrun currently boots TDX (Trust Domain Extensions) guests using qboot as the
firmware. qboot is a minimal x86 firmware designed for fast VM startup — it
initializes just enough hardware state to hand off to a Linux kernel. For TDX,
qboot sets up the initial guest memory layout and measurement regions, then jumps
to the kernel entry point.

TD-Shim is an alternative firmware specifically designed for TDX. It implements
the UEFI PI (Platform Initialization) Hand-Off Block (HOB) protocol, which is
the standard mechanism for passing platform configuration from firmware to a
payload in the UEFI ecosystem. Where qboot is a general-purpose minimal
firmware, TD-Shim is purpose-built for confidential computing and provides:

- **TDVF section-based memory layout**: Firmware sections (BFV, TempMem, TD_HOB,
  etc.) are described in a TDVF metadata table embedded in the binary, each with
  its own guest physical address, size, and TDX measurement attributes.
- **HOB-based platform description**: RAM regions, the kernel entry point, and
  payload metadata are communicated to the payload through a HOB chain rather
  than through legacy boot protocol structures.
- **Per-section TDX measurement**: Each firmware section carries its own
  `attributes` field controlling whether it is measured into the TD measurement
  registers, enabling fine-grained attestation of the boot chain.

TDVF (TDX Virtual Firmware, i.e., a full OVMF build with TDX support) also uses
the HOB protocol, but it is a complete UEFI implementation — hundreds of
megabytes of firmware that boots through SEC/PEI/DXE phases. TD-Shim is a
lightweight alternative that goes straight from reset to payload execution,
making it a better fit for libkrun's fast-boot philosophy.

## What this series does

This series adds TD-Shim as an alternative firmware for TDX guests, selectable
via a new `krun_set_tee_firmware()` API. The implementation:

1. **Parses TD-Shim firmware binaries** by reading the TDVF metadata table to
   extract section descriptions (addresses, sizes, types, measurement
   attributes).

2. **Builds a HOB chain** in guest memory containing: a PHIT (Phase Handoff
   Information Table) header, resource descriptors for each RAM region, a
   PayloadInfo GUID extension HOB with the kernel entry point, and an end-of-list
   marker.

3. **Loads firmware sections** into guest memory at the addresses specified by
   the TDVF metadata, and constructs `MeasuredRegion` entries with per-section
   attributes for TDX memory initialization.

4. **Writes a boot_params trampoline** — a small block of x86_64 machine code
   placed just before `startup_64` that patches td-shim's `boot_params` struct
   (passed via RSI) with the initrd address/size and cmdline pointer before
   jumping to the kernel. This is necessary because td-shim constructs its own
   `boot_params` and doesn't know about the initrd or command line that libkrun
   placed in guest memory.

5. **Carves a dynamic firmware hole** in the guest memory layout for TD-Shim's
   high-memory sections, replacing the fixed qboot firmware region at
   `FIRMWARE_START`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -681,7 +681,7 @@ dependencies = [
  "linux-loader",
  "tdx",
  "vm-memory",
- "vmm-sys-util 0.15.0",
+ "vmm-sys-util",
  "windows-sys",
 ]
 
@@ -709,7 +709,7 @@ version = "0.1.0-2.0.0-dev"
 dependencies = [
  "kvm-bindings",
  "kvm-ioctls",
- "vmm-sys-util 0.15.0",
+ "vmm-sys-util",
 ]
 
 [[package]]
@@ -741,7 +741,7 @@ dependencies = [
  "virtio-bindings",
  "vm-fdt",
  "vm-memory",
- "vmm-sys-util 0.15.0",
+ "vmm-sys-util",
  "windows-sys",
  "zerocopy",
 ]
@@ -839,7 +839,7 @@ dependencies = [
  "pkg-config",
  "remain",
  "thiserror 1.0.69",
- "vmm-sys-util 0.15.0",
+ "vmm-sys-util",
  "winapi",
  "zerocopy",
 ]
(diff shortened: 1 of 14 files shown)


slp:
Thanks @jakecorrenti, let's get this one merged.

slp (approved):

PR #703, opened 2026-05-31
balloon: use MADV_FREE to reclaim memory on macOS

pftbest:
The current implementation of balloon on macOS is not working correctly, because it is using the wrong flag for madvise.

The MADV_DONTNEED flag is no-op on macOS because it preserves the data, so the OS cannot discard the pages. MADV_FREE flag actually frees the memory under pressure, which is what we need for balloon.

How to test this:
1. Inside the VM fill the whole memory with incompressible data:
```bash
mount -t tmpfs -o size=90% none /tmp
dd if=/dev/urandom of=/tmp/data bs=1M
```
2. Observe on the host the memory pressure is high (run some background apps to make it even higher if needed).
3. Remove the data in the VM
```bash
rm /tmp/data
```
4. Observe any change in memory pressure

With MADV_DONTNEED flag nothing changes, balloon is not working:
Screenshot 2026-05-31 at 10 45 49

With MADV_FREE the pressure drops right away:
Screenshot 2026-05-31 at 10 52 53

Notably the per process memory usage in Activity Monitor still reports 8GB in both cases, however this numbers are not accurate, so the only way to test is to observe the pressure and total RAM usage.

### DIFF
--- a/src/devices/src/virtio/balloon/device.rs
+++ b/src/devices/src/virtio/balloon/device.rs
@@ -93,11 +93,15 @@ impl Balloon {
                     "balloon: should release guest_addr={:?} host_addr={:p} len={}",
                     desc.addr, host_addr, desc.len
                 );
+                #[cfg(target_os = "linux")]
+                let advice = libc::MADV_DONTNEED;
+                #[cfg(target_os = "macos")]
+                let advice = libc::MADV_FREE;
                 unsafe {
                     libc::madvise(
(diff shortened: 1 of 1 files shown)


slp:
Nice catch, thanks @pftbest. Tested and confirmed fixing the behavior virtio-balloon on macOS.

slp (approved):

PR #705, opened 2026-06-03
tests/virtiofs-misc: add rename-overwrite fchmod subtest

slp:
Add regression test for PR #700, which fixed a file descriptor leak on macOS when renaming over an existing file with open fds.

The bug: macOS virtio-fs uses volfs paths (/.vol/{dev}/{ino}) which become invalid when the inode's last link is removed. When rename() replaces a file, the old target loses its directory entry, breaking subsequent operations on still-open file descriptors.

This broke apt/dpkg atomic writes with errors like: "Problem closing the file /var/lib/dpkg/status - close (2: No such file or directory)"

The fix stores an fd to the doomed target before rename, preserving access to the unlinked inode for tracked fds.

The test verifies that fchmod() works on an open fd after the file has been atomically replaced via rename.

Assisted-by: Claude Code:claude-opus-4.6

### DIFF
--- a/tests/test_cases/src/test_virtiofs_misc.rs
+++ b/tests/test_cases/src/test_virtiofs_misc.rs
@@ -465,6 +465,68 @@ mod guest {
         }
     }
 
+    /// Test rename-overwrite with open file descriptors (PR #700 regression test).
+    ///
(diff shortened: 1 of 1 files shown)


slp:
Merged as part of #700

PR #706, opened 2026-06-04
Move krun-init to a separate .so library (similar to libkrunfw)

mtjhrc:
Introduce libkrun_init.so — a small standalone library produced by ffier (https://github.com/mtjhrc/ffier) from the init-blob crate, following the same pattern as libkrunfw for kernel delivery. It owns the InitConfig type and exposes it over C FFI via auto-generated bindings.

Bundling the init config builder (host side) and the guest PID-1 binary into a single .so ensures they always stay in sync — the config format between them is an internal detail, no longer guaranteed stable across versions. Callers like crun use krun_init_builder_from_oci_json() to build the config from their existing OCI JSON.

This design also means libkrun does not require using libkrun_init at all — users who want a custom init can inject arbitrary guest files manually via the virtiofs overlay and set their own init= kernel argument. libkrun is fully agnostic to what runs as PID 1.

The new krun_init_config_apply is the entry point: it injects the init binary and config into a virtiofs device as overlay files and appends init=/init.krun to the kernel command line. libkrun never sees the config format.

On top of #688's deprecated API removal and #670's Rust init rewrite, this PR also:
- Removes krun_disable_implicit_init/krun_get_default_init and all implicit init injection
- Gates krun_set_exec/krun_set_env/krun_set_workdir/krun_set_rlimits behind aws-nitro (returns -ENOTSUP otherwise)
- Stops hardcoding init=/init.krun in the kernel command line (now set by Config::apply)
- Moves init crates under init/ (the whole init/ may live in a separate repo eventually)
- Ports the tests and examples to the new krun_init_config_apply API

Depends on
- #688
- #670


mz-pdm (commented):
Just couple of nitpicks.  The change makes sense but is quite complex to review, especially "add ffier-based init config library" commit.

tests/test_cases/src/common.rs
```diff
@@ -1,14 +1,13 @@
-//! Common utilities used by multiple test
+//! Common utilities used by multiple test.
```
mz-pdm: When already changing it: s/test/tests/

examples/consoles.c
```diff
@@ -9,6 +9,32 @@
...
+    err = NULL;                                                                \
+    call;                                                                      \
+    if (err) {                                                                 \
+        flockfile(stderr);                                                     \
+        fprintf(stderr, "%s failed: ", #call);                                 \
+        krun_init_error_message(err, &stderr_writer);                          \
+        fputc('\n', stderr);                                                   \
+        funlockfile(stderr);                                                   \
+        krun_init_error_destroy(err);                                          \
+        return 1;                                                              \
```
mz-pdm: Why returning `1` here and `-1` in other similar places?

tests/test_cases/src/common.rs
mtjhrc: Fixed in the tests port commit: `test.` → `tests.`

examples/consoles.c
mtjhrc: Good catch — that was inconsistent with the `TRY` macros in `nitro.c` / `chroot_vm.c`. Changed to `return -1` in the examples port commit. (The pre-existing `return 1` paths in `main()` for usage/libkrun errors were left as-is.)

Makefile
```diff
@@ -97,11 +109,27 @@ ifeq ($(PREFIX),)
...
+	cargo build -p krun-init-blob --features ffi
+	cp target/ffier-krun_init.json init/init-blob-via-cdylib/ffier-krun_init.json
+	{ echo '/* @generated by: make gen-init-blob-bindings — do not edit manually. */'; \
+	  ffier-gen-c-header --fn-typedefs target/ffier-krun_init.json LIBKRUN_INIT_H; } > $(LIBRARY_HEADER_INIT)
+
+# Verify that checked-in generated files are up to date (for CI).
+verify-init-blob-bindings: gen-init-blob-bindings
+	@git diff --exit-code init/init-blob-via-cdylib/ffier-krun_init.json \
+		$(LIBRARY_HEADER_INIT) \
+	|| { echo "error: init-blob bindings are out of date — run 'make gen-init-blob-bindings'"; exit 1; }
```
mz-pdm: This message is misleading, `gen-init-blob-bindingsq` has just been run and running it again has no effect.  It should instruct to additionally commit the changes in the generated files.

Makefile
```diff
@@ -97,11 +109,27 @@ ifeq ($(PREFIX),)
     PREFIX := /usr/local
 endif
 
-.PHONY: install clean test test-prefix $(LIBRARY_RELEASE_$(OS)) $(LIBRARY_DEBUG_$(OS)) libkrun.pc clean-sysroot clean-all
+.PHONY: install clean test test-prefix gen-init-blob-bindings verify-init-blob-bindings $(LIBRARY_RELEASE_$(OS)) $(LIBRARY_DEBUG_$(OS)) libkrun.pc libkrun_init.pc clean-sysroot clean-all
```
mz-pdm: Why is libkrun_init.pc (as well as libkrun.pc) PHONY?

.gitignore
```diff
@@ -16,5 +18,7 @@ examples/nitro
 examples/consoles
 examples/rootfs_fedora
 test-prefix
+
+# Cross-compilation sysroots (downloaded by Makefile)
```
mz-pdm: Related?

init/init-blob/src/config.rs
```diff
@@ -0,0 +1,339 @@
...
+    /// finished [`Config`].
+    pub fn build(mut self) -> Config {
+        // FIXME: do not mixup user env vars with libkrun internal config.
+        // Inject rlimits as KRUN_RLIMITS env var.
+        if !self.rlimits.is_empty() {
+            let value = self.rlimits.join(",");
+            self.inner
+                .process
+                .env
+                .retain(|e| !e.starts_with("KRUN_RLIMITS="));
```
mz-pdm, line 266: Why are the other environment variables deleted?

init/init-blob/src/config.rs
```diff
@@ -0,0 +1,339 @@
...
+        assert_eq!(
+            json["process"]["args"],
+            serde_json::json!(["/usr/bin/bash", "--login"])
+        );
+        assert_eq!(json["process"]["cwd"], "/home/user");
+        assert_eq!(json["mounts"][0]["type"], "tmpfs");
+
+        // rlimits injected as env var
+        let env = json["process"]["env"].as_array().unwrap();
+        assert!(env.iter().any(|v| v.as_str() == Some("KRUN_RLIMITS=7=0:0")));
```
mz-pdm: I'm not sure about rlimits and environment variables, but to make the test
exhaustive:

- If the variables must be removed when rlimits are present, this should be checked here and there should be another test without `rlimits` for `env` contents.

- Otherwise the variables should be checked here.

mz-pdm (commented):
Are breaking changes for libkrun 2.0 tracked somewhere, to be put to release notes?

Minor typo in the first commit message: `src/init-blob/` -> `src/init_blob/`.

init/init-blob/src/config.rs
mtjhrc: The code in question here looks correct to me, it removes the user specified `KRUN_RLIMITS=` env var and replaces it with `rlimits` configured by the `rlimits` setter, other env variables are not touched.

Btw expanding on the FIXME above this code - what I am trying to say is, we should make `rlimits` a separate field in the json, but I didn't want to make that part of this PR as that would require changing the init-binary itself (furthermore they should eventually share the json schema/struct as a common dependency).

init/init-blob/src/config.rs
mtjhrc: Yes that's a good point the test wasn't very good. It now asserts the list of all env variables. 
I didn't add another test for the other case(s) - I plan to remove this hack anyway.

Makefile
mtjhrc: `libkrun.pc` is already PHONY on main. See the commit: https://github.com/libkrun/libkrun/commit/2113a24fa182e09c6a1d6d0204a925c08fda0a98 (maybe there is a better way to say the target depends on the env var?)

Makefile
mtjhrc: Hmm, the message was meant in the context of the CI (where that make target would run) - there it would make sense and instruct the user to regenerate the bindings locally. 

I have moved these commands to the `.github/workflows/code-quality.yml ` and added instructions to commit the files.

.gitignore
mtjhrc: Oops missed that one (Claude can't resist randomly adding useless comments... :roll_eyes:  ).
I also noticed there was an `init/init` in the `.gitignore`, that one was a rebase leftover after this has been rebased this from the old C init onto the Rust one -  removed that too.

init/init-blob/src/config.rs
mz-pdm: My confusion, sorry.  OK.

Makefile
mz-pdm: I see.  This explains why I get `libkrun.pc` owned by root in my source directory, which causes various problems; it's because I run `sudo make install`.  If libkrun.pc should match PREFIX, then I'd say the source tree version should contain the compilation PREFIX and the installation prefix should be applied to the target.  IOW `install` should generate the target file rather than copying it from the source directory.  But I don't know what's the common practice, maybe what's done here is a common pattern, although not ideal.

(The same arguments for libkrun_init.pc.)

Makefile
mz-pdm: Good idea, thanks.

slp:
LGTM, please rebase.

mtjhrc:
Wait a bit before merging, test broke on macOS.

mtjhrc:
@slp Fixed it - `vm-pause` test wasn't ported properly.
NOTE: there is another test failure on macOS, but it's pre-existing on main:
```
[virtiofs-misc] ....................... FAIL:
expected exactly "OK\n", got "FAILED: uid_gid_mapping_preexisting\n"
```
(Caused by the revert I think)

mtjhrc:
I forgot to mention I have a crun branch which uses this new API: https://github.com/mtjhrc/crun/tree/libkrun-2.0-dev-init-blob

slp (approved):

PR #708, opened 2026-06-06
init: loosen perms on /dev/kvm if it exists

dustymabe:
Best effort loosen /dev/kvm permissions to allow nested virtualization by unprivileged processes inside the microVM (usually a single purpose environment). Log errors but don't log ENOENT since the guest kernel may not support KVM or nested virtualization might not be enabled.

### DIFF
--- a/init/init.c
+++ b/init/init.c
@@ -521,6 +521,16 @@ static int mount_filesystems()
         return -1;
     }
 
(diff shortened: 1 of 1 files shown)


slp (approved):
LGTM, thanks!

PR #709, opened 2026-06-08
virtio/console: add Windows support 

lstocchi:
Add Windows backends for console I/O using native HANDLEs instead of Unix file descriptors. Introduce SendHandle in utils for thread-safe handle passing, platform-specific PortInput/PortOutput/TerminalProperties implementations using ReadFile/WriteFile/WaitForMultipleObjects, Windows terminal raw mode via Get/SetConsoleMode, and the corresponding C API with void* handle parameters behind #ifdef _WIN32.


src/devices/src/virtio/console/port_io.rs
```diff
@@ -31,26 +54,38 @@ pub trait PortTerminalProperties: Send + Sync {
...
+
+#[cfg(unix)]
 pub fn stdout() -> Result<Box<dyn PortOutput + Send>, nix::Error> {
     output_to_raw_fd_dup(STDOUT_FILENO)
 }
 
+#[cfg(unix)]
 pub fn stderr() -> Result<Box<dyn PortOutput + Send>, nix::Error> {
     output_to_raw_fd_dup(STDERR_FILENO)
 }
```
mtjhrc: IMO there are just too many `#[cfg(unix)]` and `#[cfg(windows)]` in this file. I'd say this should be split into seperate unix/windows modules. You can either do it in the same file (`#[cfg(unix)] mod unix {/* all unix stuff */}; pub use unix::*;` ) or create separate source files for unix/windows under a port_io directory.

lstocchi: Updated. Splitted them under port_io as it seemed the cleaner solution

slp (approved):
LGTM, thanks!

PR #710, opened 2026-06-08
fs/macos: on readdir, refresh directory on rewind

slp:
Use offset == 0 as an indicator that we're dealing with either a fresh directory stream or one that has been rewound. If that's the case, ensure the cache will be refreshed.

Suggested-by: Jay Gowdy <jgowdy@godaddy.com>


slp:
Supersedes #693 

@pftbest @mtjhrc PTAL

pftbest:
LGTM

PR #712, opened 2026-06-09
init_blob: align crate with the workspace

slp:
Rename directory from init-blob to init_blob, set edition to 2024, rename the package to krun-init-blob, and set version to 0.1.0-1.18.0.


mtjhrc (approved):
LGTM!

PR #713, opened 2026-06-09
Prepare for 1.18.2 stable release

slp:


mtjhrc:
This kind of sounds like it should be 1.19.0 since we are including:

-  #673

mtjhrc:
In #673 @d-e-s-o  mentioned this will eliminate the need for setting KRUN_INIT_BINARY_PATH. Does this branch alone work? I am not sure what the use-case/problem is... 
Don't we also need to add an "init" cargo feature to libkrun (enabled by default for backwards compatibility) that user can disable to not depend on `init-blob`?

d-e-s-o:
> In #673 @d-e-s-o mentioned this will eliminate the need for setting KRUN_INIT_BINARY_PATH. Does this branch alone work? I am not sure what the use-case/problem is...

My problem is that at the moment `cargo build` in a project using `libkrun` from `crates.io` does not work without the setting of `KRUN_INIT_BINARY_PATH`, because the `init` C source code is not published, see discussion [here](https://github.com/libkrun/libkrun/pull/645#issuecomment-4289994446). You cannot, generally speaking, force this variable to be set without building additional `Makefile` or other infrastructure on top of what Cargo already does.

But looking at 0127980f04489ed3e0a5c5f2aa4e784aec884a33 again, it just moves stuff around. I could swear I saw additional conditional logic yesterday. Anyway, so just picking that won't help my use case after all (when I tried it out I was relying on the local checkout, which *does* have the `init`'s C files available) :-/

mtjhrc:
> But looking at https://github.com/libkrun/libkrun/commit/0127980f04489ed3e0a5c5f2aa4e784aec884a33 again, it just moves stuff around. I could swear I saw additional conditional logic yesterday.

Right that is what I was getting at, while that PR allows you to disable using the init at **runtime** we still statically include the binary at compile time unconditionally.  I think the way to solve this in 1.x would be to make the init-blob dependency optional (via cargo feature) 

 (btw for 2.x I am thinking of linking the init-blob dynamically - #706)

d-e-s-o:
Some hackery such as https://github.com/d-e-s-o/libkrun/commit/4da2699a0281be54a5dd68eff217db11ae91b689 would make it work, but it's not the finest of approaches... A feature as you suggest would be cleaner (haven't thought through all the details, though)

Anyway, you can reproduce by just `cd`'ing into `init-blob/` and doing `cargo package`. It will attempt to build and fail, because init sources are not present.

d-e-s-o:
> I think the way to solve this in 1.x would be to make the init-blob dependency optional (via cargo feature)

Yeah. I'd be willing to look into that if nobody else feels like doing it and assuming that's a route you guys are willing to take. Let me know.

(no need to block this PR for it, of course)

mtjhrc:
I have no idea how this works, but what if we just moved the source code of `init` into the `init-blob` crate wouldn't the cargo package just include it? I mean that is what we also do with e. g. `krun_display` crate - it ships with `libkrun_display.h` inside the crate - https://docs.rs/crate/krun-display/0.1.0/source/libkrun_display.h

d-e-s-o:
> I have no idea how this works, but what if we just moved the source code of `init` into the `init-blob` crate wouldn't the cargo package just include it? I mean that is what we also do with e. g. `krun_display` crate - it ships with `libkrun_display.h` inside the crate.

Yes, I believe that would also work. However, it would nevertheless be nice to have the option of not building and including the default init blob (but it's orthogonal). If we build and include it, any downstream binary may still include this potentially large blob of bytes, despite it being nothing but dead weight (that's why right now I have to set `KRUN_INIT_BINARY_PATH` to an empty file -- I am not using the default init and don't want it to bloat my binary).

slp:
In that case, let's drop the cherry-picks for #673 and related, and leave that for `1.18.3` (we can do an early release if we managed to figure out things soon enough).

mtjhrc:
@slp I'll try giving it a go right now I think it should be fairly simple to add a cargo feature for this.

slp:
> @slp I'll try giving it a go right now I think it should be fairly simple to add a cargo feature for this.

Okay, let's hold the release for a min.

slp:
Closing this one in favor of #715

PR #714, opened 2026-06-09
Add "init-blob" cargo feature to gate bundled init binary

mtjhrc:
See the discussion in #713 for more context.

### DIFF
--- a/src/libkrun/src/lib.rs
+++ b/src/libkrun/src/lib.rs
@@ -2058,6 +2058,7 @@ const KRUN_FEATURE_AMD_SEV: u64 = 7;
 const KRUN_FEATURE_INTEL_TDX: u64 = 8;
 const KRUN_FEATURE_AWS_NITRO: u64 = 9;
 const KRUN_FEATURE_VIRGL_RESOURCE_MAP2: u64 = 10;
+const KRUN_FEATURE_INIT_BLOB: u64 = 11;
 
 #[unsafe(no_mangle)]
 pub extern "C" fn krun_has_feature(feature: u64) -> c_int {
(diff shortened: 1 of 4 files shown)


src/libkrun/src/lib.rs
```diff
@@ -2523,6 +2540,15 @@ pub extern "C" fn krun_disable_implicit_init(ctx_id: u32) -> i32 {
     KRUN_SUCCESS
 }
 
+#[unsafe(no_mangle)]
+#[cfg(all(
+    not(feature = "init-blob"),
+    not(any(feature = "tee", feature = "aws-nitro"))
+))]
+pub extern "C" fn krun_disable_implicit_init(_ctx_id: u32) -> i32 {
+    KRUN_SUCCESS
```
d-e-s-o, line 2551: Just wondering why not `ENOTSUP`? Seems mildly inconsistent with respect to other functions, but I may not have all the context.

d-e-s-o (approved):
I tried it out and it works for me. Thanks for getting this done so quickly.

src/libkrun/src/lib.rs
mtjhrc: Well because it is already disabled, so "disabling" it was a success.

The idea is that the consumer always calls this function to be compatible with the legacy automatically injected init and modern way of explicitly injecting it.

mtjhrc:
NOTE: this is technically not backwards compatible with 1.18.1 in the crates.io release though - if a consumer hypothetically used `no-default-features` this would break them, which would be unfortunate but this is unlikely and worth it IMO.

(Traditional consumers via dynamic linking are not effected, init stays as a default feature).

slp (approved):
LGTM, thanks

PR #716, opened 2026-06-09
virtio/balloon: add Windows support for free-page reporting

lstocchi:
Use DiscardVirtualMemory as the Windows equivalent of madvise for releasing guest memory in the balloon free-page reporting queue. Gate the unix AsRawFd import and add the Windows shim for the event handler.


jakecorrenti (approved):
I don't have easy access to a windows machine at the moment so I haven't tested it, but the code LGTM

slp (approved):
Can't test it yet, but also LGTM.

PR #717, opened 2026-06-10
vhost-user: add GPU device support

dorindabassey:
vhost-user: add GPU device support which
- Consolidate display/EDID code into a new shared `devices/display/` module available to both virtio-gpu and vhost-user-gpu
- Implement vhost-user GPU socket protocol (`GET_DISPLAY_INFO`, `GET_EDID`, `SCANOUT`) following QEMU's vhost-user-gpu patterns
- Expose `KRUN_VIRTIO_DEVICE_GPU` in `libkrun.h` and add  `--vhost-user-gpu=PATH` to the `chroot_vm` example

Display output (UPDATE/cursor messages) is not yet implemented; this is sufficient for GPU compute and 3D acceleration workloads.
For the Backend:
`RUST_LOG=debug ./target/debug/vhost-device-gpu --socket-path /tmp/gpu.socket --gpu-mode virglrenderer --gpu-device /dev/dri/renderDxxx`
In the guest:
`LD_LIBRARY_PATH=./target/release ./examples/chroot_vm --vhost-user-gpu=/tmp/gpu.socket / /bin/sh`
run:
`eglinfo | grep "driver name"` -> `virtio_gpu` (not `kms_swrast`)
`eglinfo | grep "renderer"` -> `virgl (host GPU model)`


slp (dismissed):
LGTM, thanks @dorindabassey !

@mtjhrc PTAL.

src/devices/src/virtio/vhost_user/device.rs
```diff
@@ -153,6 +209,19 @@ impl VhostUserDevice {
...
             .collect();
 
+        const MAX_SCANOUTS: usize = 16;
+        let gpu_scanout_state = if device_type == VIRTIO_ID_GPU {
+            let mut states = vec![GpuScanoutState::default(); MAX_SCANOUTS];
+            states[0] = GpuScanoutState {
+                enabled: true,
+                width: 1024,
+                height: 768,
+            };
```
mtjhrc: This looks like some hard coded leftover testing code, not something that should be part of final code.

src/devices/src/virtio/vhost_user/device.rs
```diff
@@ -468,17 +561,232 @@ impl VirtioDevice for VhostUserDevice {
...
+
+        let scanout_state = &self.gpu_scanout_state[scanout_id as usize];
+
+        // Guest driver requests EDID for all scanouts during initialization, but only
+        // uses EDID for scanouts marked enabled in GET_DISPLAY_INFO. For disabled
+        // scanouts, provide default dimensions to ensure valid EDID generation.
+        let (width, height) = if scanout_state.width > 0 && scanout_state.height > 0 {
+            (scanout_state.width, scanout_state.height)
+        } else {
+            (1024, 768)
```
mtjhrc: This is not really correct and conflates 2 different things in the vhost/virtio gpu terminology. There is the "display" and the "scanout".  

You need to generate the EDID based on the user configured **display** resolution (currently static since we don't propagate the display window resizing, but might become dynamic in the future) instead of the guest configured **scanout** resolution. 

What this means in the code is you need to keep track of 2 different width/heights, one for the scanout (guest configured - might be smaller than actual display!) and the one for the actual display.
I would suggest also moving https://github.com/libkrun/libkrun/blob/9222cb185251d3df03611e37f05e76850a068e52/src/devices/src/virtio/gpu/display.rs#L7-L24 to the shared module and reusing that. (and have separate variables for scanout)

--- 
Why this is wrong:

Suppose the display is 1920x1080 but the guest starts at a lower resolution — during early boot, the console might render at something like 800x600 before the full display driver loads. With the current code, that initial low-res scanout gets reflected into the EDID, so when the display driver does load and queries the EDID to find out what resolutions the monitor supports, it sees 800x600 and thinks that's the monitor's native resolution. The driver never offers anything higher, and the guest is stuck at a low resolution even though the display could handle 1920x1080. More generally, any time the guest happens to set a scanout smaller than the display, the EDID shrinks to match, and the guest loses the ability to scale back up — it's a one-way ratchet.

mtjhrc (changes requested):
So I get this doesn't have display implementation hooked up yet - which is totally fine. But if this only supports headless operation for now, shouldn't the list of configured displays be always empty? (Or that caused some problem?).

src/devices/src/virtio/vhost_user/device.rs
```diff
@@ -468,17 +561,232 @@ impl VirtioDevice for VhostUserDevice {
...
+                    r: VirtioGpuRect {
+                        x: 0,
+                        y: 0,
+                        width: state.width,
+                        height: state.height,
+                    },
+                    enabled: 1,
+                    flags: 0,
+                };
+            }
```
mtjhrc: This has the same display-vs-scanout issue as the EDID comment - this should report the display configuration, not echo back what the backend set via scanout commands (It already knows that).

src/devices/src/virtio/vhost_user/device.rs
dorindabassey: Not really leftover, at least one enabled display with non-zero dimensions is needed in GET_DISPLAY_INFO for the guest virtio-gpu driver to initialize correctly; without it the driver falls back to software rendering. The 1024×768 matches QEMU's default.
That said, I agree it shouldn't be hardcoded in the library. I've made it configurable via a new krun_set_vhost_user_gpu_options(ctx_id, width, height) API call, similar to krun_set_gpu_options for the built-in GPU. The 1024×768 is now just the default when the caller doesn't set anything.

src/devices/src/virtio/vhost_user/device.rs
dorindabassey: Fixed. EDID is now generated from gpu_display_info (user-configured display resolution) rather than the guest scanout. And I've moved DisplayInfo into the shared devices/display module and used it.

src/devices/src/virtio/vhost_user/device.rs
dorindabassey: Fixed alongside the EDID change.

src/devices/src/virtio/vhost_user/device.rs
```diff
@@ -86,6 +133,7 @@ impl VhostUserDevice {
         device_name: String,
         num_queues: u16,
         queue_sizes: &[u16],
+        gpu_display: Option<(u32, u32)>,
```
mtjhrc: I'd say accepting `DisplayInfo` struct directly would be better. The builtin GPU already does that (well it accepts a Vec of them)  - this is more flexible because it allows the `DisplayInfo` struct to carry custom EDID info.

src/vmm/src/resources.rs
mtjhrc: What do you think about just doing:
```suggestion
#[cfg(any(feature = "gpu", "vhost-user"))]
pub displays: Vec<DisplayInfo>,
```

Instead of separate `pub vhost_user_gpu_display: Option<(u32, u32)>`?

This would automatically gain you support for all of the display configuration APIs instead of having to re-implement them:
- `krun_add_display`
- `krun_display_set_refresh_rate`
- `krun_display_set_edid`
- `krun_display_set_physical_size`
- `krun_display_set_dpi`

In the future `krun_add_display` should have a way to specify *which* GPU you are adding the display for (technically you might have multiple GPUs each with different displays), but I don't think we need to handle that immediately.

mtjhrc (changes requested):

src/devices/src/virtio/vhost_user/device.rs
mtjhrc: Hmm, I wonder where the problem is. I'm pretty sure it is supposed to work without displays. For the builtin GPU we have a default of no displays attached. Maybe it's a bug in vhost-user-gpu? Or the `vhost` (but not virtio) protocol has problems with headless GPUs.

mtjhrc: I took a quick, look and it looks our builtin GPU isn't following the spec (but in practice it seems to work).

vhost-user-gpu hardcodes the config to 16 displays: https://github.com/rust-vmm/vhost-device/blob/2626361051196542f03def6af1992c79afe3b287/vhost-device-gpu/src/device.rs#L174

But our virtio-gpu (which works without displays) sets it to the number of displays we have (in this case zero):
https://github.com/libkrun/libkrun/blob/d4bb6e040ad7b6e6a6cf68ffced3b895c2753a18/src/devices/src/virtio/gpu/device.rs#L161-L167

But you are right the spec (IMO weirdly) says minimum is 1:
https://docs.oasis-open.org/virtio/virtio/v1.4/csprd01/virtio-v1.4-csprd01-diff-from-v1.2-cs01.html#x1-4770004

So now I am not sure what to do about this....  One alternative we could do is return an error from `krun_add_vhost_user_device(...,KRUN_VIRTIO_DEVICE_GPU, ...)` `if displays.is_empty()`.  


But also the linux kernel code happily accepts `0` as a proper value, it even handles it properly e.g. here:  https://github.com/torvalds/linux/blob/0ff41df1cb268fc69e703a08a57ee14ae967d0ca/drivers/gpu/drm/virtio/virtgpu_kms.c#L228

@slp Any thoughts on this? Follow the kernel and allow `0` displays or follow the spec and force having a display?

dorindabassey: Since the spec requires a minimum of 1, I'll keep the 1024×768 default, this is also what QEMU does: it defaults to 1024×768.
Callers can override it with krun_add_display().

dorindabassey: ping @mtjhrc @slp wdyt?

dorindabassey:
Hi @mtjhrc do you have sometime to look at this? Thanks!

mtjhrc (approved):
LGTM thanks! 
Tested with offscreen glmark - works and detects virgl.

src/devices/src/virtio/vhost_user/device.rs
mtjhrc: crosvm also seems to have a default display, so let's keep this.

PR #718, opened 2026-06-10
Full native Rust API + autogenerated C API

mtjhrc:
Typed Rust API for libkrun. C header and Rust cdylib client are generated from the Rust types using ffier (https://github.com/mtjhrc/ffier). The v1 C API is removed.

Depends on (in order)
- #670 — Port init binary to Rust
- #688 — Remove deprecated APIs and implicit resource creation
- #706 — Move krun-init to a separate .so library (init-blob)


examples/chroot_vm/src/main.rs
```diff
@@ -0,0 +1,118 @@
...
+
+    let mut vmm = VmmBuilder::new()
+        .vcpus(2)
+        .map_err(|e| anyhow::anyhow!("{e:?}"))?
+        .ram_mib(512)
+        .map_err(|e| anyhow::anyhow!("{e:?}"))?
+        .kernel(kernel)
+        .devices(devices)
+        .build()
+        .map_err(|e| anyhow::anyhow!("{e:?}"))?;
```
appcypher: nit: i'm wondering if having all the validation happen in the .build() method would be the cleaner api. like:

```rust
let mut vmm = VmmBuilder::new()
    .vcpus(2)
    .ram_mib(512)
    .kernel(kernel)
    .devices(devices)
    .build()
    .map_err(|e| anyhow::anyhow!("{e:?}"))?;
```

mtjhrc:
@jakecorrenti Could you please take a look at the TEE stuff and aws-nitro? Does the direction of the new api make sense? Also I can't test that at all so I have likely broken some stuff. 

Alternative would be to drop those features from the API in the meantime and you can cherry pick+fix them in a separate PR. Not sure what is the best approach here.

jakecorrenti:
> @jakecorrenti Could you please take a look at the TEE stuff and aws-nitro? Does the direction of the new api make sense? Also I can't test that at all so I have likely broken some stuff.
> 
> Alternative would be to drop those features from the API in the meantime and you can cherry pick+fix them in a separate PR. Not sure what is the best approach here.

Sure, will look today :)

src/vmm/src/linux/tee/inteltdx.rs
```diff
@@ -15,6 +15,12 @@ type Result<T> = std::result::Result<T, Error>;
 
 pub struct IntelTdx {}
```
jakecorrenti: This could be simplified by doing the following instead
```rust
#[derive(Debug)]
pub struct IntelTdx {};
```

src/vmm/src/builder.rs
```diff
@@ -1415,10 +1442,7 @@ fn load_payload(
                     return Err(StartMicrovmError::MissingKernelConfig);
                 };
 
-            #[cfg(all(feature = "vhost-user", target_os = "linux"))]
-            let use_vhost_user = !_vm_resources.vhost_user_devices.is_empty();
-            #[cfg(not(all(feature = "vhost-user", target_os = "linux")))]
-            let use_vhost_user = false;
+            let use_vhost_user = _use_vhost_user;
```
jakecorrenti: this re-assignment seems redundant unless I'm missing something

src/vmm/src/builder.rs
```diff
@@ -1659,26 +1686,23 @@ pub fn create_guest_memory(
 
     let mut shm_manager = ShmManager::new(&arch_mem_info);
 
+    #[cfg(feature = "tee")]
+    let _ = fs_shm_sizes;
```
jakecorrenti: This seems unnecessary. You could probably gate the argument behind `#[cfg(not(feature = "tdx"))]` instead

src/vmm/src/builder.rs
```diff
@@ -1659,26 +1686,23 @@ pub fn create_guest_memory(
...
             .create_gpu_region(size)
             .map_err(StartMicrovmError::ShmCreate)?;
     }
 
-    // For vhost-user devices, we need file-backed memory so the backend can mmap it
-    #[cfg(all(feature = "vhost-user", target_os = "linux"))]
-    let use_vhost_user = !vm_resources.vhost_user_devices.is_empty();
-    #[cfg(not(all(feature = "vhost-user", target_os = "linux")))]
-    let use_vhost_user = false;
+    let _ = use_vhost_user;
```
jakecorrenti: This assignment is also a no-op

src/libkrun/src/builder.rs
```diff
@@ -0,0 +1,1365 @@
...
+use vmm::vstate::MeasuredRegion;
+
+/// Builds and starts a microVM based on the current Firecracker VmResources configuration.
+///
+/// This is the default build recipe, one could build other microVM flavors by using the
+/// independent functions in this module instead of calling this recipe.
+///
+/// An `Arc` reference of the built `Vmm` is also plugged in the `EventManager`, while another
+/// is returned.
+pub(crate) fn build_microvm(
```
jakecorrenti: This definitely doesn't need to happen here, but I think we should look into maybe breaking `build_microvm()` up some. A 645 line function seems like it would warrant a re-write.

src/libkrun/src/builder.rs
```diff
@@ -0,0 +1,1365 @@
...
+            vm_resources.qboot_bundle.as_ref(),
+            #[cfg(feature = "tee")]
+            vm_resources.initrd_bundle.as_ref(),
+            vm_resources.firmware_config.as_ref(),
+            &fs_shm_sizes,
+            vm_resources
+                .gpu_virgl_flags
+                .map(|_| vm_resources.gpu_shm_size.unwrap_or(1 << 33)),
+            {
+                #[cfg(all(feature = "vhost-user", target_os = "linux"))]
```
jakecorrenti: would the following be easier to read?

```rust
if cfg!(feature = "vhost-user") && cfg!(target_os = "linux") {
    !vm_resources.vhost_user_devices.is_empty()
} else {
    false
}
```

src/libkrun/src/builder.rs
```diff
@@ -0,0 +1,1365 @@
...
+        ),
+        _ => None,
+    };
+
+    #[cfg(feature = "tdx")]
+    let mut tdx_launcher = match tee {
+        Tee::Tdx => vm
+            .tdx_secure_virt_prepare()
+            .map_err(StartMicrovmError::SecureVirtPrepare)?,
+        _ => panic!(),
```
jakecorrenti: non-blocking note: I should go in and clean this up so it's not panicking here.

src/libkrun/src/api/device_builders.rs
```diff
@@ -0,0 +1,329 @@
...
+/// during VM construction.
+pub struct MmioDeviceManager<'a> {
+    devices: Vec<Box<dyn AttachDevice<'a> + 'a>>,
+}
+
+impl Default for MmioDeviceManager<'_> {
+    fn default() -> Self {
+        Self::new()
+    }
+}
```
jakecorrenti: You should be able to just do 
```rust
#[derive(Default)]
pub struct MmioDeviceManager<'a> {
...
```

src/libkrun/src/api/device_builders.rs
```diff
@@ -0,0 +1,329 @@
...
+
+impl<'a> MmioDeviceManager<'a> {
+    /// Create an empty device manager.
+    pub fn new() -> Self {
+        Self {
+            devices: Vec::new(),
+        }
```
jakecorrenti:
```suggestion
        Self::default()
````

File: src/libkrun/src/api/device_builders.rs [hunk 11]
```diff
@@ -0,0 +1,329 @@
...
+    /// Devices are attached in the order they are added. The device must
+    /// implement [`AttachDevice`] — all built-in device types
+    /// (`FsDevice`, `ConsoleDevice`, etc.) implement this trait.
+    pub fn add(&mut self, device: impl AttachDevice<'a>) -> &mut Self {
+        self.devices.push(Box::new(device));
+        self
+    }
+}
+
+impl sealed::Sealed for MmioDeviceManager<'_> {}
```
jakecorrenti: I think it would be an over-abstraction at this point, but since `sealed::Sealed` is such a simple trait, maybe in the future we could implement the `derive` attribute for it so that we could just do 
```rust
#[derive(Sealed)]
pub struct Foo {
...
```

src/libkrun/src/api/logging.rs
```diff
@@ -0,0 +1,86 @@
...
+    Trace = 5,
+}
+
+#[repr(u32)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum LogStyle {
+    Auto = 0,
+    Always = 1,
+    Never = 2,
+}
```
jakecorrenti: Do we also want to implement the `Default` trait here? We could define `LogLevel::Error` and `LogStyle::Auto` as the defaults. I don't mind either way, just a thought.

src/libkrun/src/api/device_builders.rs
```diff
@@ -0,0 +1,329 @@
...
+            .insert_str(s)
+            .unwrap_or_else(|e| log::error!("failed to append '{s}' to cmdline: {e}"));
+    }
+}
+
+/// A resolved shared memory region with both host and guest addresses.
+///
+/// Returned by [`AttachContext::resolved_shm_region`] after guest memory has been
+/// created and the SHM region mapped.
+pub struct ResolvedShmRegion {
```
jakecorrenti: Should this be gated behind `not cfg(tee) and not cfg(nitro)`

src/libkrun/src/api/mod.rs
```diff
@@ -0,0 +1,17 @@
+pub mod device_builders;
+pub mod error;
+pub mod logging;
+pub mod payload;
+
+pub use device_builders::{
+    AttachContext, AttachDevice, DeviceManager, DeviceRequirements, MmioDeviceManager,
+    ResolvedShmRegion,
```
jakecorrenti: same gating question here regarding `ResolvedShmRegion`

src/libkrun/src/api/payload.rs
```diff
@@ -0,0 +1,219 @@
...
+        path: PathBuf,
+    },
+    #[cfg(feature = "tee")]
+    Tee {
+        bundle: vmm::vmm_config::kernel_bundle::KernelBundle,
+        qboot_bundle: vmm::vmm_config::kernel_bundle::QbootBundle,
+        initrd_bundle: vmm::vmm_config::kernel_bundle::InitrdBundle,
+        tee_config_path: PathBuf,
+    },
+}
```
jakecorrenti: where readability is concerned could probably do this here:

```suggestion
use super::error::Error;
use vmm::vmm_config::kernel_bundle;
use vmm::vmm_config::external_kernel;

// Fields are consumed by VmmBuilder once that API lands in a later commit.
#[allow(dead_code)]
pub(crate) enum PayloadKind {
    Kernel {
        bundle: kernel_bundle::KernelBundle,
    },
    External {
        kernel: external_kernel::ExternalKernel,
    },
    Firmware {
        path: PathBuf,
    },
    #[cfg(feature = "tee")]
    Tee {
        bundle: kernel_bundle::KernelBundle,
        qboot_bundle: kernel_bundle::QbootBundle,
        initrd_bundle: kernel_bundle::InitrdBundle,
        tee_config_path: PathBuf,
    },
}
```

src/libkrun/src/api/payload.rs
```diff
@@ -0,0 +1,219 @@
...
+                    qboot_bundle,
+                    initrd_bundle,
+                    tee_config_path: PathBuf::from(tee_config_path),
+                },
+                cmdline: vmm::vmm_config::kernel_cmdline::DEFAULT_KERNEL_CMDLINE.to_string(),
+            })
+        }
+        #[cfg(not(feature = "tee"))]
+        {
+            let _ = tee_config_path;
```
jakecorrenti: IMO we should do something like the following here instead:
```rust
impl Payload {
    #[cfg(not(feature = "tee"))]
    pub fn load_kurnfw() -> Result<Self, Error> { .. }

    #[cfg(feature = "tee")]
    pub fn load_krunfw(tee_config_path: &str) -> Result<Self, Error> { .. }
}

WDYT?

File: src/libkrun/src/api/payload.rs [hunk 17]
```diff
@@ -0,0 +1,219 @@
...
+
+    pub fn load_external(
+        kernel_path: &str,
+        format: KernelFormat,
+        initrd_path: Option<&str>,
+        cmdline: &str,
+    ) -> Result<Self, Error> {
+        use vmm::vmm_config::external_kernel::{ExternalKernel, KernelFormat as VmmKernelFormat};
+
+        let vmm_format = match format {
```
jakecorrenti: this feels redundant to have two identical types that represent the same thing

src/libkrun/src/api/payload.rs
```diff
@@ -0,0 +1,219 @@
...
+fn load_tee_bundles(
+    lib: &libloading::Library,
+) -> Result<
+    (
+        vmm::vmm_config::kernel_bundle::QbootBundle,
+        vmm::vmm_config::kernel_bundle::InitrdBundle,
+    ),
+    Error,
+> {
+    use vmm::vmm_config::kernel_bundle::{InitrdBundle, QbootBundle};
```
jakecorrenti: Add this to the top of the crate so you can cleanup the function signature

src/libkrun/src/api/device_builders.rs
```diff
@@ -327,3 +335,216 @@ impl<'a> DeviceManager<'a> for MmioDeviceManager<'a> {
...
+}
+
+#[cfg(not(any(feature = "tee", feature = "aws-nitro")))]
+impl FsOverlay {
+    /// Create a new empty overlay.
+    pub fn new() -> Self {
+        Self {
+            entries: Vec::new(),
+        }
+    }
```
jakecorrenti: I don't see why the Default trait is necessary here, but if you wanted to keep it just use the derivable attribute.

```suggestion
#[cfg(not(any(feature = "tee", feature = "aws-nitro")))]
#[derive(Default)]
pub struct FsOverlay {
    entries: Vec<VirtualDirEntry>,
}

#[cfg(not(any(feature = "tee", feature = "aws-nitro")))]
impl FsOverlay {
    /// Create a new empty overlay.
    pub fn new() -> Self {
        Self::default()
    }
}
```

src/libkrun/src/api/device_builders.rs
```diff
@@ -327,3 +335,216 @@ impl<'a> DeviceManager<'a> for MmioDeviceManager<'a> {
...
+    pub fn set_dax_window_size(&mut self, bytes: u64) {
+        self.shm_size = Some(bytes as usize);
+    }
+}
+
+#[cfg(not(any(feature = "tee", feature = "aws-nitro")))]
+impl<'a> FsDevice<'a> {
+    fn new_inner(tag: &str, host_path: Option<String>, read_only: bool) -> Result<Self, Error> {
+        let exit_code = Arc::new(AtomicI32::new(i32::MAX));
+        let fs = devices::virtio::Fs::new(
```
jakecorrenti: Why is this in a different `impl` block?

src/libkrun/src/api/device_builders.rs
```diff
@@ -327,3 +335,216 @@ impl<'a> DeviceManager<'a> for MmioDeviceManager<'a> {
...
+    /// # Arguments
+    ///
+    /// - `tag`: the filesystem tag visible to the guest (e.g. `"/dev/root"`).
+    /// - `host_path`: the host directory to share.
+    pub fn new(tag: &str, host_path: &str) -> Result<Self, Error> {
+        Self::new_inner(tag, Some(host_path.to_string()), false)
+    }
+
+    /// Create a read-only virtiofs device sharing a host directory.
+    pub fn new_read_only(tag: &str, host_path: &str) -> Result<Self, Error> {
```
jakecorrenti: maybe
```rust
pub fn new(tag: &str, host_Path: &str, readonly: bool) -> Result<Self, Error> { .. }
```

instead so we just have one `new()` associated function

src/libkrun/src/api/device_builders.rs
```diff
@@ -327,3 +335,216 @@ impl<'a> DeviceManager<'a> for MmioDeviceManager<'a> {
...
+            // Wire exit code from VMM into the fs device
+            fs.set_exit_code(ctx.exit_code().clone());
+            // Resolve overlay entries
+            if let Some(overlay) = self.overlay {
+                for entry in overlay.into_entries() {
+                    fs.add_virtual_entry(entry);
+                }
+            }
+            // Set up SHM region if allocated
+            #[cfg(not(any(feature = "tee", feature = "aws-nitro")))]
```
jakecorrenti: This is unnecessary since the trait implementation is already gated.

src/libkrun/src/api/device_builders.rs
```diff
@@ -528,6 +530,261 @@ impl<'a> AttachDevice<'a> for FsDevice<'a> {
...
+    pub fn builder() -> ConsoleBuilder<'a> {
+        ConsoleBuilder {
+            ports: Vec::new(),
+            tty_fds: Vec::new(),
+            _lifetime: PhantomData,
+        }
+    }
+}
+
+impl<'a> ConsoleBuilder<'a> {
```
jakecorrenti: Is there a particular reason for the multiple `impl` blocks?

init/init-binary/init-freebsd
jakecorrenti, line 1: Seems like you accidentally staged an init binary?

src/libkrun/src/api/device_builders.rs
```diff
@@ -0,0 +1,1549 @@
...
+            .insert_str(s)
+            .unwrap_or_else(|e| log::error!("failed to append '{s}' to cmdline: {e}"));
+    }
+}
+
+/// A resolved shared memory region with both host and guest addresses.
+///
+/// Returned by [`AttachContext::resolved_shm_region`] after guest memory has been
+/// created and the SHM region mapped.
+pub struct ResolvedShmRegion {
```
jakecorrenti, line 215: should this be gated behind `not tee and not nitro`?

src/libkrun/src/api/device_builders.rs
```diff
@@ -0,0 +1,1549 @@
...
+
+    /// Attach this device to the VM.
+    ///
+    /// The device should:
+    /// 1. Perform any device-specific setup using `ctx` methods
+    /// 2. Call [`ctx.register()`](AttachContext::register) to register on the transport bus
+    fn attach(self: Box<Self>, ctx: &mut AttachContext) -> Result<(), Error>;
+}
+
+mod sealed {
```
jakecorrenti, line 256: I think it would be an over-abstraction at this point, but I think we could implement the `derive` attribute here since this is such a simple trait. This would allow us to do something like:
```rust
#[derive(Sealed)]
pub struct MmioDeviceManager<'a> {
...
```

src/libkrun/src/api/device_builders.rs
```diff
@@ -0,0 +1,1549 @@
...
+}
+
+#[cfg_attr(feature = "ffi", ffier::export)]
+impl<'a> MmioDeviceManager<'a> {
+    /// Create an empty device manager.
+    pub fn new() -> Self {
+        Self {
+            devices: Vec::new(),
+        }
+    }
```
jakecorrenti, line 298: You should be able to implement the `Default` derivable attribute. That way you don't have to manually implement the `Default` trait and then it will subsequently cleanup the `new()` associated function:

```rust
#[derive(Default)]
pub struct MmioDeviceManager<'a> { .. }

impl<'a> MmioDeviceManager<'a> {
    /// Create an empty device manager.
    pub fn new() -> Self {
        Self::default()
    }
...
```

src/libkrun/src/api/device_builders.rs
```diff
@@ -0,0 +1,1549 @@
...
+
+    /// Add a device to this manager.
+    ///
+    /// Devices are attached in the order they are added. The device must
+    /// implement [`AttachDevice`] — all built-in device types
+    /// (`FsDevice`, `ConsoleDevice`, etc.) implement this trait.
+    pub fn add(&mut self, device: impl AttachDevice<'a>) -> &mut Self {
```
jakecorrenti, line 305: This signature doesn't make much sense to me? If `self` is a mutable reference we don't need to also return `self` at the end, no?

src/libkrun/src/api/device_builders.rs
```diff
@@ -0,0 +1,1549 @@
...
+    not(feature = "ffi"),
+    cfg(not(any(feature = "tee", feature = "aws-nitro")))
+)]
+impl FsOverlay {
+    /// Create a new empty overlay.
+    pub fn new() -> Self {
+        Self {
+            entries: Vec::new(),
+        }
+    }
```
jakecorrenti, line 367: Same comment as above regarding the `Default` trait implementation cleanup.

src/libkrun/src/api/device_builders.rs
```diff
@@ -0,0 +1,1549 @@
...
+    /// - `tag`: the filesystem tag visible to the guest (e.g. `"/dev/root"`).
+    /// - `host_path`: the host directory to share.
+    pub fn new(tag: &str, host_path: &str) -> Result<Self, Error> {
+        Self::new_inner(tag, Some(host_path.to_string()), false)
+    }
+
+    /// Create a read-only virtiofs device sharing a host directory.
+    pub fn new_read_only(tag: &str, host_path: &str) -> Result<Self, Error> {
+        Self::new_inner(tag, Some(host_path.to_string()), true)
+    }
```
jakecorrenti, line 458: I wonder if it would be cleaner to do something like this instead so there's only one `new` associated function? 

```rust
pub fn new(tag: &str, host_path: &str, readonly: bool) -> Result<Self, Error> { .. }
```

src/libkrun/src/api/device_builders.rs
```diff
@@ -0,0 +1,1549 @@
...
+            stdout.as_ref()
+        } else if stderr_is_tty {
+            stderr.as_ref()
+        } else {
+            None
+        };
+
+        // Port 0: default console (hvc0)
+        if let Some(tfd) = term_fd {
+            self.add_tty_port_inner("", tfd.as_fd())?;
```
jakecorrenti: According to an Agent, and I think it's correct, if you have stdin as a tty and stdout is a file, when you call `add_tty_port_inner()` it will use the stdin fd for the terminal's output as well, not just input

src/libkrun/src/api/device_builders.rs
```diff
@@ -0,0 +1,1549 @@
...
+        let host_port_map = if self.host_port_map.is_empty() {
+            None
+        } else {
+            Some(self.host_port_map)
+        };
+        let unix_ipc_port_map = if self.unix_ipc_port_map.is_empty() {
+            None
+        } else {
+            Some(self.unix_ipc_port_map)
+        };
```
jakecorrenti: if you really wanted to you could do something like this
```rust
let host_port_map = (!self.host_port_map.is_empty()).then_some(self.host_port_map);
let unix_ipc_port_map = (!self.unix_ipc_port_map.is_empty()).then_some(self.unix_ipc_port_map);
```

src/libkrun/src/api/device_builders.rs
```diff
@@ -0,0 +1,1549 @@
...
+}
+
+/// Disk image format for block devices.
+#[cfg_attr(feature = "ffi", ffier::export(cfg = "feature = \"blk\""))]
+#[cfg_attr(not(feature = "ffi"), cfg(feature = "blk"))]
+#[repr(u32)]
+#[derive(Clone, Copy, Debug)]
+pub enum DiskFormat {
+    Raw,
+    Qcow2,
```
jakecorrenti: you're missing the vmdk format. additionally, isn't this a re-implementation of the `ImageType` struct in `virtio/block/mod.rs`?

src/libkrun/src/api/device_builders.rs
```diff
@@ -0,0 +1,1549 @@
...
+            mac,
+            features,
+        )
+    }
+
+    // FIXME: use #[cfg(target_os = "linux")] on the method once ffier supports
+    // per-method cfg inside #[cfg_attr(feature = "ffi", ffier::export)] impl blocks.
+    pub fn new_tap(id: &str, tap_name: &str, mac: &[u8], features: u32) -> Result<Self, Error> {
+        #[cfg(target_os = "linux")]
+        {
```
jakecorrenti, line 1157: just gate the method behind the target os so we can avoid it even being called on disabled OSes

src/libkrun/src/api/device_builders.rs
```diff
@@ -0,0 +1,1549 @@
...
+bitflags::bitflags! {
+    #[derive(Debug, Clone, Copy, PartialEq, Eq)]
+    pub struct VirglRendererFlags: u32 {
+        const USE_EGL            = 0x001;
+        const THREAD_SYNC        = 0x002;
+        const VENUS              = 0x040;
+        const USE_ASYNC_FENCE_CB = 0x100;
+        const RENDER_SERVER      = 0x200;
+    }
+}
```
jakecorrenti: What are your thoughts on something like the following:
```rust
#[cfg(feature = "ffi")]
use ffier::export_bitflags;

#cfg(not(feature = "ffi"))]
macro_rules! export_bitflags {
    ($(#[cfg($($cfg:tt)*)])? bitflags::bitflags! { $($body:tt)* }) => {
        $(#[cfg($($cfg)*)])?
        bitflags::bitflags! { $($body)* }
    };
}

...

#[cfg(feature = "gpu")]
export_bitflags! {
    #[derive(Debug, Clone, Copy, PartialEq, Eq)]
    pub struct VirglRendererFlags: u32 {
        const USE_EGL            = 0x001;
        const THREAD_SYNC        = 0x002;
        const VENUS              = 0x040;
        const USE_ASYNC_FENCE_CB = 0x100;
        const RENDER_SERVER      = 0x200;
    }
}
```

This way we just do it once

src/libkrun/src/api/logging.rs
```diff
@@ -0,0 +1,100 @@
...
+    Info = 3,
+    Debug = 4,
+    Trace = 5,
+}
+
+#[cfg_attr(feature = "ffi", ffier::export)]
+#[repr(u32)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum LogStyle {
+    Auto = 0,
```
jakecorrenti: do we want to implement the `Default` trait on these types? Maybe `LogStyle::Auto` and `LogLevel::Error`?

src/libkrun/src/api/logging.rs
```diff
@@ -0,0 +1,100 @@
...
+        LogLevel::Info => "info",
+        LogLevel::Debug => "debug",
+        LogLevel::Trace => "trace",
+    };
+
+    let write_style = match style {
+        LogStyle::Auto => "auto",
+        LogStyle::Always => "always",
+        LogStyle::Never => "never",
+    };
```
jakecorrenti: instead lets implement `Display` for these types so we can convert it to a string:

```rust
impl std::fmt::Display for LogLevel {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        write!(f, "{:?}", self)
    }
}

...

let filter = level.to_string().to_lowercase();
```

src/libkrun/src/api/payload.rs
```diff
@@ -0,0 +1,233 @@
...
+        Ok(Payload {
+            kind: PayloadKind::Kernel { bundle },
+            cmdline: vmm::vmm_config::kernel_cmdline::DEFAULT_KERNEL_CMDLINE.to_string(),
+        })
+    }
+
+    pub fn load_krunfw_tee(tee_config_path: &str) -> Result<Self, Error> {
+        #[cfg(feature = "tee")]
+        {
+            let lib = KRUNFW.as_ref().ok_or_else(|| {
```
jakecorrenti, line 54: IMO we should have just one `load_krunfw()` associated function:

```rust
#[cfg(not(feature = "tee'"))]
pub fn load_krunfw() -> Result<Self, Error> { .. }

#[cfg(feature = "tee")]
pub fn load_krunfw(tee_config_path: &str) -> Result<Self, Error> { .. }
```

src/libkrun/src/api/payload.rs
```diff
@@ -0,0 +1,233 @@
...
+
+    pub fn load_external(
+        kernel_path: &str,
+        format: KernelFormat,
+        initrd_path: Option<&str>,
+        cmdline: &str,
+    ) -> Result<Self, Error> {
+        use vmm::vmm_config::external_kernel::{ExternalKernel, KernelFormat as VmmKernelFormat};
+
+        let vmm_format = match format {
```
jakecorrenti: it seems redundant to re-implement the type

src/libkrun/src/api/payload.rs
```diff
@@ -0,0 +1,233 @@
...
+
+    let mut qboot_size: usize = 0;
+    let qboot_host_addr = unsafe { get_qboot(&mut qboot_size) };
+    if qboot_host_addr.is_null() {
+        return Err(Error::BootError("krunfw_get_qboot returned null".into()));
+    }
+    let qboot_bundle = QbootBundle {
+        host_addr: qboot_host_addr as u64,
+        size: qboot_size,
+    };
```
jakecorrenti, line 187: if the user loaded their own firmware from a specified path, like TDShim, we don't want to also load the qboot bundle (the TDShim changes came after this PR, so you'll end up rebasing on top)

src/libkrun/src/api/vmm_builder.rs
```diff
@@ -0,0 +1,398 @@
...
+}
+
+#[cfg_attr(feature = "ffi", ffier::export)]
+impl<'a> VmmBuilder<'a> {
+    pub fn new() -> Self {
+        VmmBuilder {
+            vcpus: None,
+            ram_mib: None,
+            payload: None,
+            device_manager: None,
```
jakecorrenti:
```rust
#[derive(Default)]
pub struct VmmBuilder<'a> { .. }

impl<'a> VmmBuilder<'a> {
    pub fn new() -> Self {
        VmmBuilder::default()
    }

   ...
}
```

src/libkrun/src/api/vmm_builder.rs
```diff
@@ -0,0 +1,398 @@
...
+        Err(Error::FeatureDisabled())
+    }
+
+    /// Signal the guest to perform an orderly ACPI shutdown.
+    ///
+    /// On macOS this writes to the GPIO device's eventfd, which triggers
+    /// a restart-key press in the guest. On other platforms this is not
+    /// supported.
+    pub fn shutdown(&self) -> Result<(), Error> {
+        #[cfg(target_os = "macos")]
```
jakecorrenti: gate these on macOS at the method level, not inside the method body

jakecorrenti (commented):
Here's an initial review, still chugging along. Feel free to not implement some of these things, a lot of it is clean up

src/libkrun/src/api/vmm_builder.rs
mtjhrc: Yeah unfortunetly that's currently not well supported in the bindings generator `ffier`.  I plan to fix that eventually.

src/libkrun/src/api/payload.rs
mtjhrc: Hmm the idea here is to eventually be able to have both TEE and non-tee support be compiled together. So features are always strictly additive and don't break each other like they currently do.

include/libkrun.h
```diff
@@ -1,1420 +1,659 @@
...
-
-#define KRUN_LOG_STYLE_AUTO 0
-#define KRUN_LOG_STYLE_ALWAYS 1
-#define KRUN_LOG_STYLE_NEVER 2
-
-#define KRUN_LOG_OPTION_NO_ENV 1
+#include <string.h>
+
+typedef void* KrunError;
+typedef void* KrunMmioDeviceManager;
```
jakecorrenti, line 11: `typedef void* KrunError` was declared twice

include/libkrun.h
```diff
@@ -1,1420 +1,659 @@
...
- *
- * Arguments:
- *  "ctx_id"         - the configuration context ID.
- *  "uid"            - a user id to be set.
- *
- * Returns:
- *  Zero on success or a negative error number on failure.
- */
-int32_t krun_setuid(uint32_t ctx_id, uid_t uid);
+KrunDisplayInfoBuilder krun_display_info_builder_new(uint32_t width, uint32_t height);
```
jakecorrenti, line 545: is the intention here by removing `krun_setuid()` and `krun_setgid()` to force the user to use the libc syscall instead?

include/libkrun.h
```diff
@@ -1,1420 +1,659 @@
...
- * Returns:
- *  0        - the pause request was signalled.
- *  -ENOENT  - no running VM is registered for this ctx_id.
- *  -EIO     - failed to signal the VM.
- *  -ENOTSUP - pause is unsupported on this platform.
- */
-int32_t krun_vm_pause(uint32_t ctx_id);
+uint32_t krun_error_code(KrunError handle);
+typedef uint32_t (*krun_error_code_fn)(KrunError handle);
+void krun_error_message(KrunError handle, KrunPushStr writer);
```
jakecorrenti, line 664: `krun_error_code` was declared twice

include/libkrun.h
```diff
@@ -1,1420 +1,659 @@
...
+typedef void (*krun_error_message_fn)(KrunError handle, KrunPushStr writer);
+uint64_t krun_error_result(KrunError handle);
```
jakecorrenti, line 666: `krun_error_message` was declared twice as well.

jakecorrenti:
Do I need to build the examples differently?
```bash
$ make BLK=1 NET=1
$ sudo !! install
$ cd examples
$ make chroot_vm

gcc -o chroot_vm chroot_vm.c -O2 -g -I../include -lkrun -lkrun_init
/usr/bin/ld.bfd: /tmp/ccCGTpTJ.o: in function `main':
/home/jcorrent/libkrun/examples/chroot_vm.c:350:(.text.startup+0x4e): undefined reference to `krun_init_log'
/usr/bin/ld.bfd: /home/jcorrent/libkrun/examples/chroot_vm.c:353:(.text.startup+0x67): undefined reference to `krun_fs_overlay_new'
/usr/bin/ld.bfd: /home/jcorrent/libkrun/examples/chroot_vm.c:356:(.text.startup+0x73): undefined reference to `krun_payload_load_krunfw'
/usr/bin/ld.bfd: /home/jcorrent/libkrun/examples/chroot_vm.c:380:(.text.startup+0x159): undefined reference to `krun_mmio_device_manager_new'
/usr/bin/ld.bfd: /home/jcorrent/libkrun/examples/chroot_vm.c:384:(.text.startup+0x161): undefined reference to `krun_console_device_builder'
/usr/bin/ld.bfd: /home/jcorrent/libkrun/examples/chroot_vm.c:385:(.text.startup+0x182): undefined reference to `krun_console_builder_add_default_console'
/usr/bin/ld.bfd: /home/jcorrent/libkrun/examples/chroot_vm.c:386:(.text.startup+0x199): undefined reference to `krun_console_builder_build'
/usr/bin/ld.bfd: /home/jcorrent/libkrun/examples/chroot_vm.c:387:(.text.startup+0x1a4): undefined reference to `krun_mmio_device_manager_add'
/usr/bin/ld.bfd: /home/jcorrent/libkrun/examples/chroot_vm.c:394:(.text.startup+0x1e4): undefined reference to `krun_vhost_user_device_new'
/usr/bin/ld.bfd: /home/jcorrent/libkrun/examples/chroot_vm.c:396:(.text.startup+0x1ff): undefined reference to `krun_mmio_device_manager_add'
/usr/bin/ld.bfd: /home/jcorrent/libkrun/examples/chroot_vm.c:402:(.text.startup+0x248): undefined reference to `krun_vhost_user_device_new'
/usr/bin/ld.bfd: /home/jcorrent/libkrun/examples/chroot_vm.c:404:(.text.startup+0x263): undefined reference to `krun_mmio_device_manager_add'
/usr/bin/ld.bfd: /home/jcorrent/libkrun/examples/chroot_vm.c:410:(.text.startup+0x2ad): undefined reference to `krun_vhost_user_device_new'
/usr/bin/ld.bfd: /home/jcorrent/libkrun/examples/chroot_vm.c:412:(.text.startup+0x2c8): undefined reference to `krun_mmio_device_manager_add'
/usr/bin/ld.bfd: /home/jcorrent/libkrun/examples/chroot_vm.c:421:(.text.startup+0x313): undefined reference to `krun_vhost_user_device_new'
/usr/bin/ld.bfd: /home/jcorrent/libkrun/examples/chroot_vm.c:423:(.text.startup+0x32e): undefined reference to `krun_mmio_device_manager_add'
/usr/bin/ld.bfd: /home/jcorrent/libkrun/examples/chroot_vm.c:429:(.text.startup+0x379): undefined reference to `krun_vhost_user_device_new'
/usr/bin/ld.bfd: /home/jcorrent/libkrun/examples/chroot_vm.c:431:(.text.startup+0x394): undefined reference to `krun_mmio_device_manager_add'
/usr/bin/ld.bfd: /home/jcorrent/libkrun/examples/chroot_vm.c:437:(.text.startup+0x3dd): undefined reference to `krun_vhost_user_device_new'
/usr/bin/ld.bfd: /home/jcorrent/libkrun/examples/chroot_vm.c:439:(.text.startup+0x3f8): undefined reference to `krun_mmio_device_manager_add'
/usr/bin/ld.bfd: /home/jcorrent/libkrun/examples/chroot_vm.c:445:(.text.startup+0x441): undefined reference to `krun_vhost_user_device_new'
/usr/bin/ld.bfd: /home/jcorrent/libkrun/examples/chroot_vm.c:447:(.text.startup+0x45c): undefined reference to `krun_mmio_device_manager_add'
/usr/bin/ld.bfd: /home/jcorrent/libkrun/examples/chroot_vm.c:453:(.text.startup+0x4a8): undefined reference to `krun_vhost_user_device_new'
/usr/bin/ld.bfd: /home/jcorrent/libkrun/examples/chroot_vm.c:455:(.text.startup+0x4c3): undefined reference to `krun_mmio_device_manager_add'
/usr/bin/ld.bfd: /home/jcorrent/libkrun/examples/chroot_vm.c:462:(.text.startup+0x519): undefined reference to `krun_vhost_user_device_new'
/usr/bin/ld.bfd: /home/jcorrent/libkrun/examples/chroot_vm.c:464:(.text.startup+0x534): undefined reference to `krun_mmio_device_manager_add'
/usr/bin/ld.bfd: /home/jcorrent/libkrun/examples/chroot_vm.c:475:(.text.startup+0x5a0): undefined reference to `krun_fs_device_new'
/usr/bin/ld.bfd: /home/jcorrent/libkrun/examples/chroot_vm.c:476:(.text.startup+0x5bb): undefined reference to `krun_fs_device_set_overlay'
/usr/bin/ld.bfd: /home/jcorrent/libkrun/examples/chroot_vm.c:477:(.text.startup+0x5c6): undefined reference to `krun_mmio_device_manager_add'
/usr/bin/ld.bfd: /home/jcorrent/libkrun/examples/chroot_vm.c:497:(.text.startup+0x631): undefined reference to `krun_net_device_new_unixstream_path'
/usr/bin/ld.bfd: /home/jcorrent/libkrun/examples/chroot_vm.c:512:(.text.startup+0x64c): undefined reference to `krun_mmio_device_manager_add'
/usr/bin/ld.bfd: /home/jcorrent/libkrun/examples/chroot_vm.c:516:(.text.startup+0x65b): undefined reference to `krun_rng_device_new'
/usr/bin/ld.bfd: /home/jcorrent/libkrun/examples/chroot_vm.c:517:(.text.startup+0x672): undefined reference to `krun_mmio_device_manager_add'
/usr/bin/ld.bfd: /home/jcorrent/libkrun/examples/chroot_vm.c:519:(.text.startup+0x681): undefined reference to `krun_balloon_device_new'
/usr/bin/ld.bfd: /home/jcorrent/libkrun/examples/chroot_vm.c:520:(.text.startup+0x698): undefined reference to `krun_mmio_device_manager_add'
/usr/bin/ld.bfd: /home/jcorrent/libkrun/examples/chroot_vm.c:524:(.text.startup+0x69d): undefined reference to `krun_vmm_builder_new'
/usr/bin/ld.bfd: /home/jcorrent/libkrun/examples/chroot_vm.c:525:(.text.startup+0x6bb): undefined reference to `krun_vmm_builder_vcpus'
/usr/bin/ld.bfd: /home/jcorrent/libkrun/examples/chroot_vm.c:526:(.text.startup+0x6d9): undefined reference to `krun_vmm_builder_ram_mib'
/usr/bin/ld.bfd: /home/jcorrent/libkrun/examples/chroot_vm.c:527:(.text.startup+0x6f2): undefined reference to `krun_vmm_builder_payload'
/usr/bin/ld.bfd: /home/jcorrent/libkrun/examples/chroot_vm.c:528:(.text.startup+0x6ff): undefined reference to `krun_vmm_builder_devices'
/usr/bin/ld.bfd: /home/jcorrent/libkrun/examples/chroot_vm.c:529:(.text.startup+0x715): undefined reference to `krun_vmm_builder_split_irqchip'
/usr/bin/ld.bfd: /home/jcorrent/libkrun/examples/chroot_vm.c:531:(.text.startup+0x72e): undefined reference to `krun_vmm_builder_build'
/usr/bin/ld.bfd: /home/jcorrent/libkrun/examples/chroot_vm.c:532:(.text.startup+0x742): undefined reference to `krun_vmm_run'
/usr/bin/ld.bfd: /home/jcorrent/libkrun/examples/chroot_vm.c:504:(.text.startup+0x7d2): undefined reference to `krun_net_device_new_unixstream_fd'
/usr/bin/ld.bfd: /home/jcorrent/libkrun/examples/chroot_vm.c:531:(.text.startup+0x810): undefined reference to `krun_error_message'
/usr/bin/ld.bfd: /home/jcorrent/libkrun/examples/chroot_vm.c:531:(.text.startup+0x837): undefined reference to `krun_error_destroy'
/usr/bin/ld.bfd: /home/jcorrent/libkrun/examples/chroot_vm.c:482:(.text.startup+0x883): undefined reference to `krun_vsock_device_new'
/usr/bin/ld.bfd: /home/jcorrent/libkrun/examples/chroot_vm.c:489:(.text.startup+0x8a6): undefined reference to `krun_mmio_device_manager_add'
/usr/bin/ld.bfd: /home/jcorrent/libkrun/examples/chroot_vm.c:486:(.text.startup+0x8c8): undefined reference to `krun_vsock_device_add_port_forward'
/usr/bin/ld.bfd: /home/jcorrent/libkrun/examples/chroot_vm.c:489:(.text.startup+0x8df): undefined reference to `krun_mmio_device_manager_add'
/usr/bin/ld.bfd: /home/jcorrent/libkrun/examples/chroot_vm.c:350:(.text.startup+0x917): undefined reference to `krun_error_message'
/usr/bin/ld.bfd: /home/jcorrent/libkrun/examples/chroot_vm.c:350:(.text.startup+0x93e): undefined reference to `krun_error_destroy'
/usr/bin/ld.bfd: /home/jcorrent/libkrun/examples/chroot_vm.c:509:(.text.startup+0xa61): undefined reference to `krun_error_destroy'
collect2: error: ld returned 1 exit status
make: *** [Makefile:27: chroot_vm] Error 1
```

include/libkrun.h
mtjhrc: It wasn't intentional; I forgot about them, That said, I'm not sure we should restore them.

The intended v2 direction is to provision all required host resources during device setup and build(), then separate that from starting the VM so callers can drop privileges before actually starting. I still need to look into the lifecycle carefully, including device and vCPU initialization, to make sure that separation actually works properly.

For now, I suppose I can keep them removed?

mtjhrc:
> Do I need to build the examples differently?
> 
> ```shell
> $ make BLK=1 NET=1
> $ sudo !! install
> $ cd examples
> $ make chroot_vm
> 

You need to use `FFI=1` to generate the C API symbols. Without FFI the extern "C" symbols are never generated. I suppose the `make` could automatically enable FFI feature, but it shouldn't be the default in Cargo.toml (you don't want the C api symbols if you are consuming this via native Rust static linking).

```shell
$ make FFI=1 BLK=1 NET=1  VHOST_USER=1
$ sudo !! install
$ cd examples
$ make chroot_vm 
```

src/libkrun/src/api/device_builders.rs
mtjhrc: It was meant so you can chain the methods calls like builder pattern. But yeah I am not sure if I should keep it here.

src/libkrun/src/api/device_builders.rs
mtjhrc: I'll fix this as a follow-up, this currently doesn't fully work with ffier, there are issues with the generated bindings (I think it partially works but there are issues with it).

src/libkrun/src/api/device_builders.rs
mtjhrc: Done! Thanks, seems like a good way to do it!

include/libkrun.h
mtjhrc: Good catch, I'll look into it in the bindings generator side. 


Btw here's whats going on:

So we export these 3 things:
```
crate::api::error::Error = 1,
trait ffier_builtins::Error = 13,
Error for crate::api::error::Error,
```

Problem is both the trait and the concrete error type are called `Error`, so combined with the prefix both generate `KrunError`. This also means we end up generating  both `krun_error_code()` and `krun_error_message()` twice. 

They are generated twice:
- once for the concrete type 
-  once for any possible error which implements the error trait (so the function is generic over any error type (that's what the comment `/* KrunError | KrunVtableError */` means).

jakecorrenti:
I noticed I was getting the following error when running the `chroot_vm` and `launch-tee` examples: `[2026-09-08T15:49:48.182592Z ERROR krun::vmm::builder] Failed to restore terminal mode: ENOTTY: Not a typewriter`.

I had Grok do some debugging and it came up with this: 

On VM shutdown, libkrun tries to restore the host terminal’s original settings after putting it in raw mode for the virtio console. That restore
  is failing with ENOTTY because the fd it tries to use is no longer a TTY (or no longer the same fd).

  The log comes from here:

   src/libkrun/src/vmm/builder.rs lines 2234-2238

                  vmm.exit_observers.push(Arc::new(Mutex::new(move || {
                      if let Err(e) =
                          term_restore_mode(unsafe { BorrowedFd::borrow_raw(raw_fd) }, &old_mode)
                      {
                          log::error!("Failed to restore terminal mode: {e}")

  What changed in the PR

  v1 (main) called setup_terminal_raw_mode on the original stdio fds (STDIN_FILENO, etc.) passed by the caller. Those fds stay open for the whole
  process lifetime.

  v2 (this PR) does two things differently:

  1. chroot_vm now passes dups instead of the original fds:

   examples/chroot_vm.c lines 381-381

          CHECK(krun_console_builder_add_default_console(cb, dup(STDIN_FILENO), dup(STDOUT_FILENO), dup(STDERR_FILENO), &krun_err));

  2. ConsoleDevice::attach() puts terminal fds in tty_fds, enables raw mode on them, then drops ConsoleDevice when attach finishes — which
     closes those fds:

   src/libkrun/src/api/device_builders.rs lines 850-853

          for fd in &self.tty_fds {
              ctx.setup_terminal_raw_mode(fd.as_fd());
          }
          Ok(())

  After attach, tty_fds is dropped and those fds are closed, but the exit observer still holds the old fd number. On shutdown, tcsetattr() runs
  against a closed/reused fd → ENOTTY.


I also ran into a couple issues with the `launch-tee` example, but something along these lines would fix it:

```diff
diff --git a/examples/launch-tee.c b/examples/launch-tee.c
index d5032bf8..16620618 100644
--- a/examples/launch-tee.c
+++ b/examples/launch-tee.c
@@ -105,8 +105,9 @@ int main(int argc, char *const argv[])
     }

     // Load the TEE payload from the config file.
+    KrunStr no_firmware = { .data = NULL, .len = 0 };
     CHECK(KrunPayload payload = krun_payload_load_krunfw_tee(KRUN_STR(argv[optind + 1]),
-        td_shim_path ? KRUN_STR(td_shim_path) : KRUN_STR(""), &krun_err));
+        td_shim_path ? KRUN_STR(td_shim_path) : no_firmware, &krun_err));
     krun_payload_append_cmdline(payload, KRUN_STR("init=/init.krun KRUN_RLIMITS=6=4096:8192 KRUN_WORKDIR=/"));

     // Build the VM.
diff --git a/src/libkrun/src/api/payload.rs b/src/libkrun/src/api/payload.rs
index 0a003d3e..4ecf764e 100644
--- a/src/libkrun/src/api/payload.rs
+++ b/src/libkrun/src/api/payload.rs
@@ -64,7 +64,8 @@ impl Payload {

             let bundle = load_kernel_bundle(lib)?;
             #[cfg(feature = "tdx")]
-            let (qboot_bundle, initrd_bundle) = if firmware_path.is_some() {
+            let use_td_shim = firmware_path.is_some_and(|s| !s.is_empty());
+            let (qboot_bundle, initrd_bundle) = if use_td_shim {
                 (None, load_initrd_bundle(lib)?)
             } else {
                 let (qboot, initrd) = load_tee_bundles(lib)?;
diff --git a/src/libkrun/src/api/vmm_builder.rs b/src/libkrun/src/api/vmm_builder.rs
index 1c0307d5..71747372 100644
--- a/src/libkrun/src/api/vmm_builder.rs
+++ b/src/libkrun/src/api/vmm_builder.rs
@@ -395,7 +395,7 @@ fn build_vm(builder_cfg: VmmBuilder<'_>) -> Result<Vmm<'_>, Error> {

     let mut event_manager = EventManager::new().map_err(|e| Error::Internal(format!("{e:?}")))?;

-    let (sender, _receiver) = unbounded();
+    let (sender, receiver) = unbounded();

     #[cfg(all(target_arch = "aarch64", target_os = "macos"))]
     let shutdown_efd = if builder_cfg.shutdown_support {
@@ -420,6 +420,19 @@ fn build_vm(builder_cfg: VmmBuilder<'_>) -> Result<Vmm<'_>, Error> {
     )
     .map_err(|e| Error::BootError(format!("{e:?}")))?;

+    #[cfg(target_arch = "x86_64")]
+    if builder_cfg.split_irqchip {
+        crate::vmm::worker::start_worker_thread(inner.clone(), receiver.clone()).map_err(
+            |e| Error::Internal(format!("worker thread: {e}")),
+        )?;
+    }
+
+    #[cfg(any(feature = "amd-sev", feature = "tdx"))]
+    crate::vmm::worker::start_worker_thread(inner.clone(), receiver.clone()).map_err(|e| {
+        Error::Internal(format!("worker thread: {e}"))
+    })?;
+
+
     Ok(Vmm {
         inner: VmmInner::Vmm {
             vmm: inner,
```

mtjhrc:
With the help of Gemini 3.7 flash (btw Gemini seems to be better at that then Claude)
- rebased on main
- I tried to fix the fd lifetime issues with consoles. The console APIs now accept BorrowedFd<'a> (instead of `OwnedFd`) where 'a is the lifetime of the console device. Eventually the goal is to have proper annations that track the lifetime through the codebase. Basically I want to try propagating a 'vmm lifetime internally as much as possible. But this requires cleaning up the devices code. (this is also related with the goal of being able to stop the vm without `_exit`-ing the process.)
- renamed `Error` -> `VmmError` (this is a work around for the duplicate definitions in the header)
- applied Jake's TDX patch (with some modification) 

Hopefully this should now be good enough to merge. There might be more stuff broken (also I rebased this multiple times today with Gemini... hopefully I didn't miss/create more issues). But like we talked about we should fix that in follow up PRs because iterating on this huge PR is becoming quite tedious and very hard to track.

@slp

mtjhrc:
There are some CI failures. It seems like the apt repo for google chrome is broken (wrong checksum) which makes `apt update` fail in the CI - hopefully it fixes itself tomorrow.

jakecorrenti (approved):
functionally looks good to me :)

jakecorrenti:
Re-ran the tests

src/libkrun/src/vmm/builder.rs
```diff
@@ -2364,280 +2221,24 @@ fn attach_mmio_device(
...
-                port_io::output_to_raw_fd_dup(err_fd).unwrap(),
-            ));
-        }
-
-        Ok(ports)
-    }
-}
-
-#[cfg(windows)]
-fn is_valid_handle(h: *mut core::ffi::c_void) -> bool {
```
lstocchi, line 2513: This got lost during the refactor

slp (approved):
LEEEROOOOOYYY JENNNKIIINSS

lstocchi (commented):
too late 😆 I was porting the new part on Win to test it. I saw we got lost something but i'll re-add them in follow ups. A mess without CI and the project that cannot be compiled. For the rest it seemed OK :) 

 > LEEEROOOOOYYY JENNNKIIINSS

A WoW player 😆 Wave from a human priest. Happy old days 😆

PR #721, opened 2026-06-11
ci: trust libkrun/krun tap before installing formulae on macOS

jakecorrenti:
Homebrew now refuses to install formulae from third-party taps that haven't been explicitly trusted, breaking all macOS CI jobs.


slp:
While there, could you please also update it to use the new repo (`libkrun/krun`)?

slp (approved):
Thanks a lot, @jakecorrenti !

PR #723, opened 2026-06-13
virtio-fs: restore the original thread credentials, not euid/egid 0

hansjorg:
`scoped_cred`'s `Drop` resets the worker thread's effective uid/gid to a
hardcoded **0**, which assumes the server runs as root. When the VMM runs as
an unprivileged user that has been granted `CAP_SETUID`/`CAP_SETGID` (e.g.
via systemd's `AmbientCapabilities=`), the credential-switching path is
active (`set_creds` sees the capability), but the hardcoded restore breaks
it:

1. The first switched request moves euid from the server's uid to the guest
   uid; both are non-zero, so the thread's effective capability set
   survives. Its `Drop` then sets euid to **0** — silently parking the
   worker thread with host-root credentials.
2. The next switched request moves euid 0 → non-zero, which **clears the
   thread's effective capability set** (capabilities(7), *"Effect of user ID
   changes on capabilities"*).
3. That request's `Drop` now fails with EPERM (CAP_SETUID is gone), and the
   thread is permanently stuck with the guest uid's credentials. Every
   subsequent request handled by this thread — including requests on behalf
   of guest root — fails EACCES/EPERM against host files that uid cannot
   access.

Any guest workload issuing requests as more than one uid wedges the
filesystem within a few operations. A Debian guest running `apt-get update`
(whose fetch methods drop privileges to `_apt`) reproduces it reliably: once
a worker thread is poisoned, the InRelease/Packages renames fail with
`rename failed, Permission denied` even though they are issued by guest
root.

The fix captures the thread's effective id before switching and restores
that value on drop. For a root-run server the captured id is 0 and behavior
is unchanged. For an unprivileged server with caps the thread only ever
switches between non-zero uids, which never clears the effective capability
set, so switching keeps working — and the worker thread no longer escalates
itself to host euid 0 as a side effect (also a hardening win: an
unprivileged-by-design VMM should not quietly end up doing host filesystem
operations as euid 0).

Found while running libkrun-based microVMs under a systemd service with
`AmbientCapabilities=CAP_SETUID CAP_SETGID CAP_CHOWN CAP_DAC_OVERRIDE`;
verified against v1.18.1.

### DIFF
--- a/src/devices/src/virtio/fs/linux/passthrough.rs
+++ b/src/devices/src/virtio/fs/linux/passthrough.rs
@@ -2159,3 +2159,112 @@ impl FileSystem for PassthroughFs {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    // A non-zero uid the server can be made to run as, plus a guest
+    // uid distinct from it and from root. Both must be mappable when
+    // the test runs inside a user namespace (e.g. `buildah unshare`),
+    // so keep them inside the conventional 65536-wide subid range.
+    const SERVER_UID: libc::uid_t = 1;
+    const SERVER_GID: libc::gid_t = 1;
+    const GUEST_UID: libc::uid_t = 1000;
+
+    /// Regression test for the `scoped_cred` credential restore.
+    ///
+    /// The fs server switches the worker thread's effective uid per
+    /// request (`ScopedUid`) and restores it on drop. When the server
+    /// runs as a non-root user that holds CAP_SETUID — e.g. an
+    /// unprivileged daemon granted the capability via systemd's
+    /// `AmbientCapabilities=` — restoring to a hardcoded euid 0 (the
+    /// old behavior) wedges the thread: the next switch from euid 0 to
+    /// a non-zero uid clears its effective capability set, the restore
+    /// after that fails EPERM, and the thread is stranded at the guest
+    /// uid, failing every later request (including ones for guest
+    /// root). Restoring the *previous* euid keeps switching between
+    /// non-zero uids only, which never clears the capabilities.
+    ///
+    /// This drives the real `ScopedUid` through repeated switches and
+    /// asserts the thread is left back at the server uid. Credential
+    /// changes are per-thread (raw syscalls, by design), so it runs on
+    /// a dedicated thread and leaves the rest of the test binary
+    /// untouched. It needs CAP_SETUID and the ability to move to a
+    /// non-zero uid; without them (a plain `cargo test` as a normal
+    /// user) it skips rather than fails.
+    #[test]
+    fn scoped_uid_restores_server_uid_without_wedging() {
+        let outcome = std::thread::spawn(scoped_uid_no_wedge_body)
+            .join()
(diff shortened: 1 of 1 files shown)


src/devices/src/virtio/fs/linux/passthrough.rs
```diff
@@ -2145,3 +2159,112 @@ impl FileSystem for PassthroughFs {
...
+    /// substrate could not be set up and the test should skip; a real
+    /// regression (the thread left wedged) panics via `assert` so the
+    /// test fails.
+    fn scoped_uid_no_wedge_body() -> Result<(), String> {
+        if !has_cap(None, CapSet::Effective, Capability::CAP_SETUID).map_err(|e| e.to_string())? {
+            return Err("missing CAP_SETUID (run as root or under a userns with caps)".into());
+        }
+
+        // Keep the permitted caps across the upcoming uid change.
+        if unsafe { libc::prctl(libc::PR_SET_KEEPCAPS, 1, 0, 0, 0) } != 0 {
```
mz-pdm, line 2218: I wonder about the zeros here, my `man PR_SET_KEEPCAPS` mentions only a single argument. But the tests works for me with or without the zeros.

slp:
Thanks @hansjorg ! The fix looks good, but I don't think the unit test can be used to check against regressions since it can only be reproduced in a very specific context, so I'd suggest dropping it.

mz-pdm (approved):
Thank you for the fix, looks good to me.

mz-pdm:
> I don't think the unit test can be used to check against regressions since it can only be reproduced in a very specific context, so I'd suggest dropping it.

@slp The test works for me in a krun Podman container, which is not that specific. So I think it's still useful.

slp:
> > I don't think the unit test can be used to check against regressions since it can only be reproduced in a very specific context, so I'd suggest dropping it.
> 
> @slp The test works for me in a krun Podman container, which is not that specific. So I think it's still useful.

Does it fail if you revert the fix?

mz-pdm:
> Does it fail if you revert the fix?

Yes:

```
thread '<unnamed>' (1730) panicked at src/devices/src/virtio/fs/linux/passthrough.rs:2248:9:
assertion `left == right` failed: worker thread wedged after repeated switches: euid was not restored to the server uid (the scoped_cred Drop regressed to restoring euid 0)
  left: 1000
 right: 1

thread 'virtio::fs::linux::passthrough::tests::scoped_uid_restores_server_uid_without_wedging' (1729) panicked at src/devices/src/virtio/fs/linux/passthrough.rs:2186:14:
worker thread panicked: Any { .. }
test virtio::fs::linux::passthrough::tests::scoped_uid_restores_server_uid_without_wedging ... FAILED
```

(Yes, it's easy to miss if the test just skips, but I checked previously different scenarios and the test looks doing what it is expected to do.)

slp:
> > Does it fail if you revert the fix?
> 
> Yes:

OK, it's not the usual context in which we run the tests in CI, but let's keep it anyways.

slp (approved):

PR #725, opened 2026-06-15
init: enforce static linking

### LINKED ISSUE #724: Linking issue with the new rust init.krun depending on the image
When running `docker.io/cloudflare/cloudflared:latest` with the latest libkrun build I got:

```console
[2026-06-14T21:08:36.252580Z ERROR init_or_kernel] /init.krun: error while loading shared libraries: libgcc_s.so.1: cannot open shared object file: No such file or directory
```

It turns out the rust-built init.krun is **not** a static binary and links to libgcc dynamically. The init binary only works if the libgcc inside the image and the libgcc from the linking environment of libkrun are ABI-compatible. I built my libkrun under Arch Linux environment and apparently that is not compatible with whatever version the cloudflared image is shipping.

Reverting to c version of init fixes this currently.

chaserhkj:
change krun-init to require static linking, since dynamic targets are unreliable and inconsistent within container filesystems.

Fixes #724

### DIFF
--- a/.github/actions/setup-build-env/action.yml
+++ b/.github/actions/setup-build-env/action.yml
@@ -13,6 +13,16 @@ runs:
         rustup toolchain install nightly
         rustup component add rust-src --toolchain nightly
 
+    - name: Add musl target
+      shell: bash
+      run: |
+        case $(uname -m) in
+          x86_64)  TARGET="x86_64-unknown-linux-musl"  ;;
+          aarch64|arm64) TARGET="aarch64-unknown-linux-musl" ;;
+          *)       echo "unknown arch: $(uname -m)"; exit 1 ;;
+        esac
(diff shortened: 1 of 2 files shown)


jakecorrenti:
looks like the changes definitely work :) @chaserhkj seems like we need to update the CI workflows as well to make sure that we're adding the musl target

jakecorrenti:
Hi @chaserhkj are you still planning on following up on this? If not one of us can take it over to fix the problem.

chaserhkj:
Sorry this was on my backlog for quite a while and haven't looked at it yet, will take a look today or tomorrow

chaserhkj:
Updated CI after rebasing to main, @jakecorrenti PTAL

jakecorrenti (approved):
LGTM! Thanks for the update

PR #728, opened 2026-06-17
[stable-1.19.x] deps: fix vm-memory version conflict when used as a Rust dependency

### LINKED ISSUE #726: Unable to compile `libkrun` post v1.16.0 as cargo dependency - transient dependency version mismatches
Any release post v1.16.0 exhibits multiple versions of `vm-memory` being used and consequently missing trait bound implementations. When linking `libkrun = { git = "https://github.com/containers/libkrun", features = ["blk", "net", "input", "gpu"], tag = "v1.19.0" }`.

jakecorrenti:
When libkrun is added as a Rust crate dependency, Cargo resolves two incompatible versions of vm-memory (0.17.1 and 0.18.0), causing compilation failures. The root cause is imago's wide vm-memory constraint (>=0.16, <0.19) which lets Cargo pick 0.18.0 for imago while everything else resolves to 0.17.1. Within the workspace this is hidden by Cargo.lock, but external consumers don't inherit it.

Drop imago's vm-memory feature and manually convert VolatileSlice to IoVector/IoVectorMut using ptr_guard()/ptr_guard_mut(). Pin all vm-memory dependencies to =0.17.1 to prevent version splits from newly published semver-compatible releases (e.g., 0.17.2).

Fixes: #726

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -478,9 +478,9 @@ checksum = "2304e00983f87ffb38b55b444b5e3b60a884b5d30c0fca7d82fe33449bbe55ea"
 
 [[package]]
 name = "imago"
-version = "0.2.2"
+version = "0.2.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6e8e4b92aa0dd860579cfba776dbf0918a3a7ac5cb601af7d3fc835e71592a5b"
+checksum = "ae7cfee876c698a1a2ed9c705ab18f21acbed82110f19b51cc458de73426fe2c"
 dependencies = [
  "async-trait",
  "bincode",
@@ -492,7 +492,6 @@ dependencies = [
  "rustc_version",
  "tokio",
  "tracing",
- "vm-memory",
  "windows-sys",
 ]
 
@@ -786,7 +785,7 @@ dependencies = [
  "serde_json",
  "tdx",
  "vm-memory",
- "vmm-sys-util 0.14.0",
+ "vmm-sys-util 0.15.0",
  "zstd",
 ]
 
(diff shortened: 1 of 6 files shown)


drahnr:
I can confirm locally vendoring and patching `imago @ v0.2.2`, pinning the re-export of `vm-memory` to `0.17.1` works as expected for `libkrun@v1.19.0`

slp:
I think we should wait for a new release of linux-loader and then update our crates to be at 0.18.0 instead.

jakecorrenti:
> I think we should wait for a new release of linux-loader and then update our crates to be at 0.18.0 instead.

We'll need to also wait for https://github.com/rust-vmm/vhost/pull/348

drahnr:
Given that the project is hardly usable without, an interim patch release would be very appreciated.

slp:
> > I think we should wait for a new release of linux-loader and then update our crates to be at 0.18.0 instead.
> 
> We'll need to also wait for [rust-vmm/vhost#348](https://github.com/rust-vmm/vhost/pull/348)

Hm... that one is going to take longer. OK, what do you think about retargeting this change for `stable-1.19.x`? Then we can do the right thing for `main`.

slp:
> Given that the project is hardly usable without, an interim patch release would be very appreciated.

This project has existed for years before even having a release on crates.io, and I think many people have found it usable.

jakecorrenti:
> > > I think we should wait for a new release of linux-loader and then update our crates to be at 0.18.0 instead.
> > 
> > 
> > We'll need to also wait for [rust-vmm/vhost#348](https://github.com/rust-vmm/vhost/pull/348)
> 
> Hm... that one is going to take longer. OK, what do you think about retargeting this change for `stable-1.19.x`? Then we can do the right thing for `main`.

I think that's a good compromise. I'll get around to it today

jakecorrenti:
> Hm... that one is going to take longer. OK, what do you think about retargeting this change for `stable-1.19.x`? Then we can do the right thing for `main`.

Should be set now

slp:
thanks @jakecorrenti !

slp (approved):

PR #729, opened 2026-06-17
init: filter tsi_hijack tokens from kernel cmdline args

jakecorrenti:
The Linux kernel forwards unrecognised bare-word parameters from before the `--` delimiter as init arguments. The vmm places tsi_hijack and tsi_hijack_unix on the kernel cmdline, and these leak into the user's command as spurious arguments, causing errors like "/bin/sh: tsi_hijack_unix: No such file or directory".

Filter out tsi_hijack* tokens from the init's argv before building the exec arguments. Also update tsi_enabled() to match any tsi_hijack prefix so it recognises tsi_hijack_unix.

This issue presented itself once we no longer allowed for implicit resources to be created.

### DIFF
--- a/init/src/env.rs
+++ b/init/src/env.rs
@@ -99,8 +99,8 @@ fn setup_dhcp(iface: &str, sock: i32) {
     }
 }
 
-/// Returns true if `tsi_hijack` appears in the kernel command line before any
-/// `--` delimiter. Mirrors `tsi_enabled()` in init.c.
+/// Returns true if any `tsi_hijack` token appears in the kernel command line
+/// before the `--` delimiter.
 #[cfg(target_os = "linux")]
 pub fn tsi_enabled() -> bool {
     let Ok(cmdline) = std::fs::read_to_string("/proc/cmdline") else {
@@ -109,7 +109,7 @@ pub fn tsi_enabled() -> bool {
     cmdline
         .split_whitespace()
         .take_while(|tok| *tok != "--")
-        .any(|tok| tok == "tsi_hijack")
+        .any(|tok| tok.starts_with("tsi_hijack"))
 }
 
 /// Brings up `dummy0` and assigns it 203.0.113.1/24 (IANA TEST-NET-3) so
--- a/init/src/main.rs
+++ b/init/src/main.rs
@@ -121,7 +121,16 @@ fn main() -> anyhow::Result<()> {
     // The kernel places everything after `--` in the cmdline as this
     // process's argv[1..].  The C init built exec_argv by replacing argv[0]
     // with KRUN_INIT (or /bin/sh) and keeping argv[1..] in every branch.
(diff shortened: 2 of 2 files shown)


mtjhrc (commented):
IMO putting the user payload args into kernel args is inherently broken and my plan with this (and I already talked to @slp about it) is that we will remove the kernel command line parsing for the user payload completely. User payload will always be provided via the json. To simplify this we will have helper functions for constructing the json (and also have a function for setting an OCI json string). This is mostly already implemented in #706. 

Btw while working on that PR I noticed `krun_set_exec/krun_set_env/krun_set_workdir/krun_set_rlimits` are also used for aws nitro and I think it forwards them via vsock. Could we also use the same json format for aws nitro too? (and for the TEE variants too). I think it would be quite nice to unify the payload configuration format to our custom json schema (based on OCI for now, but we can add other fields for TEE or anything if we need them). In the aws nitro variant we could send the json via vsock, no? Wdyt?

jakecorrenti:
> IMO putting the user payload args into kernel args is inherently broken and my plan with this (and I already talked to @slp about it) is that we will remove the kernel command line parsing for the user payload completely. User payload will always be provided via the json. To simplify this we will have helper functions for constructing the json (and also have a function for setting an OCI json string). This is mostly already implemented in #706.
> 
> Btw while working on that PR I noticed `krun_set_exec/krun_set_env/krun_set_workdir/krun_set_rlimits` are also used for aws nitro and I think it forwards them via vsock. Could we also use the same json format for aws nitro too? (and for the TEE variants too). I think it would be quite nice to unify the payload configuration format to our custom json schema (based on OCI for now, but we can add other fields for TEE or anything if we need them). In the aws nitro variant we could send the json via vsock, no? Wdyt?

I don't have an issue with it, and I don't see why that wouldn't be possible.

@tylerfanelli what are your thoughts on this as well?

jakecorrenti:
> IMO putting the user payload args into kernel args is inherently broken and my plan with this (and I already talked to @slp about it) is that we will remove the kernel command line parsing for the user payload completely. User payload will always be provided via the json. To simplify this we will have helper functions for constructing the json (and also have a function for setting an OCI json string). This is mostly already implemented in #706.

Since that PR is still in draft how do we want to handle it in the meantime? Isn't main technically broken if someone just uses `krun_add_vsock`? 

> 
> Btw while working on that PR I noticed `krun_set_exec/krun_set_env/krun_set_workdir/krun_set_rlimits` are also used for aws nitro and I think it forwards them via vsock. Could we also use the same json format for aws nitro too? (and for the TEE variants too). I think it would be quite nice to unify the payload configuration format to our custom json schema (based on OCI for now, but we can add other fields for TEE or anything if we need them). In the aws nitro variant we could send the json via vsock, no? Wdyt?

PR #730, opened 2026-06-19
virtio/fs: add support for Windows

lstocchi:
This is probably the biggest and most problematic piece to port on Windows. So i was not sure how to organize the PR.

I had to split it somehow to make the review "easier". Claude suggested this way, but if you have other preferences let me know.

I am planning to make a comparison with 9P (used by podman on Hyperv) to see the performances we have here.
I'm quite satisfied tbh, so iexpect great stats. There is always chance to improve it but i wanted to push it as it's easier to have a foundation to work on later.


slp (approved):
While I haven't tested it, the changes look good and are nicely isolated, so let's get it merged. Thanks @lstocchi !

PR #732, opened 2026-06-19
vmm: properly gate edk2 import, fix crate.io build

nohajc:
I had more trouble compiling libkrun from crates.io, related to the EFI feature. This should fix it for the stable release.


slp (approved):
LGTM, thanks @nohajc !

PR #733, opened 2026-06-22
virtio/blk: add Windows support for block device

lstocchi:
Port the block device to Windows by providing platform-specific implementations for disk identity and file descriptor access.

On Windows, build_device_id() uses GetFileInformationByHandle to derive a unique device identifier from volume serial number and file index, mirroring the Unix approach that combines st_dev, st_rdev, and st_ino.


slp:
Please rebase to fix the tests.

slp:
Code LGTM, thanks!

slp (approved):

PR #734, opened 2026-06-22
virtio-fs: map host process UID/GID to root in guest on macOS

benoitf:
When using virtiofs on macOS, pre-existing host files that lack the user.containers.override_stat xattr are returned to the guest with the raw host UID/GID (typically 501 on macOS). This causes volume-mounted directories to appear as owned by uid 501 instead of root inside the guest VM, breaking container workloads that expect root ownership.

Apple's Virtualization.framework maps the host user's UID to root in the guest. This commit aligns libkrun's macOS virtiofs passthrough with that behavior: when no xattr override is present, files owned by the host process's EUID/EGID are reported as owned by root (uid/gid 0) to the guest.

Files that already have the override_stat xattr (e.g. files created or chowned by the guest) are unaffected — the xattr values take precedence. Files owned by other users on the host retain their raw UIDs since they don't match the process EUID.

The host EUID/EGID values are cached via OnceLock to avoid repeated syscalls on every stat operation.

Fixes: https://github.com/containers/podman/issues/21402

PR #735, opened 2026-06-22
virtio-fs: map host process UID/GID to root in guest on macOS (1.19.x)

benoitf:
Here I'm using a PR to 1.19.x branch as krunkit is not really booting using the main branch and I would like to have the fix available in krunkit
backport of https://github.com/libkrun/libkrun/pull/734 to a branch

When using virtiofs on macOS, pre-existing host files that lack the user.containers.override_stat xattr are returned to the guest with the raw host UID/GID (typically 501 on macOS). This causes volume-mounted directories to appear as owned by uid 501 instead of root inside the guest VM, breaking container workloads that expect root ownership.

On macOS, libkrun's virtiofs passthrough stores guest-visible ownership in an xattr (user.containers.override_stat). Files created or chowned by the guest get this xattr set, so subsequent stat() calls return the correct guest UID/GID. However, pre-existing host files never have this xattr, so stat_common() falls through and returns the raw macOS stat values — exposing the host user's UID (e.g. 501) to the guest.

Apple's Virtualization.framework (used by the applehv provider) does not exhibit this problem because it maps the host user's UID to root in the guest. This commit aligns libkrun's behavior: when no xattr override is present and the file's UID/GID matches the host process's EUID/EGID, report it as root (uid/gid 0) to the guest.

Files that already have the override_stat xattr (e.g. files created or chowned by the guest) are unaffected — the xattr values take precedence. Files owned by other users on the host retain their raw UIDs since they don't match the process EUID. The host EUID/EGID values are cached via OnceLock to avoid repeated syscalls on every stat operation.

Fixes: https://github.com/containers/podman/issues/21402

### DIFF
--- a/src/devices/src/virtio/fs/macos/passthrough.rs
+++ b/src/devices/src/virtio/fs/macos/passthrough.rs
@@ -161,6 +161,16 @@ fn einval() -> io::Error {
     linux_error(io::Error::from_raw_os_error(libc::EINVAL))
 }
 
+fn host_euid() -> u32 {
+    static EUID: std::sync::OnceLock<u32> = std::sync::OnceLock::new();
+    *EUID.get_or_init(|| unsafe { libc::geteuid() })
+}
+
+fn host_egid() -> u32 {
+    static EGID: std::sync::OnceLock<u32> = std::sync::OnceLock::new();
+    *EGID.get_or_init(|| unsafe { libc::getegid() })
+}
+
 fn item_to_value(item: &[u8], radix: u32) -> Option<u32> {
     match std::str::from_utf8(item) {
         Ok(val) => match u32::from_str_radix(val, radix) {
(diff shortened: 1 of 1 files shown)


jakecorrenti (changes requested):
Same comment as https://github.com/libkrun/libkrun/pull/734#pullrequestreview-4544442977

Also: if there's an issue booting krunkit with the main branch please file an issue so it can be investigated.

benoitf:
@jakecorrenti I've updated the commit on both PRs

on my end I checked with podman 6 RC1


```
podman run --entrypoint=sh -it -v $HOME/tmp/data:/etcd quay.io/coreos/etcd:v3.5.0 -c "ls -la /"                                                                                                     total 8
dr-xr-xr-x.   1 root root   29 Jun 22 13:17 .
dr-xr-xr-x.   1 root root   29 Jun 22 13:17 ..
-rwxr-xr-x.   1 root root    0 Jan 27  2021 .dockerenv
drwxr-xr-x.   2 root root 4096 Jan 27  2021 bin
drwxr-xr-x.   2 root root    6 Nov 22  2020 boot
drwxr-xr-x.   5 root root  360 Jun 22 13:17 dev
drwxr-xr-x.  25 root root 4096 Jan 27  2021 etc
drwxr-xr-x.   2 root root   64 Jun 22 08:39 etcd
```

`etcd` folder has now `root.root`

benoitf:
> Same comment as [#734 (review)](https://github.com/libkrun/libkrun/pull/734#pullrequestreview-4544442977)
> 
> Also: if there's an issue booting krunkit with the main branch please file an issue so it can be investigated.

I opened https://github.com/libkrun/krunkit/issues/106

jakecorrenti:
@benoitf if you rebase https://github.com/libkrun/libkrun/pull/742 should fix the code quality failure

jakecorrenti (approved):

benoitf:
@jakecorrenti ok rebased

benoitf:
> 5 workflows awaiting approval

slp:
Let's cherry-pick #734 once merged instead of merging this one to `stable-1.19.x`, so we have a reference of the commits that are also present in `main`.

benoitf:
sure 👍

benoitf:
is there @Mergifyio installed ?
to use https://docs.mergify.com/commands/backport/#syntax command

PR #736, opened 2026-06-22
docs: add CLAUDE.md as symlink to AGENTS.md

benoitf:
in my previous PRs some guidelines were not used as it's not read by default. I added a symlink.

Claude Code reads CLAUDE.md, not AGENTS.md. Add a symlink so the existing AGENTS.md content is picked up automatically.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,1 @@
+AGENTS.md


jakecorrenti (commented):
We have https://github.com/libkrun/libkrun/blob/main/.claude/CLAUDE.md which will inline the contents of AGENTS.md and should work as-is. I'm not sure why you're agent wasn't loading the context

benoitf:
@jakecorrenti maybe because you're not using the good path location ?

https://code.claude.com/docs/en/claude-directory#choose-the-right-file

benoitf:
anyway, I'll close my PR, was trying to help (this new PR used the correct configuration)

jakecorrenti:
I'm looking into it some more, maybe I have some memory cached that's making it work for me

jakecorrenti:
@benoitf can you try this out: https://github.com/libkrun/libkrun/pull/738

PR #739, opened 2026-06-22
virtio/console: fix TX data loss race on shutdown

jakecorrenti:
The 'Integration Tests (Linux x86_64)' CI job has been failing on all recent PRs.  The most basic test, configure-vm-2cpu-1GiB, produces an empty stdout.txt despite the VM booting cleanly and the guest-agent running successfully.  The diagnostic log shows:

  TRACE process_tx] tx unparked, queue len 0
  INFO  krun_vmm] Vmm is stopping.

The TX thread woke up, saw an empty queue, and exited — losing the guest's final println!("OK") output.

Root cause: race in pop_head_blocking
--------------------------------------

The TX thread parks via thread::park() when the queue is empty and is woken either by notify_tx() (from the event-manager thread, which fires when the guest writes the virtio queue notification eventfd) or by port::shutdown() (which sets stop=true then calls unpark()).

The old code:
```rust
  loop {
      match queue.pop(mem) {
          Some(d) => break Some(d),
          None => {
              interrupt.signal_used_queue();
              if stop.load(Acquire) { break None; }   // exits here
              thread::park();
          }
      }
  }
```

The race window:

  TX thread               Shutdown thread         Event-manager thread
  ─────────────────────────────────────────────────────
  queue.pop() -> None
  interrupt.signal_used_queue()
  stop.load() -> false    stop.store(true)
  thread::park()          tx_thread.unpark()
  <- wakes
  queue.pop() -> None     (guest wrote "OK\n"; notification in flight)
  stop.load() -> true
  break None <- EXITS                             notify_tx() -> unpark()
                                                  [TX thread already gone]

The guest's final bytes are in the virtio avail ring when the TX thread exits.  The event-manager's notify_tx() arrives after termination.  The stdout.txt file is empty; check() returns TestOutcome::Fail.

Why only x86_64?
-----------------

The aarch64 self-hosted runner sets KRUN_NO_UNSHARE=1, skipping the buildah-unshare + network-namespace wrapper.  Without that wrapper the per-test process lifetime is different (longer VM teardown path), making the race window substantially narrower.  It does not surface in practice on aarch64.

Fix: drain the queue with the disable/enable_notification fence ---------------------------------------------------------------

The correct pattern — already used by the block and net device workers (block/worker.rs, net/worker.rs) — is to suppress guest notifications while draining, then call enable_notification() which does a SeqCst fence and re-reads avail_idx.  If new entries arrived while notifications were suppressed, enable_notification returns true and we drain again.  Only after the fence confirms the queue is empty do we check stop.  If stop is true at that point, no bytes remain unprocessed and we can return None safely.

The stop check is deliberately placed after the drain+fence: if stop was set while the guest was writing its last bytes, we flush them before exiting.

Assisted-by: Claude Code: claude-sonnet-4-6

### DIFF
--- a/src/devices/src/virtio/console/process_tx.rs
+++ b/src/devices/src/virtio/console/process_tx.rs
@@ -64,17 +64,36 @@ fn pop_head_blocking<'mem>(
     stop: &AtomicBool,
 ) -> Option<DescriptorChain<'mem>> {
     loop {
-        match queue.pop(mem) {
-            Some(descriptor) => break Some(descriptor),
-            None => {
-                interrupt.signal_used_queue();
-                if stop.load(Ordering::Acquire) {
-                    break None;
-                }
-                thread::park();
-                log::trace!("tx unparked, queue len {}", queue.len(mem))
-            }
+        // Suppress guest notifications while draining to avoid a TOCTOU race:
+        // without suppression, the guest could add a descriptor and send a
+        // notification between queue.pop() returning None and thread::park(),
+        // and the notification would be lost if the TX thread is stopping.
+        queue.disable_notification(mem).unwrap_or_default();
+
+        // Pop one descriptor if available.
+        if let Some(descriptor) = queue.pop(mem) {
+            return Some(descriptor);
         }
+
+        // Re-enable notifications. enable_notification re-reads avail_idx with
+        // a SeqCst fence; if new entries arrived while we had notifications
+        // suppressed, it returns Ok(true) and we must drain again.  On Err,
+        // treat conservatively as if new entries may have arrived (notifications
+        // may not be fully armed), so also retry the drain.
+        if !matches!(queue.enable_notification(mem), Ok(false)) {
+            continue;
+        }
+
+        // Queue is confirmed empty and notifications are armed.  Check stop
+        // only after the drain+fence: if stop was set while the guest was
+        // writing its final bytes, we still flush them before returning None.
+        if stop.load(Ordering::Acquire) {
+            return None;
+        }
+
+        interrupt.signal_used_queue();
+        thread::park();
+        log::trace!("tx unparked, queue len {}", queue.len(mem));
     }
 }
 


mtjhrc:
Btw have you been able to reproduce this locally on your machine? For me it always works locally. But since it's a race condition that's certainly possible.  

Some prior observations about the flakeiness of `configure-vm-2cpu-1GiB`:
@mz-pdm was hitting an issue on his machine (with podman/krun) but it went away after host kernel upgrade. What's interesting is that this test was also failing in CI (randomly but less often than now) with the same error message (something failed in guest kernel with ENOMEM regarding a workqueue - I can't remember the details). I have stopped reading the msg and just rerunning the CI test manually - waiting for Ubuntu to update the kernel (under the impression it is the same issue).

jakecorrenti:
> Btw have you been able to reproduce this locally on your machine? For me it always works locally. But since it's a race condition that's certainly possible.
> 
> Some prior observations about the flakeiness of `configure-vm-2cpu-1GiB`: @mz-pdm was hitting an issue on his machine (with podman/krun) but it went away after host kernel upgrade. What's interesting is that this test was also failing in CI (randomly but less often than now) with the same error message (something failed in guest kernel with ENOMEM regarding a workqueue - I can't remember the details). I have stopped reading the msg and just rerunning the CI test manually - waiting for Ubuntu to update the kernel (under the impression it is the same issue).

I'm seeing it locally as well (this is a branch without the changes in this PR):
```bash
     Running unittests src/lib.rs (target/debug/deps/test_cases-701846697c599688)

running 1 test
test tests::all_testcases_have_unique_names ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

   Doc-tests test_cases

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

   Compiling test_cases v0.0.0 (/home/jcorrent/libkrun.bak/tests/test_cases)
   Compiling guest-agent v0.0.0 (/home/jcorrent/libkrun.bak/tests/guest-agent)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.56s
   Compiling krun-sys v1.11.1 (/home/jcorrent/libkrun.bak/krun-sys)
   Compiling test_cases v0.0.0 (/home/jcorrent/libkrun.bak/tests/test_cases)
   Compiling runner v0.0.0 (/home/jcorrent/libkrun.bak/tests/runner)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.28s
FreeBSD kernel: target/freebsd-kernel/freebsd-kern.bin
FreeBSD sysroot or init/init-freebsd not found; FreeBSD tests will be skipped.
(Run 'make' with BUILD_BSD_INIT=1 in the libkrun root to build FreeBSD assets.)
gvproxy (cached): /home/jcorrent/libkrun.bak/tests/target/gvproxy
[configure-vm-2cpu-1GiB] ... FAIL:
expected exactly "OK\n", got ""
(See test artifacts at: /tmp/libkrun-tests.LRmpVwoAhaIv)

FAIL - 0/1 passed
Error:
make: *** [Makefile:311: test] Error 1
```


EDIT: I got it to work locally, I missed a libkrunfw not found failure and ran this to get things working:
```bash
    RUST_LOG=trace \
    LIBKRUN_LIB_PATH="/home/jcorrent/libkrun/test-prefix/lib64/:/usr/local/lib64/" \
    PKG_CONFIG_PATH="/home/jcorrent/libkrun/test-prefix/lib64/pkgconfig/" \
    ./run.sh test --test-case "configure-vm-2cpu-1GiB"
```

lstocchi:
I see the tests succeeded but i wanted to share this if you want to try it out or for a future hint if something similar happens again. I'm afraid the park/unpark is weak against races.
When i see a similar pattern i always think about consumer/producer channels as they're great for dealing with a shutdown. When the other end is closed, the receiver knows it needs to do its work one last time (catching "OK\n" in this case).

I asked gemini for writing the code. I'm on windows atm so i didn't run the test but it compiles fine. 

Something like 
```
enum PortState {
    Active {
        ...
        tx_notifier: Option<Sender<()>>,
```

process function
We get inside the Err branch if the other side of the channel is dropped

```
fn pop_head_blocking<'mem>(
    queue: &mut Queue,
    mem: &'mem GuestMemoryMmap,
    interrupt: &InterruptTransport,
    tx_receiver: &Receiver<()>,
) -> Option<DescriptorChain<'mem>> {
    loop {
        match queue.pop(mem) {
            Some(descriptor) => break Some(descriptor),
            None => {
                interrupt.signal_used_queue();

                match tx_receiver.recv() {
                    Ok(_) => {
                        continue;
                    }
                    Err(_) => {
                        if let Some(descriptor) = queue.pop(mem) {
                            break Some(descriptor);
                        }
                        break None;
                    }
                }
            }
        }
    }
}
```

the shutdown 

```
pub fn shutdown(&mut self) {
        if let PortState::Active {
            stopfd,
            stop,
            tx_thread,
            rx_thread,
            tx_notifier,
        } = &mut self.state
        {
            stop.store(true, Ordering::Release);

            // This safely wakes up process_tx without needing thread::unpark()
            drop(tx_notifier.take());

            if let Some(tx_thread) = mem::take(tx_thread) {
                if let Err(e) = tx_thread.join() {
                    log::error!(
                        "Failed to flush tx for port {port_id}, thread panicked: {e:?}",
                        port_id = self.port_id
                    )
                }
            }
```

the notify 
```
pub fn notify_tx(&self) {
        if let PortState::Active {
            tx_notifier: Some(sender),
            ..
        } = &self.state
        {
            // Send a wake-up signal. If the thread is dead, this safely fails.
            let _ = sender.send(()); 
        }
    }
```

mtjhrc:
> I see the tests succeeded but i wanted to share this if you want to try it out or for a future hint if something similar happens again. I'm afraid the park/unpark is weak against races. When i see a similar pattern i always think about consumer/producer channels as they're great for dealing with a shutdown. When the other end is closed, the receiver knows it needs to do its work one last time (catching "OK\n" in this case).

Right, but I think the whole design of this code is not good in hindsight - we should be polling the eventfd for virtio queue directly in the worker thread (along with another eventfd for quitting). This would eliminate the eventfd readable -> thread::unpark() forwarding which seems useless.

mtjhrc:
@jakecorrenti 

I see you improved the enomem workaround, have you been able to reproduce the `workqueue: Failed to create a worker thread` locally? If yes what kernel (and distro I guess) did you use on the host? Or you just saw it fail with that in the CI?

jakecorrenti:
> @jakecorrenti
> 
> I see you improved the enomem workaround, have you been able to reproduce the `workqueue: Failed to create a worker thread` locally? If yes what kernel (and distro I guess) did you use on the host? Or you just saw it fail with that in the CI?

No, I haven't been able to get it to fail locally, I just saw it fail with that in the CI. I have the logs from the CI runs that failed if you'd like to see those.

This is still a draft as well, because I would like to _try_ and get a reproducer locally so that I'm not just throwing code against the wall to see what sticks.

jakecorrenti:
> @jakecorrenti
> 
> I see you improved the enomem workaround, have you been able to reproduce the `workqueue: Failed to create a worker thread` locally? If yes what kernel (and distro I guess) did you use on the host? Or you just saw it fail with that in the CI?

This is the log from the CI run **after** the TX race condition fix, that led to the ENOMEM workaround fixes:

```bash
--- snip --- 
[2026-06-22T17:42:00.769215Z TRACE krun_devices::virtio::console::process_tx] tx unparked, queue len 1
[2026-06-22T17:42:00.769217Z TRACE krun_devices::virtio::console::process_tx] Tx VolatileSlice { addr: 0x7f6f1e74aec8, size: 1, bitmap: (), mmap: None }, write_volatile 1 bytes
[2026-06-22T17:42:00.769219Z ERROR init_or_kernel] [    2.597935] workqueue: Failed to create a worker thread: -12
[2026-06-22T17:42:00.769221Z TRACE krun_devices::virtio::console::process_tx] Tx add used 1
[2026-06-22T17:42:00.769223Z DEBUG krun_devices::virtio::mmio[console]] interrupt: signal_used_queue
[2026-06-22T17:42:00.797693Z DEBUG krun_vmm] using vcpu exit code: 0
[2026-06-22T17:42:00.797713Z INFO  krun_vmm] Vmm is stopping.
--- snip ---
```

mtjhrc:
Do we plan to support older host kernels (assuming it is still that KVM issue)? If not maybe we can upgrade the CI to use ubuntu-26.04 with 7.0.0 kernel  (its technically a still preview though) and drop the ENOMEM workaround altogether? @slp 

I guess users of libkrun 1.x could use the env var if they have a buggy kernel? But for libkrun 2.0 we should drop this IMO.

slp:
> Do we plan to support older host kernels (assuming it is still that KVM issue)? If not maybe we can upgrade the CI to use ubuntu-26.04 with 7.0.0 kernel (its technically a still preview though) and drop the ENOMEM workaround altogether? @slp
> 
> I guess users of libkrun 1.x could use the env var if they have a buggy kernel? But for libkrun 2.0 we should drop this IMO.

Let's do this for both 2.0 and stable-1.19.x.

jakecorrenti:
@slp @mtjhrc just to clarify, what's the best course of action here? Am I closing this in favor of updating the CI or are we going to use this until ubuntu-26.04 is no longer preview?

slp:
> @slp @mtjhrc just to clarify, what's the best course of action here? Am I closing this in favor of updating the CI or are we going to use this until ubuntu-26.04 is no longer preview?

Let's keep the first commit.

slp:
Please rebase to fix CI.

slp (approved):
LGTM, thanks @jakecorrenti !

PR #740, opened 2026-06-23
virtio/gpu: complete macOS blob mapping (populate map_ptr, allow export-less blobs)

jovemexausto:
## Summary
- Populate `RutabagaResource::map_ptr` for gfxstream blob resources, which the macOS `resource_map_blob` path requires to install a hypervisor mapping (SHM exports via `mmap`; host-visible blobs with no exportable descriptor via gfxstream's own `stream_renderer_resource_map`).
- Don't require a successful blob export to install the mapping: the export handle was only inspected as a validity gate and is never used by the mapping itself, so host-visible `VkDeviceMemory` (which MoltenVK can't export) can still be mapped via its `map_ptr`.
- Accept the `SHM` handle type in addition to `APPLE`, and log a reason on every failure path instead of a bare `ERR_UNSPEC`.

## Why
On macOS, gfxstream returns SHM-backed and host-visible blob exports. Upstream `resource_map_blob` now maps via `rutabaga.map_ptr()`, but the gfxstream component never set it, so every `RESOURCE_MAP_BLOB` failed with "no map ptr available". This completes the path so blob-backed GPU resources actually map during initialization. Companion to google/gfxstream#151 (the host-renderer side); together they build and run gfxstream as the virtio-gpu backend on Apple Silicon.

## Validation
- `krun-rutabaga-gfx --features gfxstream` builds; `krun-devices --features gpu` type-checks on macOS/AArch64.
- This is the distilled, upstream form of the mapping code we run to reach Android `boot_completed` and drive four concurrent virtual displays under load on macOS/HVF.

### DIFF
--- a/src/devices/src/virtio/gpu/virtio_gpu.rs
+++ b/src/devices/src/virtio/gpu/virtio_gpu.rs
@@ -905,11 +905,6 @@ impl VirtioGpu {
         let map_ptr = self.rutabaga.map_ptr(resource_id).map_err(|_| ErrUnspec)?;
 
         if let Ok(export) = self.rutabaga.export_blob(resource_id) {
-            debug!(
-                "resource_map_blob: export handle_type={} map_ptr={:x}",
-                export.handle_type, map_ptr
-            );
-
             if export.handle_type == RUTABAGA_MEM_HANDLE_TYPE_APPLE
                 || export.handle_type == RUTABAGA_MEM_HANDLE_TYPE_SHM
             {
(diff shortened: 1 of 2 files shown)


jovemexausto:
Companion change for the gfxstream fix [here](https://github.com/google/gfxstream/pull/151). Both patches are related and were validated together in the macOS/HVF Android integration path.

slp:
Thanks for the PR. In the next couple weeks, we're going to replace the interface between libkrun and virglrenderer to stop exposing `map_ptr` and rely on `virgl_renderer_resource_map`, hiding the implementation details in virglrenderer itself. Then we'll drop the vendored rutabaga_gfx, using the one published in crates.io instead.

This means we won't be able to merge this PR, at least in its current form. If there's still a functional issue, we can work on it on top of the new interface.

PR #741, opened 2026-06-23
cpuid: advertise the TSC-deadline timer on AMD hosts

ZauberNerd:
On AMD hosts the guest doesn't see the LAPIC TSC-deadline timer (CPUID.01H:ECX[24]) and calibrates the APIC timer against the PIT at every boot, ~200ms in calibrate_APIC_clock(). AmdCpuidTransformer maps leaf 0x1 to common::update_feature_info_entry, which omitted the bit; only the Intel transformer set it.

KVM emulates the TSC-deadline timer for the in-kernel LAPIC regardless of host support, so move the bit into the common path. AMD guests skip calibration; the Intel transformer already calls common, so its own (now redundant) set is removed and Intel output is unchanged.

In-guest boot drops ~200ms on AMD Ryzen 7 5700G / KVM (guest dmesg: `Calibrating APIC timer ...` → `TSC deadline timer available`).

Scope: x86_64/KVM only — filter_cpuid runs solely on the Linux backend; macOS/HVF and Windows/WHP have separate CPUID paths. `cargo test -p krun-cpuid` passes.

Assisted-by: Claude Code / claude-opus-4.8

### DIFF
--- a/src/cpuid/src/transformer/common.rs
+++ b/src/cpuid/src/transformer/common.rs
@@ -38,6 +38,8 @@ pub fn update_feature_info_entry(
     // X86 hypervisor feature
     entry.ecx.write_bit(ecx::HYPERVISOR_BITINDEX, true);
 
+    entry.ecx.write_bit(ecx::TSC_DEADLINE_TIMER_BITINDEX, true);
+
     entry
         .ebx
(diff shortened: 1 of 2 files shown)


slp:
I don't have a system to check the improvement, but the change looks good to me. Thanks @ZauberNerd !

slp (approved):

PR #743, opened 2026-06-23
 macOS: fix musl cross-linking and remove stale Linux sysroot

mtjhrc:
On macOS, building the init binary for aarch64-unknown-linux-musl failed because no cross-linker was configured — the Apple system linker doesn't understand GNU ld flags emitted by rustc for musl targets.

This PR configures the env vars in the Makefile instead and uses it for both guest-agent in tests and init.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -33,7 +33,7 @@ make install
 make PREFIX=$HOME/.local install
 ```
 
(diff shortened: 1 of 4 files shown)


slp:
Nice cleanup, thanks. Please rebase to fix CI.

slp (approved):
LGTM, thanks @mtjhrc !

PR #744, opened 2026-06-23
tests/virtiofs-misc: add UID/GID mapping subtests

mtjhrc:
Add 2 subtests to virtiofs-misc that verify UID/GID ownership as seen from inside the guest:

- uid_gid_mapping_preexisting: a file created on the host before VM boot must appear as root-owned in the guest. This acts as a regression test for issue described and fixed in PR #734
- uid_gid_mapping_chown: chown to a non-root UID/GID is reflected in stat, and chown back to root works.

Assisted-by: OpenCode:claude-opus-4.6

Requires #743 to fix the `make test` linking failure on main branch, backport to 1.19.x doesn't require it because it uses the older `init.c`.

### DIFF
--- a/tests/runner/src/main.rs
+++ b/tests/runner/src/main.rs
@@ -164,17 +164,19 @@ fn run_single_test(
             eprintln!("WARNING: buildah not available, running without namespace isolation.");
             eprintln!("Tests may fail if the required network ports are already in use.");
         }
-        Command::new(&executable)
-            .arg("start-vm")
+        let mut cmd = Command::new(&executable);
+        cmd.arg("start-vm")
             .arg("--test-case")
             .arg(test_case.name)
             .arg("--tmp-dir")
             .arg(&test_dir)
             .stdin(Stdio::piped())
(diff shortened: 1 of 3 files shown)


slp:
Looks like we need to skip this test if `buildah` is not present in the system.

Tested outside of the runner, with `KRUN_NO_UNSHARE=1`:

```
[virtiofs-misc] ....................... WARNING: buildah not available, running without namespace isolation.
Tests may fail if the required network ports are already in use.
FAIL:
expected exactly "OK\n", got "FAILED: uid_gid_mapping_preexisting, uid_gid_mapping_chown\n"
```

mtjhrc:
> Looks like we need to skip this test if buildah is not present in the system.

Done.

Btw. thinking about this - this will probably need a refactor later as we add more fs subtests here to clean up the organization (each subtest might want to have a host and guest part too). We support skipping whole testcases already (each testcase is a virtual machine), but subtests are a new concept here.

slp (approved):
Thanks @mtjhrc !

PR #745, opened 2026-06-23
ci: switch to ubuntu-26.04 runners

slp:
Let's switch to newer kernels, hoping to avoid the buggy kernel in 24.04 that leads to https://github.com/libkrun/libkrun/issues/314


jakecorrenti (approved):
lgtm, thanks!

mtjhrc (approved):
LGTM thanks!

PR #747, opened 2026-06-23
ci: include runner image in cache keys

mtjhrc:
The cargo cache key only used runner.os and runner.arch, which don't distinguish between Ubuntu versions (e.g. 24.04 vs 26.04). When switching runner images, stale cached build artifacts (like bindgen's libclang linker flags) would reference libraries from the old image that no longer exist on the new one, causing link failures like 'cannot find -lclang-18.1.3'.

Add env.ImageOS to the cargo cache key so each runner image gets its own cache namespace.

Assisted-by: OpenCode:claude-opus-4.6

### DIFF
--- a/.github/actions/setup-build-env/action.yml
+++ b/.github/actions/setup-build-env/action.yml
@@ -20,7 +20,7 @@ runs:
           ~/.cargo/registry
           ~/.cargo/git
           target
-        key: ${{ runner.os }}-${{ runner.arch }}-cargo-${{ hashFiles('**/Cargo.lock') }}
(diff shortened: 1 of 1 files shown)


jakecorrenti:
When I do `gh cache list --repo libkrun/libkrun` the image cache still just shows `Linux-X64-cargo-<hash>`. I could be verifying this wrong, though.

Do i have to wait for this to merge to see anything show up?

mtjhrc:
Well the CI works now and I don't want to debug if this solves anything. We can revisit this in 2028 (Ubuntu 28.04) :stuck_out_tongue_winking_eye:

PR #758, opened 2026-06-26
virtio/fs: implement simplified permission semantics [stable test]

slp:
So far, we were always trying to be as close as possible to the common
    Linux file system semantics. But, there are certain scenarios (namely,
    acting as a container volume) in which simplified semantics may be
    preferable.

    To address that need, let's introduce configurable permission semantics,
    with two initial modes: LinuxComplete and LinuxSimplified. The first is
    exactly the same we had until 1.18.0. The second, is a simplified
    version with this differences:

    - Idmaps are not supported.
    - Ownership bits are ignored, always returning the uid/gid from the
      process requesting the operation within the guest (obtained from
      `Context`).
    - Permissions bits are stored in the host, no as extended attributes.

### DIFF
--- a/src/devices/src/virtio/fs/device.rs
+++ b/src/devices/src/virtio/fs/device.rs
@@ -21,6 +21,7 @@ use super::virtual_entry::VirtualDirEntry;
 use super::worker::FsWorker;
 use super::ExportTable;
 use super::{defs, defs::uapi};
+use crate::virtio::passthrough::PermissionSemantics;
 use crate::virtio::InterruptTransport;
 
 #[derive(Copy, Clone)]
@@ -46,6 +47,7 @@ pub struct Fs {
     acked_features: u64,
     device_state: DeviceState,
     config: VirtioFsConfig,
+    allow_idmap: bool,
     shm_region: Option<VirtioShmRegion>,
     passthrough_cfg: Option<passthrough::Config>,
     read_only: bool,
@@ -60,6 +62,7 @@ pub struct Fs {
 impl Fs {
     pub fn new(
         fs_id: String,
+        semantics: PermissionSemantics,
         shared_dir: Option<String>,
         exit_code: Arc<AtomicI32>,
         read_only: bool,
@@ -74,14 +77,18 @@ impl Fs {
 
         let fs_cfg = shared_dir.map(|root_dir| passthrough::Config {
             root_dir,
+            semantics,
             ..Default::default()
         });
+
+        let allow_idmap = matches!(semantics, PermissionSemantics::LinuxComplete);
 
         Ok(Fs {
             avail_features,
             acked_features: 0,
             device_state: DeviceState::Inactive,
             config,
+            allow_idmap,
             shm_region: None,
             passthrough_cfg: fs_cfg,
             read_only,
@@ -194,6 +201,7 @@ impl VirtioDevice for Fs {
             queue_evts,
             interrupt.clone(),
             mem.clone(),
+            self.allow_idmap,
             self.shm_region.clone(),
             self.passthrough_cfg.clone(),
             self.read_only,
--- a/src/devices/src/virtio/fs/device.rs
+++ b/src/devices/src/virtio/fs/device.rs
@@ -5,6 +5,7 @@ use std::io::Write;
 use std::sync::atomic::{AtomicI32, AtomicU64, Ordering};
 use std::sync::Arc;
 use std::thread::JoinHandle;
+use std::time::Duration;
 
 use utils::eventfd::{EventFd, EFD_NONBLOCK};
 #[cfg(target_os = "macos")]
@@ -75,9 +76,18 @@ impl Fs {
         config.tag[..tag.len()].copy_from_slice(tag.as_slice());
         config.num_request_queues = 1;
 
+        let attr_timeout = if matches!(semantics, PermissionSemantics::LinuxSimplified) {
+            // As uid/gid are context-dependent, attributes can't be cached.
+            Duration::from_secs(0)
+        } else {
+            // The value defined as default in virtio-fs.
+            Duration::from_secs(5)
+        };
+
         let fs_cfg = shared_dir.map(|root_dir| passthrough::Config {
             root_dir,
             semantics,
+            attr_timeout,
             ..Default::default()
         });
 
--- a/src/devices/src/virtio/fs/macos/passthrough.rs
+++ b/src/devices/src/virtio/fs/macos/passthrough.rs
@@ -161,16 +161,6 @@ fn einval() -> io::Error {
     linux_error(io::Error::from_raw_os_error(libc::EINVAL))
 }
 
-fn host_euid() -> u32 {
-    static EUID: std::sync::OnceLock<u32> = std::sync::OnceLock::new();
-    *EUID.get_or_init(|| unsafe { libc::geteuid() })
-}
-
-fn host_egid() -> u32 {
-    static EGID: std::sync::OnceLock<u32> = std::sync::OnceLock::new();
-    *EGID.get_or_init(|| unsafe { libc::getegid() })
-}
-
 fn item_to_value(item: &[u8], radix: u32) -> Option<u32> {
     match std::str::from_utf8(item) {
         Ok(val) => match u32::from_str_radix(val, radix) {
@@ -365,21 +355,11 @@ fn stat_common(
     host: bool,
 ) -> io::Result<bindings::stat64> {
     if !host {
-        match uid {
-            Some(uid) => st.st_uid = uid,
-            None => {
-                if st.st_uid == host_euid() {
-                    st.st_uid = 0;
-                }
-            }
+        if let Some(uid) = uid {
+            st.st_uid = uid;
         }
-        match gid {
-            Some(gid) => st.st_gid = gid,
-            None => {
-                if st.st_gid == host_egid() {
-                    st.st_gid = 0;
-                }
-            }
+        if let Some(gid) = gid {
+            st.st_gid = gid;
         }
         if let Some(mode) = mode {
             if mode as u16 & libc::S_IFMT == 0 {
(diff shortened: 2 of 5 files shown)


mtjhrc:
Interesting! Could you share more details for use cases/bugs this is trying to fix?
Is the problem just the fact that we don't want to store (I suppose lazily add if missing?) the `xattr` to mounted container volumes? Or the problem is also that permission semantics don't end up mapped correctly?

mtjhrc:
Could semantics like this work instead?

If the file doesn't have `user.containers.override_stat` `xattr` assume it is always owned by UID+GID `0`, read the real permission bits. Never report the real UID/GID from host (they are bogus for the guest no?) If the user in container `chown()`s the file write the  `user.containers.override_stat` to store the new UID/GID. But for the actual permission bits always use the real host ones.

slp:
> Interesting! Could you share more details for use cases/bugs this is trying to fix?
> Is the problem just the fact that we don't want to store (I suppose lazily add if missing?) the `xattr` to mounted container volumes? Or the problem is also that permission semantics don't end up mapped correctly?

We basically need to mimic the behavior from other container tools on macOS, which is in some aspects significantly different from what you would expect from a Linux container, but has set the user expectations on that platform.

The most significant aspect is ignoring the ownership bits and always return the uid/gid of the process requesting the operation in the guest.

> Could semantics like this work instead?
> 
> If the file doesn't have `user.containers.override_stat` `xattr` assume it is always owned by UID+GID `0`, read the real permission bits. Never report the real UID/GID from host (they are bogus for the guest no?) If the user in container `chown()`s the file write the `user.containers.override_stat` to store the new UID/GID. But for the actual permission bits always use the real host ones.

No. We need to ignore uid/gid changes from the guest, and always go with the ones in `Context`. This completely breaks the common Linux filesystem semantics, but it's what users have come to expect when sharing volumes on macOS (and I'm pretty sure this will also be the case on Windows).

On the good side, this has allowed us to avoid a significant amount of stat/xattr syscalls, so the simplified mode should also perform better to some degree.

mtjhrc:
> We basically need to mimic the behavior from other container tools on macOS, which is in some aspects significantly different from what you would expect from a Linux container, but has set the user expectations on that platform.

> This completely breaks the common Linux filesystem semantics, but it's what users have come to expect 

Oh, I wasn't aware of that convention. Well in this context this PR makes much more sense now!

benoitf:
hello, I did some tests

it works on some command line but it's not reporting the same results than podman/applehv or docker for these commands

with docker:
```
$ docker run -it -v $HOME/tmp/data:/etcd quay.io/coreos/etcd:v3.5.0 sh                                                                                                                                                                                                                                   # ls -la /
total 68
drwxr-xr-x   1 root root 4096 Jun 26 11:36 .
drwxr-xr-x   1 root root 4096 Jun 26 11:36 ..
-rwxr-xr-x   1 root root    0 Jun 26 11:36 .dockerenv
drwxr-xr-x   2 root root 4096 Jan 27  2021 bin
drwxr-xr-x   2 root root 4096 Nov 22  2020 boot
drwxr-xr-x   5 root root  360 Jun 26 11:36 dev
drwxr-xr-x   1 root root 4096 Jun 26 11:36 etc
drwxr-xr-x   2 root root   64 Jun 25 09:00 etcd   <--- owner is root.root
```

with podman/applehv:
```
$ podman run -it -v $HOME/tmp/data:/etcd quay.io/coreos/etcd:v3.5.0 sh                                                                                                                                                                                                                                    # ls -la /
total 8
dr-xr-xr-x.   1 root root   29 Jun 26 11:45 .
dr-xr-xr-x.   1 root root   29 Jun 26 11:45 ..
-rwxr-xr-x.   1 root root    0 Jan 27  2021 .dockerenv
drwxr-xr-x.   2 root root 4096 Jan 27  2021 bin
drwxr-xr-x.   2 root root    6 Nov 22  2020 boot
drwxr-xr-x.   5 root root  360 Jun 26 11:45 dev
drwxr-xr-x.  25 root root 4096 Jan 27  2021 etc
drwxr-xr-x.   2 root root   64 Jun 25 09:00 etcd <--- root.root
```


with this PR I have
```
$ podman run -it -v $HOME/tmp/data:/etcd quay.io/coreos/etcd:v3.5.0 sh                                                                                                                                                                                                                                 # ls -la /
total 8
dr-xr-xr-x.   1 root root   29 Jun 26 11:38 .
dr-xr-xr-x.   1 root root   29 Jun 26 11:38 ..
-rwxr-xr-x.   1 root root    0 Jan 27  2021 .dockerenv
drwxr-xr-x.   2 root root 4096 Jan 27  2021 bin
drwxr-xr-x.   2 root root    6 Nov 22  2020 boot
drwxr-xr-x.   5 root root  360 Jun 26 11:38 dev
drwxr-xr-x.  25 root root 4096 Jan 27  2021 etc
drwxr-xr-x.   2  998  998   64 Jun 25 09:00 etcd   <--- owner is 998.998
```


and using OpenShell:

```bash
openshell sandbox create --from base --driver-config-json '{"podman":{"mounts":[{"type":"bind","source":"/Users/benoitf/git/tmp/hello","target":"/sandbox/work"}]}}'
```

when I go to the /sandbox folder, `work` is owned by `root:root` while with docker or podman/docker it is owned by `sandbox:sandbox`
But it might be related to the sandboxing process

with podman/libkrun/this patch
```
$ ls -la
total 8
drwxr-xr-x. 7 sandbox sandbox 103 Jun 26 11:29 .
dr-xr-xr-x. 1 root    root     61 Jun 26 11:29 ..
drwxr-xr-x. 3 sandbox sandbox  20 May 29 16:24 .agents
-rw-r--r--. 1 sandbox sandbox 174 May 29 16:24 .bashrc
drwxr-xr-x. 3 sandbox sandbox  20 May 29 16:24 .claude
-rw-r--r--. 1 sandbox sandbox  32 May 29 16:24 .profile
drwxr-xr-x. 3 sandbox sandbox  20 May 29 16:24 .uv
drwxr-xr-x. 4 sandbox sandbox 110 May 29 16:24 .venv
drwxr-xr-x. 7 root    root    224 Jun 25 11:53 work  <-- it's owned by root so I don't have access because my user-id of the sandboxed user is `sandbox`
```


with podman/applehv:

```
$ openshell sandbox create --from base --driver-config-json '{"podman":{"mounts":[{"type":"bind","source":"/Users/benoitf/git/tmp/hello","target":"/sandbox/work"}]}}'                                                               
                                                                                                                                                                                                                                                                                                                                    Created sandbox: diverting-urial
sandbox@sandbox-diverting-urial:~$ ls -la
total 8
drwxr-xr-x. 7 sandbox sandbox 103 Jun 26 11:47 .
dr-xr-xr-x. 1 root    root     61 Jun 26 11:47 ..
drwxr-xr-x. 3 sandbox sandbox  20 May 29 16:24 .agents
-rw-r--r--. 1 sandbox sandbox 174 May 29 16:24 .bashrc
drwxr-xr-x. 3 sandbox sandbox  20 May 29 16:24 .claude
-rw-r--r--. 1 sandbox sandbox  32 May 29 16:24 .profile
drwxr-xr-x. 3 sandbox sandbox  20 May 29 16:24 .uv
drwxr-xr-x. 4 sandbox sandbox 110 May 29 16:24 .venv
drwxr-xr-x. 7 sandbox sandbox 224 Jun 25 11:53 work  <--- owner is sandbox.sandbox and the user is sandbox
```


when I'm connecting to the OpenShell container using podman command:

```
podman exec -ti 81f759f353ad bash                                                                                                                                                                                                                                                                                 

root@sandbox-diverting-urial:/sandbox# ls -la /sandbox/
total 12
drwxr-xr-x. 7 sandbox sandbox 124 Jun 26 11:50 .
dr-xr-xr-x. 1 root    root     61 Jun 26 11:47 ..
drwxr-xr-x. 3 sandbox sandbox  20 May 29 16:24 .agents
-rw-------. 1 sandbox sandbox   7 Jun 26 11:50 .bash_history
-rw-r--r--. 1 sandbox sandbox 174 May 29 16:24 .bashrc
drwxr-xr-x. 3 sandbox sandbox  20 May 29 16:24 .claude
-rw-r--r--. 1 sandbox sandbox  32 May 29 16:24 .profile
drwxr-xr-x. 3 sandbox sandbox  20 May 29 16:24 .uv
drwxr-xr-x. 4 sandbox sandbox 110 May 29 16:24 .venv
drwxr-xr-x. 7 root    root    224 Jun 25 11:53 work  <----------- owned by root.root

# then I change my user
root@sandbox-diverting-urial:/sandbox# su - sandbox
sandbox@sandbox-diverting-urial:~$ ls -la
total 12
drwxr-xr-x. 7 sandbox sandbox 124 Jun 26 11:50 .
dr-xr-xr-x. 1 root    root     61 Jun 26 11:47 ..
drwxr-xr-x. 3 sandbox sandbox  20 May 29 16:24 .agents
-rw-------. 1 sandbox sandbox   7 Jun 26 11:50 .bash_history
-rw-r--r--. 1 sandbox sandbox 174 May 29 16:24 .bashrc
drwxr-xr-x. 3 sandbox sandbox  20 May 29 16:24 .claude
-rw-r--r--. 1 sandbox sandbox  32 May 29 16:24 .profile
drwxr-xr-x. 3 sandbox sandbox  20 May 29 16:24 .uv
drwxr-xr-x. 4 sandbox sandbox 110 May 29 16:24 .venv
drwxr-xr-x. 7 sandbox sandbox 224 Jun 25 11:53 work <------------- owned by sandbox.sandbox


```

benoitf:
I had probably some missing files

now it works as expected on my end

I have the same output on macOS using podman/applehv, podman/libkrun or docker

thanks

benoitf:
the difference is on the mount point which is always root.root but files have correct ownership

benoitf:
with latest changes it's working like podman/applehv or docker 🚀

PR #761, opened 2026-06-26
AGENTS.md: ask agents to add fewer comments

slp:
As a reviewer, I've been fooled a couple times by comments in code that provided a convincing explanation, but were ultimately wrong.

Let's ask agents to use the same policy we use for humans: code must be self-explanatory and the main source of reference. Only counter-intuitive code should be commented.

This should help reviewers to be less distracted by the comments and focus on the actual code produced by the agents.


jakecorrenti (approved):
LGTM, nice addition

PR #530, opened 2026-02-04
Rewrite virtio-net IO to be more efficient 

mtjhrc:
This PR 
- introduces new abstractions for dealing with virtio queues (`RxQueueProducer`, `TxQueueConsumer`) in an efficient manner 
- rewrite all 3 virtio-net device backends to use the abstractions (copy straight from guest memory into the backend, no intermediary buffer)
- new integration tests for virtio-net over tap, passt, gvproxy   

TODO:
- [ ] more code cleanup
- [ ] at least some basic benchmark data (e.g. iperf)
- [ ] more manual testing besides the simple integration smoke tests 

Adresses: #385, #405
supersedes: https://github.com/containers/libkrun/pull/493

PR #762, opened 2026-06-26
vmm,devices: VM snapshot/restore (HVF)

vadika:
**VM snapshot/restore (HVF-first, macOS/aarch64). Tracking #748.** Supersedes M0 #753. Builds on the pause primitive from #767.

Boot a VM, snapshot it to a directory, restore onto a fresh VM, resume — the restored guest picks up mid-workload, runs to completion, and powers off.

## Capture

`krun_snapshot_request` sends `VmCtl::Snapshot(path, flags)` down the same channel `krun_vm_pause` uses. The event loop then: freezes the vCPUs (`Vmm::pause`), quiesces every device, saves the GIC, serializes device + vCPU state, dumps guest RAM, and exits.

Saved: vCPU regs (incl. pointer-auth keys, vtimer + `mach_absolute_time` anchor for CNTVCT continuity), the in-kernel GIC dist/redist blob + per-vCPU ICC, guest RAM, and device state (PL031 RTC + virtio devices via their MMIO transport, including the virtio-mmio ISR). FD-free — transport/queue state only, never host fds.

## Device quiesce

Freezing the vCPUs stops the guest but not a device's worker threads, which own the live queue indices. `VirtioDevice::pause()` stops the workers and hands the queues back; `save()` reads the exact indices from them. `resume(queues)` restarts a device, replacing the old restart hook. Per @mtjhrc's review: how a device stops its workers is its own business, and `pause` must not drain (a worker blocked on a stalled host fd would deadlock) — it stops at a descriptor boundary. Implemented for block, virtio-fs, console, rng and balloon; devices that can't quiesce reject the snapshot rather than being captured mid-flight.

## Restore

`krun_restore(ctx, path, 0)` takes an already-configured context (config-validate — no setter allowlist), validates it against the manifest (arch, memory, vCPU count, and the ordered device topology), hydrates RAM/vCPU/GIC/device state, then enters the guest like `krun_start_enter`.

- Untrusted-snapshot hardening: sysreg/ICC register ids are validated against the capture allowlists; queue rings are validated before use; a malformed vmstate returns an error instead of panicking.
- Device workers are restarted only after the vCPUs report the GIC blob is restored, so an IRQ a worker raises can't be overwritten. Queue eventfds are re-kicked so in-flight descriptors complete.
- Multi-vCPU and nested-virtualization restore are rejected (not yet supported), rather than silently corrupting.

The C API is two functions: `krun_snapshot_request` and `krun_restore`. Both provisional; the internals (`Snapshottable` in the devices crate, arch-keyed `Aarch64VcpuState`, the `vmstate` format) are API-agnostic and survive the typed-API reshape in #718. See `examples/snapshot.c` and the README for usage + limitations.

## Testing (macOS/HVF)

- `snapshot-create` / `snapshot-resume`: capture, restore onto a fresh VM, assert the guest resumes mid-loop and powers off.
- `snapshot-fs-capture`: snapshot while the guest hammers virtio-fs, so the fs worker is quiesced mid-I/O.
- `vm-pause` unregressed. Unit round-trips for `vmstate`, vCPU/RTC/queue, `Manifest::validate`. `clippy -D warnings` (base + `net,blk`) + `fmt` clean.

## Deferred

Per-device internal state beyond queues (e.g. virtio-fs open fids, so a guest that did fs I/O can't resume yet) and a partially-consumed descriptor. `KRUN_SNAPSHOT_RESUME` (capture-and-continue) returns `-ENOTSUP`. Multi-vCPU restore. `MAP_PRIVATE` of `memory.img` (restore copies today). KVM/Linux backend.

### DIFF
--- a/include/libkrun.h
+++ b/include/libkrun.h
@@ -1365,6 +1365,32 @@ int32_t krun_vm_resume(uint32_t ctx_id);
 int32_t krun_snapshot_request(uint32_t ctx_id, const char *snapshot_path,
                               uint32_t flags);
 
+/**
+ * Restore a VM from a snapshot and run it. "ctx_id" must be an already-
+ * configured context whose shape (architecture, memory size, vCPU count)
+ * matches the snapshot's "manifest.json"; krun_restore() validates that, then
+ * hydrates guest RAM, device, and vCPU state onto it and enters the guest.
+ * Currently macOS/HVF only; returns -ENOTSUP elsewhere and on TEE builds.
+ *
+ * Devices are reconfigured fresh by the caller (with new host fds); the saved
+ * transport state binds onto them, so open file descriptors need not survive.
+ *
+ * Arguments:
+ *  "ctx_id"        - the configuration context ID.
+ *  "snapshot_path" - the snapshot directory written by krun_snapshot_request().
+ *  "flags"         - reserved, must be 0.
+ *
+ * Notes:
+ *  Like krun_start_enter(), on success this assumes control of the process and
+ *  does not return until the restored guest exits.
+ *
+ * Returns:
+ *  -EINVAL  - bad path, non-zero flags, or a config that doesn't match the
+ *             snapshot manifest.
+ *  -ENOTSUP - restore is unsupported on this platform/build.
+ */
+int32_t krun_restore(uint32_t ctx_id, const char *snapshot_path, uint32_t flags);
+
 #ifdef __cplusplus
 }
 #endif
--- a/include/libkrun.h
+++ b/include/libkrun.h
@@ -1336,6 +1336,35 @@ int32_t krun_vm_pause(uint32_t ctx_id);
  */
 int32_t krun_vm_resume(uint32_t ctx_id);
 
+/* Snapshot flags. */
+#define KRUN_SNAPSHOT_EXIT 0   /* capture, then exit */
+#define KRUN_SNAPSHOT_RESUME 1 /* capture, then resume (unimplemented) */
+
+/**
+ * Thread-safe, out-of-band snapshot trigger for a VM already running under
+ * krun_start_enter() on another thread. The VM's event loop freezes the vCPUs,
+ * serializes the VM into "snapshot_path", and exits the process.
+ *
+ * The snapshot is written as a directory containing "manifest.json", "vmstate",
+ * and "memory.img".
+ *
+ * Currently macOS/HVF only; returns -ENOTSUP elsewhere and on TEE builds.
+ *
+ * Arguments:
+ *  "ctx_id"        - the configuration context ID of the running VM.
+ *  "snapshot_path" - destination directory for the snapshot.
+ *  "flags"         - KRUN_SNAPSHOT_EXIT.
+ *
+ * Returns:
+ *  0        - the snapshot request was signalled.
+ *  -EINVAL  - bad path or unsupported flags.
+ *  -ENOENT  - no running VM is registered for this ctx_id.
+ *  -EIO     - failed to signal the VM.
+ *  -ENOTSUP - snapshots are unsupported on this platform/build.
+ */
+int32_t krun_snapshot_request(uint32_t ctx_id, const char *snapshot_path,
+                              uint32_t flags);
+
 #ifdef __cplusplus
 }
 #endif
--- a/src/devices/src/virtio/mmio.rs
+++ b/src/devices/src/virtio/mmio.rs
@@ -17,8 +17,9 @@ use super::device_status;
 use super::*;
 use crate::bus::BusDevice;
 use crate::legacy::IrqChip;
+use crate::snapshot::{DeviceState, Snapshottable};
 use utils::{byte_order, eventfd::EventFd};
-use vm_memory::{GuestAddress, GuestMemoryMmap};
+use vm_memory::{Bytes, GuestAddress, GuestMemoryMmap};
 
 //TODO crosvm uses 0 here, but IIRC virtio specified some other vendor id that should be used
 const VENDOR_ID: u32 = 0;
@@ -78,6 +79,12 @@ pub struct MmioTransport {
     queue_config: Vec<QueueConfig>,
     shm_region_select: u32,
     interrupt: InterruptTransport,
+    // Per-instance registration id, used as the snapshot key (id()).
+    name: String,
+    // Queue addresses/sizes captured at activation. The live queues are moved
+    // into the device (and then its worker threads), so this is the only place
+    // the transport can recover their config for a snapshot.
+    activated_queues: Vec<QueueSnapshot>,
 }
 
 struct InterruptTransportInner {
@@ -168,12 +175,17 @@ impl MmioTransport {
         mem: GuestMemoryMmap,
         intc: IrqChip,
         device: Arc<Mutex<dyn VirtioDevice>>,
+        id: String,
     ) -> Result<MmioTransport, CreateMmioTransportError> {
         let locked = device
             .try_lock()
             .expect("Mutex of VirtioDevice should not be locked when calling MmioTransport::new");
 
-        let debug_log_target = format!("{}[{}]", module_path!(), locked.device_name());
+        // The per-instance registration id (e.g. the block drive id, or "fs0"),
+        // not the device kind — snapshot keys device state by it, so it must be
+        // unique across devices of the same kind.
+        let name = id;
+        let debug_log_target = format!("{}[{}]", module_path!(), name);
         let queue_config: Vec<QueueConfig> = locked.queue_config().to_vec();
         drop(locked);
 
@@ -193,6 +205,8 @@ impl MmioTransport {
             queue_evts,
             queue_config,
             shm_region_select: 0,
+            name,
+            activated_queues: Vec::new(),
         })
     }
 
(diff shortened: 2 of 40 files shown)


src/devices/src/virtio/mmio.rs
```diff
@@ -527,6 +541,94 @@ impl BusDevice for MmioTransport {
...
+        &self.name
+    }
+
+    fn save(&self) -> Result<DeviceState, crate::Error> {
+        // The live queues' running indices are owned by the device's worker
+        // threads, but the vCPUs are frozen and the workers quiesced at snapshot
+        // time, so every consumed buffer has been returned: next_avail ==
+        // next_used == the guest's used-ring index. Recover the addresses from
+        // the activation-time config and the index from guest memory.
+        let queues = self
```
mtjhrc: This doesn't seem correct, "the workers quiesced at snapshot time" by what mechanism? Looking at the code they are still running processing user request(s) it's just the guest's vcpus are frozen. 

I would really suggest redesigning this. They way I imagine this working is implementing new methods for `VirtioDevice` trait `fn pause() -> Vec<DeviceQueue>` which explicitly transfers ownership of the queues to the transport layer (mmio) and `fn resume(mem: GuestMemoryMmap, queues: Vec<DeviceQueue>)` which resumes (called instead of activate). I don't have the exact model in my head but this is roughly the direction I think this should go.

src/devices/src/virtio/mmio.rs
```diff
@@ -173,7 +180,8 @@ impl MmioTransport {
             .try_lock()
             .expect("Mutex of VirtioDevice should not be locked when calling MmioTransport::new");
 
-        let debug_log_target = format!("{}[{}]", module_path!(), locked.device_name());
+        let name = locked.device_name().to_string();
```
mtjhrc: The `device_name()` isn't unique currently. You might have multiple devices with the same name. (e.g. multiple virtio-fs devices would all be just "fs" currently).  So this name isn't really suitable as a serialization key.

mtjhrc (changes requested):

include/libkrun.h
```diff
@@ -1300,6 +1300,51 @@ int32_t krun_set_root_disk_remount(uint32_t ctx_id, const char *device, const ch
...
+ *                    krun_snapshot_create().
+ *  "flags"         - KRUN_SNAPSHOT_EXIT.
+ *
+ * Returns:
+ *  Zero on success or a negative error number on failure. Notably -ENOENT if no
+ *  VM is registered for "ctx_id", or -ENOTSUP on unsupported platforms/builds.
+ */
+int32_t krun_snapshot_request(uint32_t ctx_id, const char *snapshot_path,
+                              uint32_t flags);
+
```
mtjhrc, line 1393: So I might have misunderstood or missed this from the previous discussion. I don't see a point for krun_snapshot_create (which starts the VM) to exist as a separate function.

As I already stated https://github.com/libkrun/libkrun/pull/753#issuecomment-4811088677, I'm not convinced we should provide a built-in guest trigger for the snapshot — the embedder can trivially implement that themselves e.g. via vsock, virtio console ports, or via other future communication priimitves. So the only way to actually trigger the snapshot is `krun_snapshot_request`. But since `krun_snapshot_request` already accepts the path and flags, I don't see what `krun_snapshot_create` adds.

src/devices/src/virtio/mmio.rs
vadika: Agreed — the comment is wrong, the workers keep running; only the vCPUs are frozen. Proposing to fix it with the `pause()`/`resume()` shape you suggested, built on the existing worker machinery:

- Worker `run()` returns its `DeviceQueue` on stop instead of dropping it. `VirtioDevice::pause()` signals `worker_stopfd`, joins, and returns the live queues; `resume(mem, interrupt, queues)` re-spawns the worker with them (subsumes `resume_after_restore`).
- Every worker already blocks in `epoll.wait({virtq_evt, stop_fd})`, processes a batch synchronously, then loops — so `stop_fd` fires at a descriptor boundary with `add_used` done and no in-flight host I/O. The queue indices are already consistent with guest memory there, so `save()` snapshots them directly (drops the used-ring reconstruction and the false comment).
- Capture: freeze vCPUs → `pause()` every device → snapshot queues + state → dump RAM. The `Resume` flag then becomes `resume()` on the same path.

The boundary property holds for block and fs (synchronous). Still auditing console (per-port I/O threads), net, and vsock for async completions that land after the worker returns to epoll — those need `pause()` to drain, not just join. Anything that can't quiesce → `pause()` errors → snapshot rejected.

Does this match what you had in mind?

src/devices/src/virtio/mmio.rs
vadika: Fixed. The name was a device kind (`fs`, `block`), not unique. The manifest now records the ordered device topology, and restore matches saved state to live devices by position after validating the fresh config reproduces that topology exactly (kind, count, order). Duplicate, missing, extra, or reordered devices are rejected instead of mis-binding state.

include/libkrun.h
vadika: Agreed, dropped it. `krun_snapshot_create` is gone; the control channel is registered for every VM in `krun_start_enter`, so `krun_snapshot_request(ctx, path, flags)` on a running VM is the only trigger, and there's no built-in guest trigger. The C API is now just `krun_snapshot_request` + `krun_restore`.

vadika:
@mtjhrc , @slp  asking for another round of revews :)

src/devices/src/virtio/mmio.rs
mtjhrc: > Worker run() returns its DeviceQueue on stop instead of dropping it. VirtioDevice::pause() signals worker_stopfd, joins, and returns the live queues; resume(mem, interrupt, queues) re-spawns the worker with them (subsumes resume_after_restore).

Yes pretty much. But I wouldn't concentrate on a `worker_stopfd` being used by every device. The mechanism here would just be to call the VirtioDevice::pause/VirtioDevice::resume method. How it signals it's internal worker thread(s) is up to each device. Though currently yes, an EventFd make sense in every device I guess.

Also if a device has some state owned by it's worker threads (e.g. like an unsent buffer) the  I can think of 2 ways to handle it in `VirtioDevice::pause`:
 - upon stopping the worker thread apart from returing the Queue(s) (as we must) we also return our internal state object and save it in the device. That object will be serialized when Snapshotable::save is called.
 - we don't actually exit from the device worker thread, instead we just block in the worker thread and let it serialize it's data and sends it via a channel. I imagine we may need this pattern for virtio GPU. 

> Every worker already blocks in epoll.wait({virtq_evt, stop_fd}), processes a batch synchronously, then loops — so stop_fd fires at a descriptor boundary with add_used done and no in-flight host I/O.

Yea so ideally we would be able to serialize the state of a partially consumed descriptor (before add_used) was called. This is however quite complex (depending on the device). I'm not sure what is a good immediate solution, however getting snapshot working (at least for some device would be really nice now).  

I'm hoping to eventually rewrite vsock and virtio-net using some future abstraction like this: https://github.com/libkrun/libkrun/pull/530/changes/7c80d974d50cc8b8113be0cdf39788e3476c2a4f (PR #530, unfinished 😞 ). Such utility would allow us to serialize even states like this: descriptor chain from guest has N bytes written on the host but hasn't been `add_used` yet because it is not finished yet (X bytes remaining). That PR never got merged because I hit performance problems, mainly on macOS so I need to revisit it for 2.0 hopefully.  

> those need pause() to drain, not just join.

Regarding draining anything, I don't think that is a good strategy. Imagine a pathological scenario where we have a host application communicating with a guest application over a unix socket (exposed to the guest as a virtio console port or as a vsock). Once you stop the guest it stops responding, but the host application might need the guest to respond first before continuing to process the rest of the messages in the socket's queue. But since the socket is full libkrun can't drain anything into the socket. So we basically hit a deadlock here.

> Capture: freeze vCPUs → pause() every device → snapshot queues + state → dump RAM

Sounds good 👍

vadika: Implemented, folded into this PR (last commit). `VirtioDevice` gained `pause()` (stop workers, return the live queues) and `resume(queues)`; the transport snapshots the exact indices off the returned queues instead of reconstructing them, and `save_state`/`restore_state` carry per-device state (the console's open ports). Kept  constraints: the mechanism is per-device (no `worker_stopfd` in the trait), and `pause` never drains — the console tx thread now polls the stop fd alongside the output fd and abandons a blocked write rather than deadlocking.

Done for block, virtio-fs, console, rng, balloon; the rest reject the snapshot. Partial-descriptor state and the vsock/net rewrite you mentioned (#530) are still out of scope. A `snapshot-fs-capture` test snapshots while the fs worker is mid-I/O.

vadika:
@mtjhrc , @slp  asking for another review round?

vadika:
@mtjhrc , @slp  -- politely asking for another review round :-)

slp:
@vadika Sorry for the late answer. Since this PR is touching critical code paths, we've decided to delay reviewing it until we're working on 2.1. That is, after 2.0 is tagged and released (something that should happen in ~4 weeks), we'll come back to this one.

I know it's a bummer, but we need to keep the focus on 2.0 for a little bit more. Thanks!

vadika:
> @vadika Sorry for the late answer. Since this PR is touching critical code paths, we've decided to delay reviewing it until we're working on 2.1. That is, after 2.0 is tagged and released (something that should happen in ~4 weeks), we'll come back to this one.
> 
> I know it's a bummer, but we need to keep the focus on 2.0 for a little bit more. Thanks!

LGTM, ping me when ready!

ChengyuZhu6:
@vadika , @mtjhrc , @slp
I’m interested in working on the KVM side, starting with pause/resume on Linux.
The KVM vCPU code already has most of the basic machinery, but it isn’t wired up to `krun_vmm_handle_pause / resume` yet.
I’d keep this separate from #762. If this sounds useful, I can start looking into it once the KVM work is ready to begin.

PR #765, opened 2026-06-29
macos: support VMs with more than 64GB of RAM

### LINKED ISSUE #764: Guest RAM capped at ~62 GiB on Apple Silicon: default 36-bit IPA not raised on macOS 15+ (40-bit) hosts
## Summary

On macOS / Apple Silicon, libkrun creates every HVF VM with the **default 36-bit guest IPA** (Intermediate Physical Address = 64 GiB of guest physical address space). In practice this caps a single guest at roughly **62-63 GiB of usable RAM regardless of host memory**, because the guest physical layout (RAM base + any shared GPU/VRAM region) has to fit under that 64 GiB window.

Apple's Hypervisor.framework has supported a configurable, larger guest IPA since macOS 13, and reports a **40-bit (1 TiB)** maximum on capable hardware running macOS 15+. So the ~62 GiB ceiling is a **software default, not a hardware or OS limit** on modern large-memory Macs. This issue asks libkrun to support guests larger than that window on machines that have the memory.

## Why it matters

Large-memory unified-memory Macs (M-series Max/Ultra, 96-512 GiB) are a primary libkrun/krunkit use case for GPU-accelerated LLM inference. The 36-bit IPA bounds a single VM's RAM well below what the hardware can hold, which in turn bounds model selection. On a 128 GiB Mac, more than half the machine's memory is unreachable by a single guest purely because of the default IPA size.

## Current behavior (observation)

`HvfVm::new` (`src/hvf/src/lib.rs`) creates the VM config and calls `hv_vm_create` without ever calling `hv_vm_config_set_ipa_size`, so HVF uses its 36-bit default. Two related observations:

- The guest-IPA config functions (`hv_vm_config_get_max_ipa_size` / `hv_vm_config_set_ipa_size`) are not present in the generated bindings (`src/hvf/src/bindings.rs`) -- only `hv_ipa_t` and the memory-map functions are.
- `HvfVm::new` does not currently receive the guest RAM size, so the VM-creation path has no input that would let it choose a larger IPA. The machine config (with `mem_size_mib`) is available at the caller (`src/vmm/src/macos/vstate.rs`).

These are just notes on where the 36-bit default currently comes from -- the design of any fix is entirely up to the maintainers.

## Evidence the hardware / OS supports more

Measured on an M4 Max / 128 GiB / macOS 26.5.1:

```bash
$ sysctl kern.hv.ipa_size_4k kern.hv.ipa_size_16k
kern.hv.ipa_size_4k: 1099511627776     # 2^40 = 1 TiB  (4K granule)
kern.hv.ipa_size_16k: 4398046511104    # 2^42 = 4 TiB  (16K granule)
```

These are the maximum IPA sizes the kernel reports for this host. Note a version nuance worth being aware of: macOS 13/14 reported a 39-bit max that is not a valid ARM `PARange` value and is unusable; **macOS 15+ reports a valid 40-bit max**, which is the version where larger guests became viable. On the M1 base part `ID_AA64MMFR0_EL1.PARange` is 36-bit (the floor), so any larger-IPA path has to degrade gracefully on hardware/OS that does not support it.

## Prior art

QEMU's HVF backend addressed the same limitation: "hvf: arm: Support creating VMs with 64+GB of RAM on macOS 15+" (merged late 2024). Linked only as a reference point for how another HVF consumer handled it -- not as a prescription for libkrun:

- https://www.mail-archive.com/qemu-devel@nongnu.org/msg1062741.html
- https://www.mail-archive.com/qemu-devel@nongnu.org/msg1094318.html

## Downstream note

krunkit currently hardcodes the 36-bit assumption (it rejects RAM above 61440 MiB and sizes the VRAM region against a ~62 GiB budget). If/however libkrun raises the effective ceiling, krunkit would need a corresponding adjustment -- flagged here only so the two are tracked together; the libkrun-side decision drives it.

slp:
Check if the last address in the VM goes beyond the 64GB line and, if that's the case, try to use a 40-bit IPA, if supported by the system.

This enables us to both run VMs with more than 64GB of RAM and use GPU windows larger than such value.

Be conservative and keep using a 36-bit IPA if the guest doesn't require a larger one.

Fixes: #764

### DIFF
--- a/src/hvf/src/lib.rs
+++ b/src/hvf/src/lib.rs
@@ -22,6 +22,7 @@ use std::fmt::{Display, Formatter};
 use std::sync::{Arc, LazyLock};
 use std::time::Duration;
 
+use arch::ArchMemoryInfo;
 #[cfg(all(target_arch = "aarch64", target_os = "macos"))]
 use arch::aarch64::sysreg::{SYSREG_MASK, sys_reg_name};
 use log::debug;
@@ -104,6 +105,10 @@ const EC_AA64_BKPT: u64 = 0x3c;
 pub enum Error {
     EnableEL2,
     FindSymbol(libloading::Error),
+    IpaGetDefault,
+    IpaGetMaximum,
+    IpaSet,
+    IpaTooLow,
     MemoryMap,
     MemoryUnmap,
     NestedCheck,
@@ -127,6 +132,13 @@ impl Display for Error {
         match self {
             EnableEL2 => write!(f, "Error enabling EL2 mode in HVF"),
             FindSymbol(err) => write!(f, "Couldn't find symbol in HVF library: {err}"),
+            IpaGetDefault => write!(f, "Error getting IPA default value"),
+            IpaGetMaximum => write!(f, "Error getting IPA maximum value"),
+            IpaSet => write!(f, "Error setting IPA value"),
+            IpaTooLow => write!(
+                f,
+                "The VM requires an IPA larger than the maximum supported in this system"
+            ),
             MemoryMap => write!(f, "Error registering memory region in HVF"),
             MemoryUnmap => write!(f, "Error unregistering memory region in HVF"),
             NestedCheck => write!(
@@ -238,7 +250,7 @@ static HVF: LazyLock<libloading::Library> = LazyLock::new(|| unsafe {
 });
 
 impl HvfVm {
-    pub fn new(nested_enabled: bool) -> Result<Self, Error> {
+    pub fn new(arch_mem_info: &mut ArchMemoryInfo, nested_enabled: bool) -> Result<Self, Error> {
         let config = unsafe { hv_vm_config_create() };
         if nested_enabled {
             let set_el2_enabled: libloading::Symbol<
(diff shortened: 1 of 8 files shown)


src/hvf/src/lib.rs
```diff
@@ -378,7 +420,45 @@ impl HvfVcpu<'_> {
...
+        };
+        if ret != HV_SUCCESS {
+            return Err(Error::VcpuInitialRegisters);
+        }
+        let ret = unsafe {
+            hv_vcpu_set_sys_reg(
+                self.vcpuid,
+                hv_sys_reg_t_HV_SYS_REG_ID_AA64MMFR0_EL1,
+                (val & !0xf) | (ipa_index as u64),
+            )
```
lstocchi, line 456: Not an expert about macos so i gave a look to understand what you're doing. I read that `ID_AA64MMFR0_EL1` is read-only but it seems you're setting its value. What am i missing?

slp: It's read-only from the perspective of the guest, but the hypervisor can set the initial values of most system registers before starting the VM.

jakecorrenti (approved):
Gave the code a quick sanity check, didn't run it (don't have a mac within enough memory anyways), but don't seem to see anything that might be problematic.

PR #766, opened 2026-06-29
hvf: fix various UBs

slp:
Fix various UBs from casting non-mutable variables to mutable.

### DIFF
--- a/src/hvf/src/lib.rs
+++ b/src/hvf/src/lib.rs
@@ -333,7 +333,7 @@ pub struct HvfVcpu<'a> {
 impl HvfVcpu<'_> {
     pub fn new(mpidr: u64, nested_enabled: bool) -> Result<Self, Error> {
         let mut vcpuid: hv_vcpu_t = 0;
-        let vcpu_exit_ptr: *mut hv_vcpu_exit_t = std::ptr::null_mut();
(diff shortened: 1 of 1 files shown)


src/hvf/src/lib.rs
```diff
@@ -349,7 +349,7 @@ impl HvfVcpu<'_> {
         let ret = unsafe {
             hv_vcpu_create(
                 &mut vcpuid,
-                &vcpu_exit_ptr as *const _ as *mut *mut _,
+                &mut vcpu_exit_ptr as *mut *mut _,
```
mtjhrc, line 352: Since you are already changing this - I would suggest using `&raw mut` instead. https://doc.rust-lang.org/std/primitive.pointer.html#3-create-it-using-raw

mtjhrc (changes requested):

src/hvf/src/lib.rs
slp: As the variables are stack allocated, I think this cast is more appropriate than `&raw mut`.

mtjhrc: > As the variables are stack allocated, I think this cast is more appropriate than `&raw mut`.

Functionally both should work the same here. (note that there are memory aliasing differences but they make no difference in this case). I prefer  `&raw mut` because you wouldn't need any cast here at all.
But anyways I won't block the PR on this.

mtjhrc (approved):

PR #767, opened 2026-06-29
vmm,hvf,libkrun: add krun_vm_pause / krun_vm_resume

vadika:
Live, thread-safe pause/resume for a running VM on macOS/HVF. Suggested by @mtjhrc in #762 — the snapshot path needs this freeze primitive anyway, so it's split out to land first.

**API**
- `krun_vm_pause(ctx_id)` — freeze every vCPU at an instruction boundary.
- `krun_vm_resume(ctx_id)` — wake them.

Out-of-band (call from another thread while `krun_start_enter` blocks), idempotent, `-ENOTSUP` off macOS.

**How**
Per-VM eventfds signalled by the C API; the VM event loop runs the freeze/wake. Pause breaks each vCPU out of HVF (`vcpu_request_exit`) and parks it; resume advances the vtimer offset by the time spent paused, so the guest's CNTVCT stays continuous and armed timers don't fire en masse on wake. Device worker threads are notification-driven, so a frozen guest leaves them idle — no per-device handling.

**Test**
`vm-pause` (macOS/HVF): pauses a running guest mid-workload from another thread, holds past its natural completion, resumes. Asserts the guest woke and finished, its virtual clock stayed continuous across the pause, and the pause actually held it. Passes locally; `clippy -D warnings` and `fmt` clean.

Follow-up: #762 rebases on top and drops the now-shared primitives.

### DIFF
--- a/src/libkrun/src/lib.rs
+++ b/src/libkrun/src/lib.rs
@@ -39,10 +39,14 @@ use std::sync::LazyLock;
 use std::sync::Mutex;
 use std::sync::atomic::{AtomicI32, Ordering};
 use utils::eventfd::EventFd;
+#[cfg(target_os = "macos")]
+use utils::pollable_channel::PollableChannelSender;
 #[cfg(target_os = "windows")]
 use utils::windows::AsRawFd;
 #[cfg(target_os = "windows")]
 use utils::windows::SendHandle;
+#[cfg(target_os = "macos")]
+use vmm::VmCtl;
 use vmm::resources::{
     DefaultVirtioConsoleConfig, PortConfig, SerialConsoleConfig, TsiFlags, VirtioConsoleConfigMode,
     VmResources, VsockConfig,
@@ -425,6 +429,13 @@ fn with_cfg(ctx_id: u32, f: impl FnOnce(&mut ContextConfig) -> i32) -> i32 {
 
 static CTX_MAP: Lazy<Mutex<HashMap<u32, ContextConfig>>> = Lazy::new(|| Mutex::new(HashMap::new()));
 static CTX_IDS: AtomicI32 = AtomicI32::new(0);
+
+/// Out-of-band control channel for running VMs, keyed by ctx id. Populated by
+/// `krun_start_enter` for every macOS VM so `krun_vm_pause` / `krun_vm_resume`
+/// can reach it from another thread.
+#[cfg(target_os = "macos")]
+static VM_CTL: Lazy<Mutex<HashMap<u32, PollableChannelSender<VmCtl>>>> =
+    Lazy::new(|| Mutex::new(HashMap::new()));
 
 fn log_level_to_filter_str(level: u32) -> &'static str {
     match level {
@@ -2965,6 +2976,14 @@ pub extern "C" fn krun_start_enter(ctx_id: u32) -> i32 {
     #[cfg(any(feature = "amd-sev", feature = "tdx"))]
     vmm::worker::start_worker_thread(_vmm.clone(), _receiver.clone()).unwrap();
 
+    // Register the control channel for every macOS VM so another thread can
+    // freeze/wake this guest while `krun_start_enter` blocks here.
+    #[cfg(target_os = "macos")]
+    VM_CTL
+        .lock()
+        .unwrap()
+        .insert(ctx_id, _vmm.lock().unwrap().vm_ctl_sender());
+
     loop {
         match event_manager.run() {
             Ok(_) => {}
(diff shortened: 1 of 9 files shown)


vadika:
@mtjhrc @slp politely asking for a review :)

src/vmm/src/macos/vstate.rs
```diff
@@ -485,29 +496,68 @@ impl Vcpu {
...
-            } else {
-                receiver.recv().unwrap();
+        if !self.vcpu_list.should_wait(hvf_vcpuid) {
+            return;
+        }
+        let paused = if let Some(timeout) = timeout {
+            select! {
+                recv(receiver) -> _ => false,
+                recv(self.event_receiver) -> ev => matches!(ev, Ok(VcpuEvent::Pause)),
+                recv(after(timeout)) -> _ => false,
```
mz-pdm, line 518: So now the channel is expected to get disconnected when the vCPU is paused? Can the WFE channel be disconnected also for another reason that should be handled differently?

mz-pdm (commented):
I can't provide an authoritative review as I'm not an expert here but as far as I understand the PR (the source comments help a lot!), I can't see any obvious problem.

src/vmm/src/macos/vstate.rs
vadika: Good catch — no, the WFE channel is never disconnected during normal operation (senders live in VcpuList/interrupt controller for the VM lifetime; pause goes through the separate event channel). My `select!` regressed the old explicit `Disconnected => panic!` into `_ => false`, which would silently busy-spin instead of failing loud. Fixed: the receiver arm now `.expect("WFE channel closed unexpectedly")`, restoring prior semantics.

mz-pdm (dismissed):
AFAICT

src/libkrun/src/lib.rs
```diff
@@ -426,6 +426,13 @@ fn with_cfg(ctx_id: u32, f: impl FnOnce(&mut ContextConfig) -> i32) -> i32 {
 static CTX_MAP: Lazy<Mutex<HashMap<u32, ContextConfig>>> = Lazy::new(|| Mutex::new(HashMap::new()));
 static CTX_IDS: AtomicI32 = AtomicI32::new(0);
 
+/// Out-of-band live pause/resume control for running VMs, keyed by ctx id:
+/// `(pause_evt, resume_evt)`. Populated by `krun_start_enter` for every macOS VM
+/// so `krun_vm_pause` / `krun_vm_resume` can reach it from another thread.
+#[cfg(target_os = "macos")]
+static PAUSE_CTL: Lazy<Mutex<HashMap<u32, (EventFd, EventFd)>>> =
+    Lazy::new(|| Mutex::new(HashMap::new()));
```
mtjhrc, line 438: I was thinking about whether there's a race condition where:
```
krun_vm_pause(ctx_id);
// immediately, without any sleep:
krun_vm_resume(ctx_id);
```
could hang — specifically, if kqueue delivered the resume event before the pause event when both pipes are ready in the same batch. The resume would be a no-op (VM not paused yet), then the pause would fire with no resume coming.

For single-threaded callers I don't think this can happen — I think write() synchronously enqueues the the file descriptor readiness before returning, so the pause event is always ahead of the resume event in the kqueue ready list. For multi-threaded callers (pause and resume from different threads simultaneously), the PAUSE_CTL mutex serializes the writes, preserving the ordering.

So I think the current code is safe, but it was not obvious to me why. Personally I would suggest a single PollableChannel<Enum { Pause, Resume }> (https://github.com/mtjhrc/libkrun/blob/622a0430/src/utils/src/pollable_channel.rs#L12-L13) instead of two separate eventfds — ordering would be more obviously correct. This could then probably also be used for the snapshot mechanism PR? What do you think?

mtjhrc (changes requested):
Great work here! Test passes. Sorry it took a while I was on vacation as someone already told you on Matrix. 

I looked at the vCPU pausing code and the time offset, to me it looks correct but I wasn't familiar in that area either. 

I am not sure about using the two EventFd's for pause/resume - see the comment.

src/libkrun/src/lib.rs
vadika: Agreed: `polly` calls `process()` once per ready fd in whatever order the poller returns them, so with two fds the ordering is poller-defined, not write-defined. `PAUSE_CTL`s mutex serialized the writes but nothing downstream of that. Your `resume();pause();` inversion hangs the VM, and `resume()` before `pause()` is order-dependent for the same reason.

Switched to a single `PollableChannel<VmCtl>` . Enum is `VmCtl { Pause, Resume }`, so the snapshot PR can add variants without touching the plumbing.

One prerequisite: `pollable_channel` had no in-tree users and does not work on macOS as-is. `EventFd::new` there tested `flag == EFD_NONBLOCK`, so `EFD_NONBLOCK | EFD_SEMAPHORE` left the pipe blocking and `try_recv()` would block on an empty pipe while holding the queue mutex. Fixed as a separate commit. (`send_many` is also wrong on the pipe emulation — one 8-byte token for N messages — but nothing calls it, so I left it.)

Test now does a `krun_vm_pause()` immediately followed by `krun_vm_resume()` with no sleep, which the old code passed only by luck.

mtjhrc (approved):

PR #768, opened 2026-06-30
init: Added GID and UID parsing

RiverNewbury:
Added parsing for uids and gids from the config file to address [issue 123](https://github.com/libkrun/libkrun/issues/123). 

Currently this works for standard use cases (i.e. becoming a user and never changes but doesn't work with more advanced changing of users. The below shows the behaviour. Also no matter the user we become inside the VM the uid the file is published under remains the same. which isn't quite what I expected.

EDIT: I should clarify that in these examples I entered the container as 1000:1000 and then used sudo to write a file as root.

```
-rw-r--r--. 1 10999 10999 0 Jun 30 15:33 crun-myuser
-rw-r--r--. 1 user  user  0 Jun 30 15:33 crun-root
-rw-r--r--. 1 10999 10999 0 Jun 30 15:32 krun-myuser
-rw-r--r--. 1 10999 10999 0 Jun 30 15:32 krun-root
```

I **think** this is the code required to get this working inside libkrun but other code may be required in crun as said in the original issue.

### DIFF
--- a/src/init_blob/init/init.c
+++ b/src/init_blob/init/init.c
@@ -29,6 +29,8 @@
 #include <sys/types.h>
 #include <sys/wait.h>
 
+#include <grp.h>
+
 #if __linux__
 #include <linux/reboot.h>
 #include <linux/vm_sockets.h>
@@ -157,6 +159,36 @@ int unmount_config_iso()
     return unmount("/mnt", 0);
 }
 #endif
+
+static void setgid_wrapper(const char *gid)
+{
+    char *endptr;
+    errno = 0;
+    unsigned long val = strtoul(gid, &endptr, 10);
+    if (errno != 0 || *endptr != '\0' || val > (gid_t)-1) {
+        printf("Can't parse GID\n");
+        exit(125);
+    }
+    if (setgroups(0, NULL) != 0 || setgid((gid_t)val) != 0) {
+        printf("Can't set GID\n");
+        exit(125);
+    }
+}
+
+static void setuid_wrapper(const char *uid)
+{
+    char *endptr;
+    errno = 0;
+    unsigned long val = strtoul(uid, &endptr, 10);
+    if (errno != 0 || *endptr != '\0' || val > (uid_t)-1) {
+        printf("Can't parse UID\n");
+        exit(125);
+    }
+    if (setuid((uid_t)val) != 0) {
+        printf("Can't set UID\n");
+        exit(125);
+    }
+}
 
 static void set_rlimits(const char *rlimits)
 {
@@ -860,8 +892,8 @@ char **concat_entrypoint_argv(char **entrypoint, char **config_argv)
     return argv;
 }
 
-static int config_parse_file(char ***argv, char **workdir,
-                             const char *config_file)
+static int config_parse_file(char ***argv, char **workdir, char **uid,
+                             char **gid, const char *config_file)
 {
     jsmn_parser parser;
     jsmntok_t *tokens;
@@ -870,7 +902,8 @@ static int config_parse_file(char ***argv, char **workdir,
     off_t data_len;
     char **config_argv;
     char **entrypoint;
-    int parsed_env, parsed_workdir, parsed_args, parsed_entrypoint;
+    int parsed_env, parsed_workdir, parsed_args, parsed_entrypoint, parsed_uid,
+        parsed_gid;
     int num_tokens;
     int ret = -1;
     int fd;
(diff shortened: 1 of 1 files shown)


src/init_blob/init/init.c
```diff
@@ -158,6 +158,22 @@ int unmount_config_iso()
 }
 #endif
 
+static void setgid_wraper(const char *gid)
```
jakecorrenti:
```suggestion
static void setgid_wrapper(const char *gid)
```

src/init_blob/init/init.c
```diff
@@ -158,6 +158,22 @@ int unmount_config_iso()
...
+{
+    if (setgid(strtoull(gid, (char**) NULL, 10)) != 0) {
+        printf("Can't set GID");
+    }
+}
+
+static void setuid_wraper(const char *uid)
```
jakecorrenti:
```suggestion
static void setuid_wrapper(const char *uid)
```

src/init_blob/init/init.c
```diff
@@ -158,6 +158,22 @@ int unmount_config_iso()
 }
 #endif
 
+static void setgid_wraper(const char *gid)
+{
+    if (setgid(strtoull(gid, (char**) NULL, 10)) != 0) {
+        printf("Can't set GID");
```
jakecorrenti: If `setgid()` fails, and same with `setuid()`, we're still going to exhibit the buggy behavior to the user. I think we should be exiting here instead of just printing the error and falling back to root (0).

You're also not checking `errno` for EINVAL or ERANGE (check strtoul(3))

jakecorrenti (changes requested):
Couple of comments.

Also make sure this is formatted as there's some extra newlines here and there.

src/init_blob/init/init.c
RiverNewbury: Yeah I hadn't thought about that. Have updated it not to use the same `exit(125)` as if `setup_redirects` fails - this seems to be the one given if the docker daemon has an internal issue which seems to make sense to me.

RiverNewbury:
I think that should be all the requested changes done - sorry about the silly spelling mistakes and not formatting - I'm not very used to C codebases so had to do a bit of digging to work out even what code formatters were available/ the projected used. Hope every thing is all good now :upside_down_face:

jakecorrenti:
Hi @RiverNewbury thanks for getting those changes done. I apologize for taking so long to get back to this, but do you think you could rebase and see if that fixes the CI failures? Thanks :)

RiverNewbury:
Hi - no worries about the delay! We're all busy IRL and I don't get time too much to check out PRs etc either. Unless I'm missing something I believe this is still on the most recent version of `stable-1.19.x` - the git log shows my commit is on top of `728df81` which is the most recent in the branch I'm merging into. I'm afraid I dont' know wy the code quality etc checks failed...

jakecorrenti (approved):
Not on my laptop at the moment, but the code LGTM! Just rebased your branch on top of the CI fixes.

src/init_blob/init/init.c
```diff
@@ -158,6 +158,26 @@ int unmount_config_iso()
 }
 #endif
 
+static void setgid_wrapper(const char *gid)
+{
+    errno = 0;
+    gid_t gid_int = strtoull(gid, (char **)NULL, 10);
```
slp: Not all versions of `strtoull` set errno to EINVAL when it can't make the conversion. It'd be safer passing `endptr` and checking it in addition to errno.

src/init_blob/init/init.c
```diff
@@ -1542,6 +1581,15 @@ int main(int argc, char **argv)
             exit(125);
         }
 #endif
+        // Set group first as otherwise we don't have permission to set user
+        // It wants to go after all our priviledged activities
```
slp: priviledged -> privileged

slp (changes requested):

src/init_blob/init/init.c
RiverNewbury: Ah didn't realise that! Thanks for the spot!

slp:
@RiverNewbury I don't see any changes addressing the comments. Perhaps you missed committing the changes before pushing them?

RiverNewbury:
> @RiverNewbury I don't see any changes addressing the comments. Perhaps you missed committing the changes before pushing them?

You're exactly right - I did the ammend commit and it changed the hash but forgot to add beforehand. Sorry about that

slp (approved):
LGTM. Added some small tweaks myself. Thanks!

Issue #772, opened 2026-07-06
vsock guest-host stream deadlock at 240 KB per connection (credit starvation + dropped bytes on partial writes


oliver-schick:
### Summary
On a single vsock connection, a **guest→host** stream **deadlocks after exactly 245760 bytes (240 KB)** whenever the guest sends faster than the host-side onward socket drains. It affects both networking paths: TSI-hijacked INET (`tsi_stream.rs`) and explicit `krun_add_vsock_port` (`unix.rs`). Separately, data is **silently dropped/corrupted** on partial/`EAGAIN` writes (observed downstream as truncated payloads and `TLSV1_ALERT_PROTOCOL_VERSION` from TLS servers).

### Environment
- libkrun **v1.18.0** (the code paths below are unchanged on `main`)
- libkrunfw **5.5.0** (guest kernel 6.12.91) — reproduced regardless of guest-kernel version
- Host: Linux 6.14 x86_64, KVM, rootless

### Reproduce
Boot any VM (implicit TSI, or `krun_add_vsock_port`). From the guest, open **one** connection to a host endpoint and `send()` a payload > ~240 KB where the host consumer exerts any backpressure (a normal socket whose reader isn't instantaneously draining). The transfer hangs; the guest blocks in `send()` after exactly 245760 bytes have crossed. `< 245760` bytes always succeeds. Verified with a **bare `AF_VSOCK` client in the guest** (no extra userspace on either side), so it is not a consumer-side bug.

Smaller single writes (≤16 KB, e.g. TLS records) crossing a *fast* consumer can mask it until cumulative in-flight bytes hit the window.

### Root cause — `src/devices/src/virtio/vsock/{tsi_stream,unix}.rs`, `sendmsg`
```rust
match send(self.fd.as_raw_fd(), buf, flags) {
    Ok(sent) => {
        if sent != buf.len() {
            error!("couldn't set everything: buf={}, sent={}", buf.len(), sent); // <-- remainder DROPPED
        }
        self.tx_cnt += Wrapping(sent as u32);
        sent as i32
    }
    Err(err) => { /* returns -errno; on EAGAIN the packet is dropped */ }
}
...
if ret > 0 && (self.tx_cnt - self.last_tx_cnt_sent).0 as usize >= (defs::CONN_TX_BUF_SIZE / 2) {
    // credit update only here
}
```
[...]

PR #773, opened 2026-07-06
vsock: don't drop guest-host bytes on partial/EAGAIN writes

oliver-schick:
## vsock: don't drop guest→host bytes on partial/`EAGAIN` writes

Fixes #<ISSUE_NUMBER> (the remaining half — see note on the deadlock below).

### Bug
Both vsock host proxies (`tsi_stream.rs`, `unix.rs`) flush guest→host data with a single non-blocking
`send()`. On a short write or `EAGAIN`, `sendmsg` logs `couldn't set everything` and **discards the unsent
remainder** — there is no tx buffer and no `EPOLLOUT`-driven resume. Under load this silently corrupts the
stream; downstream we observed it as `TLSV1_ALERT_PROTOCOL_VERSION` from TLS servers mid-transfer.

### Fix
Buffer the unsent tail and finish it event-driven (no blocking, no busy-wait):
- `sendmsg` appends the packet to a per-connection `tx_buf` and flushes what it can;
- on `EAGAIN` the proxy registers `EventSet::OUT` and completes the flush from `process_event` when the
  socket is writable again;
- a single `rearm()` helper adds `OUT` to whatever RX interest each existing re-arm site already requests,
  so the credit-pause / connect / accept state machine is unchanged;
- `tx_cnt` advances only for bytes actually written, and buffered-but-unwritten bytes still hold guest
  credit (no `fwd_cnt` for them), so `tx_buf` is naturally bounded by the connection's window — correct
  end-to-end backpressure, bounded memory. The credit logic itself is left exactly as-is.

### On the deadlock in the linked issue
The issue also reports a guest→host **deadlock at exactly 245760 B**: the guest operates on a ~256 KiB
window (the AF_VSOCK default) while `set_buf_alloc` advertises `CONN_TX_BUF_SIZE`, and the credit-update
threshold was `CONN_TX_BUF_SIZE/2`. On **`main` this is already largely addressed** — the threshold is now
`peer_buf_alloc / 2`, i.e. window-based — so this PR deliberately does **not** touch credit logic; it only
fixes the still-present byte-drop.

### Validation
Runtime-validated on **v1.18.0** (where the credit bug also reproduced): guest→host 256 KiB / 512 KiB /
1 MiB single connections that previously wedged / corrupted now complete, and TLS through the tunnel no
longer sees `protocol_version` alerts. On current **main** the patch **applies and compiles cleanly**
(`cargo build --release` green). I did not runtime-test on main — my test harness drives the v1.18.0 VMM ABI
and can't boot the current dev VMM — so a maintainer runtime check on main would be appreciated. Happy to
split `unix.rs` out, or adjust the buffering approach, as you prefer.

### DIFF
--- a/src/devices/src/virtio/vsock/tsi_stream.rs
+++ b/src/devices/src/virtio/vsock/tsi_stream.rs
@@ -58,6 +58,8 @@ pub struct TsiStreamProxy {
     push_cnt: Wrapping<u32>,
     pending_accepts: u64,
     unixsock_path: Option<PathBuf>,
+    // Unsent guest->host bytes, flushed on EventSet::OUT — never dropped (see drain_tx).
+    tx_buf: Vec<u8>,
 }
 
 impl TsiStreamProxy {
@@ -138,6 +140,7 @@ impl TsiStreamProxy {
             push_cnt: Wrapping(0),
             pending_accepts: 0,
             unixsock_path: None,
+            tx_buf: Vec::new(),
         })
     }
 
@@ -176,6 +179,7 @@ impl TsiStreamProxy {
             push_cnt: Wrapping(0),
             pending_accepts: 0,
             unixsock_path: None,
+            tx_buf: Vec::new(),
         }
     }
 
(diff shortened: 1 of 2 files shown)


jakecorrenti (changes requested):
Hi @oliver-schick, a couple things while I do a full review:

Please make sure you're signing your commits: `git commit --amend -s`. Also, please make sure to add some information in the body of the commit using imperative mood.

oliver-schick:
Thanks @jakecorrenti! Addressed both:

- **Sign-off:** amended with `git commit --amend -s` — DCO should be green now.
- **Commit body:** rewrote it in imperative mood, describing the bug (single non-blocking `send()` drops the unsent tail and never re-arms `EventSet::OUT`) and the event-driven fix (buffer the tail, resume the flush from `process_event`, credit logic untouched).

Ready for your full review whenever you have time. Happy to split `unix.rs` into its own commit, or adjust the buffering approach, if you'd prefer.

For background, I've added a self-contained client+server reproducer and the real-world context over in #772 (it reproduces on bare `AF_VSOCK` as well as TSI) — verified there that this PR turns the stall into a clean full-stream transfer on v1.18.0.

oliver-schick:
Withdrawing this. Re-tested on stock v1.19.4 (which has 1bebfaac but not this change): the guest->host stream completes with no byte loss, so the deadlock I was chasing was entirely the credit-threshold bug 1bebfaac fixed — not the dropped-tail path this PR touches.

The single non-blocking send() drop-on-short-write in sendmsg is still technically there, but I can't produce a case where it loses data on current releases, so there's nothing worth merging here.

Thanks @jakecorrenti for the review and @slp for pointing at the fix. Closing.

PR #776, opened 2026-07-09
virtio-blk: surface krun_add_disk2 block_id as device ID/serial

dunnry:
VIRTIO_BLK_T_GET_ID previously returned a file-derived image id, so guests saw arbitrary serials under /sys/block/*/serial and could not locate disks by the block_id passed to krun_add_disk2. Use the supplied block id as the device ID so exact-serial discovery works.

This path also exists on stable-1.19.x and stable-1.18.x, so this may be suitable for the cherry-pick-1.19.5 and cherry-pick-1.18.2 labels.

Tests:
- cargo fmt --all -- --check
- git diff --check origin/main..HEAD
- cargo test -p krun-devices --features blk disk_image_id

### DIFF
--- a/src/devices/src/virtio/block/device.rs
+++ b/src/devices/src/virtio/block/device.rs
@@ -145,20 +145,30 @@ impl DiskProperties {
         Ok(device_id)
     }
 
-    fn build_disk_image_id(disk_file: &File) -> Vec<u8> {
+    fn build_disk_image_id(disk_file: &File, block_id: &str) -> Vec<u8> {
         let mut default_id = vec![0; VIRTIO_BLK_ID_BYTES as usize];
-        match Self::build_device_id(disk_file) {
-            Err(_) => {
-                warn!("Could not generate device id. We'll use a default.");
+
+        // The public libkrun disk API accepts a caller-provided block_id. Make
+        // that the virtio-blk GET_ID value so Linux can expose it as
+        // /sys/block/<dev>/serial. Fall back to the historical backing-file
+        // derived id only for callers that pass an empty block_id.
+        let disk_id = if block_id.is_empty() {
+            match Self::build_device_id(disk_file) {
+                Err(_) => {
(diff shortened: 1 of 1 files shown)


src/devices/src/virtio/block/device.rs
```diff
@@ -469,3 +479,47 @@ impl VirtioDevice for Block {
...
+
+        let image_id = DiskProperties::build_disk_image_id(backing.as_file(), long_id);
+
+        assert_eq!(image_id.len(), VIRTIO_BLK_ID_BYTES as usize);
+        assert_eq!(
+            &image_id,
+            &long_id.as_bytes()[..VIRTIO_BLK_ID_BYTES as usize]
+        );
+    }
+}
```
mz-pdm, line 538: Would it make sense to add a test for the empty block id case?  To check that the result has the right size and nothing bad (like an unhandled error) happens?

mz-pdm (dismissed):
LGTM, just one optional comment.

src/devices/src/virtio/block/device.rs
dunnry: Added an empty block_id test that exercises the fallback path and asserts the returned ID has the virtio-blk fixed size.

mz-pdm (approved):

slp (approved):
LGTM, thanks!

mtjhrc:
@slp I marked this as "cherry-pick-1.19.5" as mentioned in the PR description so we don't forget about this. Not sure if that is correct assessment of whether this can break consumers, but we can still discuss that when compiling PRs for the release.

I think @dunnry would want this in a stable release.

slp:
> @slp I marked this as "cherry-pick-1.19.5" as mentioned in the PR description so we don't forget about this. Not sure if that is correct assessment of whether this can break consumers, but we can still discuss that when compiling PRs for the release.

Thanks!

PR #777, opened 2026-07-10
clippy: fix new -D warnings lints (byte_char_slices, redundant_reference)

vadika:
CI floats `rustup update stable`; the newer clippy (1.97) fails the `-D warnings` gates on three pre-existing spots unrelated to any feature work:

- `rutabaga_gralloc/formats.rs`: DRM fourcc consts under `byte_char_slices` — same bytes as a dereferenced byte string, values unchanged.
- `virtio/input/worker.rs`: `&buffer` passed to `{:?}` under `redundant_reference` — inlined.
- `vmm/builder.rs`: bz2 magic compared against `[b'B', b'Z', b'h']` under `byte_char_slices` — compared against `b"BZh"`.

Pulling them out so the clippy jobs go green independently of the PRs that keep tripping over them.


mtjhrc (approved):
LGTM, thanks!

PR #778, opened 2026-07-10
fs/macos: strip com.apple.* xattrs invalid in Linux guests

BinSquare:
## Problem

On macOS hosts, files carry Apple-private extended attributes in the `com.apple.*` namespace (e.g. `com.apple.provenance`, `com.apple.quarantine`). The virtio-fs macOS passthrough currently forwards these to the Linux guest verbatim.

That namespace has no meaning on Linux, and exposing it breaks xattr-aware guest tooling. For example, `rsync -X` reads the source xattrs and then calls `lsetxattr(..., "com.apple.*", ...)` on the destination; filesystems like tmpfs and overlayfs reject the unknown namespace with `EOPNOTSUPP`, producing spurious errors and non-zero exit codes for otherwise-valid copies.

## Fix

Treat `com.apple.*` as invisible to the guest:

- `getxattr` returns `ENODATA` for any `com.apple.*` name.
- `listxattr` omits `com.apple.*` entries (both the size computation and the returned buffer).

The guest then sees a consistent empty set for these attributes

### DIFF
--- a/src/devices/src/virtio/fs/macos/passthrough.rs
+++ b/src/devices/src/virtio/fs/macos/passthrough.rs
@@ -37,6 +37,8 @@ use super::super::multikey::MultikeyBTreeMap;
 const XATTR_KEY: &[u8] = b"user.containers.override_stat\0";
 const SECURITY_CAPABILITY: &[u8] = b"security.capability\0";
 
+const MACOS_XATTR_PREFIX: &[u8] = b"com.apple.";
+
 const UID_MAX: u32 = u32::MAX - 1;
 
 type Inode = u64;
@@ -2344,6 +2346,10 @@ impl FileSystem for PassthroughFs {
             return Err(linux_error(io::Error::from_raw_os_error(libc::EACCES)));
         }
 
+        if name.to_bytes().starts_with(MACOS_XATTR_PREFIX) {
+            return Err(linux_error(io::Error::from_raw_os_error(libc::EOPNOTSUPP)));
+        }
+
         let mut mflags: i32 = 0;
         if (flags as i32) & bindings::LINUX_XATTR_CREATE != 0 {
             mflags |= libc::XATTR_CREATE;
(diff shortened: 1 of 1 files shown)


mz-pdm (dismissed):
LGTM

src/devices/src/virtio/fs/macos/passthrough.rs
```diff
@@ -37,6 +37,17 @@ use super::super::multikey::MultikeyBTreeMap;
...
 
+/// macOS-private xattr namespace prefix. Attributes in this namespace (e.g.
+/// `com.apple.provenance`, `com.apple.quarantine`) have no meaning in a Linux
+/// guest. Exposing them causes xattr-aware Linux tools like `rsync -X` to
+/// attempt `lsetxattr(..., "com.apple.*", ...)` on destination filesystems
+/// (tmpfs, overlayfs) that reject the unknown namespace with EOPNOTSUPP,
+/// producing spurious errors and non-zero exit codes.
+///
+/// We strip this prefix from `listxattr` and return `ENODATA` from `getxattr`
+/// so that the Linux guest sees a consistent empty set of these attributes.
```
slp: This is already explained in the commit message, and the name of the constant is self-explanatory. Please drop the comment block.

src/devices/src/virtio/fs/macos/passthrough.rs
```diff
@@ -2400,6 +2411,10 @@ impl FileSystem for PassthroughFs {
             return Err(linux_error(io::Error::from_raw_os_error(libc::EACCES)));
         }
 
+        if name.to_bytes().starts_with(MACOS_XATTR_PREFIX) {
+            return Err(linux_error(io::Error::from_raw_os_error(libc::ENODATA)));
```
slp, line 2410: For safety and consistency, we also need to reject setting the attribute in `setxattr` with `ENOTSUP`, and removing it in `removexattr` with `ENODATA`.

slp (changes requested):
Thanks for the PR! The change makes sense to me, just requested minor changes.

src/devices/src/virtio/fs/macos/passthrough.rs
BinSquare: Dropped the comment as suggested

slp (approved):
LGTM, thanks!

PR #780, opened 2026-07-13
devices/rng: fix AsRawFd import for Windows compilation

lstocchi:
The device itself already works cross-platform since it uses rand::OsRng for entropy, which delegates to BCryptGenRandom on Windows. This commit just imports AsRawFd based on the platform.


dorindabassey (approved):
LGTM

mtjhrc (approved):
LGTM!

PR #781, opened 2026-07-14
lib: preserve libkrunfw loading errors

xhebox:


mz-pdm (commented):
Thank you for the patch, looks useful.  It might help for easier review and tracking to put some info into the commit message, even if the commit title looks obvious.

src/libkrun/src/lib.rs
```diff
@@ -138,30 +139,47 @@ pub struct KrunfwBindings {
...
+impl Default for KrunfwBindingsResult {
+    fn default() -> Self {
+        Self(KrunfwBindings::new())
+    }
+}
+
+impl KrunfwBindingsResult {
+    fn as_ref(&self) -> Result<&KrunfwBindings, &String> {
+        self.0.as_ref()
     }
```
mz-pdm, line 159: Is it necessary to introduce this struct rather than simply using `Result<KrunfwBindings, String>` directly?  This is one of the things that could be put into the commit message for easier understanding of the change.

xhebox: Yes. Otherwise we need a whole manual Default derive.

mz-pdm: OK.

slp (approved):
This is useful and LGTM, thanks!

PR #782, opened 2026-07-14
tests: Improve test runner ergonomics

mtjhrc:
Test runner improvements:

- Fix argument quoting in run.sh (`$*` → `"$@"`) so arguments with spaces aren't split
- Support comma-separated glob patterns in `--test-case` with `!` negation (e.g. `TEST="net-*,!net-tap"`) to run all network tests except tap.
- Add `LIBKRUNFW_SRC=` to build and install libkrunfw from a source tree automatically into test-prefix, for iterating on kernel + libkrun changes together
- Remove the `"all"` special case value for testcase name. It is superseded by `"*"` pattern instead


slp:
LGTM, thanks!

slp (approved):

PR #783, opened 2026-07-14
include: fix const-correctness for `krun_add_net_tap`

simondeziel:
Align the C header signature of `krun_add_net_tap` with its underlying Rust implementation.

Fixes commit f8bdcc6d5d1774c80ae311969f2ec623717cfccb


mz-pdm (approved):
Thanks for the fix.

(I think we have some more misplaced `const`s in the file.)

mtjhrc (approved):
Thanks for the fix!

PR #785, opened 2026-07-16
display: accumulate damage regions across updates

bilelmoussaoui:
Track damage rectangles in a persistent Region that accumulates across multiple update() calls, rather than creating a new region from a single rectangle each time. The accumulated region is cleared after snapshotting.

This ensures MemoryTextureBuilder sees the full damaged area when multiple updates arrive between frames, preventing partial rendering artifacts.

Assisted-by: Claude Code: opus-4.6

cc @mtjhrc


mz-pdm (approved):
Not that I know how the updates work exactly, but the change looks plausible to me.

mtjhrc (approved):
LGTM thanks! 
I don't really have a good reproducer that makes use of damage areas to test this though.

PR #786, opened 2026-07-16
virtio/gpu: Allow setting display name / manufacturer name

bilelmoussaoui:
The Cuttlefish APEX `com.google.cf.input.config` ships IDC files that map touchscreen input to specific displays. The mapping uses a hash derived from the display's EDID manufacturer ID and name. The IDC files hardcode the hash for `GGL + CrosvmDisplay`, values crosvm normally produces.

Without this patch, libkrun's virtio-gpu generated a generic EDID. Android's GuestFrameComposer reads that EDID and computes a display ID hash that doesn't match the APEX IDC files, so touch input fails to route to the correct display.

The patch exposes `krun_display_set_edid_name()` so callers can set the manufacturer and display name in the generated EDID. Setting them to GGL/CrosvmDisplay produces the exact hash the APEX IDC files expect, making touch routing work correctly on both displays without modifying any Android code.

### DIFF
--- a/include/libkrun.h
+++ b/include/libkrun.h
@@ -510,6 +510,23 @@ int32_t krun_add_display(uint32_t ctx_id, uint32_t width, uint32_t height);
 int32_t krun_display_set_edid(uint32_t ctx_id, uint32_t display_id, const uint8_t* edid_blob, size_t blob_size);
 
 /**
+ * Set the EDID manufacturer and display name for a display.
+ *
+ * These values are used to generate the EDID reported to the guest.
+ * Either parameter may be NULL to leave that field unchanged.
+ *
+ * Arguments:
+ *  "ctx_id"        - the configuration context ID.
+ *  "display_id"    - the ID of the display (range: 0 to KRUN_MAX_DISPLAYS - 1)
+ *  "manufacturer"  - 3-character PNP manufacturer ID (e.g. "GGL"), or NULL
+ *  "display_name"  - display product name (max 13 characters), or NULL
+ *
+ * Returns:
+ *  Zero on success or a negative error number on failure.
+ */
+int32_t krun_display_set_edid_name(uint32_t ctx_id, uint32_t display_id, const char* manufacturer, const char* display_name);
+
+/**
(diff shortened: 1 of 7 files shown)


mz-pdm (commented):
Looks good to me overall, just some minor comments.

src/libkrun/src/lib.rs
```diff
@@ -1606,6 +1606,60 @@ pub unsafe extern "C" fn krun_display_set_edid(
...
+                return -libc::EINVAL;
+            }
+            edid_params.display_name = name_str.to_string();
+        }
+
+        KRUN_SUCCESS
+    })
+}
+
+#[cfg(not(feature = "gpu"))]
```
mz-pdm: How about `vhost-user` feature; honoured in the definition above?

include/libkrun.h
```diff
@@ -509,6 +509,23 @@ int32_t krun_add_display(uint32_t ctx_id, uint32_t width, uint32_t height);
...
+ * Set the EDID manufacturer and display name for a display.
+ *
+ * These values are used to generate the EDID reported to the guest.
+ * Either parameter may be NULL to leave that field unchanged.
+ *
+ * Arguments:
+ *  "ctx_id"        - the configuration context ID.
+ *  "display_id"    - the ID of the display (range: 0 to KRUN_MAX_DISPLAYS - 1)
+ *  "manufacturer"  - 3-character PNP manufacturer ID (e.g. "GGL"), or NULL
+ *  "display_name"  - display product name (max 12 characters), or NULL
```
mz-pdm: Max 12 or 13?

mtjhrc (changes requested):
LGTM, just address the review comments by Milan.

include/libkrun.h
bilelmoussaoui: indeed. Fixed.

bilelmoussaoui:
Should be good now, thanks

bilelmoussaoui:
Pushed some tests as well

mtjhrc (approved):
Looks good now, thanks!

PR #787, opened 2026-07-16
display/input: add ABS_MT_TRACKING_ID for multitouch

bilelmoussaoui:
Include ABS_MT_TRACKING_ID in the multitouch capability bitmap so the guest can properly track individual touch contacts across touch events.

Assisted-by: Claude Code: opus-4.6


mtjhrc (approved):
LGTM thanks! 

So does multi touch work correctly after this in your testing? The ABS_MT_TRACKING_ID were already emitted correctly and we were just missing the feature bit?

bilelmoussaoui:
Yes, without it touch input wasn't working on top of aaos

PR #788, opened 2026-07-16
init/aws-nitro: rewrite init binary in Rust

jakecorrenti:
Replaces the C implementation of the AWS Nitro Enclave init binary with a Rust rewrite (`init/aws-nitro/`). The new binary performs the same PID-1 responsibilities — bootstrapping the enclave guest from kernel module loading through application launch — but is easier to maintain and extend alongside the rest of the libkrun Rust codebase.

Also re-enables `krun_add_virtiofs*` for the `aws-nitro` feature and fixes the nitro example to set `/dev/stdout` as the console output.

This is mostly a 1:1 port. However, one notable thing I did change was how the proxies signaled that they were ready. I moved away from using signal handlers and decided to use pipes instead.


init/aws-nitro/src/archive/mod.rs
```diff
@@ -0,0 +1,38 @@
...
+        }
+
+        let mut data = Vec::new();
+        entry
+            .read_to_end(&mut data)
+            .context("unable to read entry data")?;
+        if !data.is_empty() {
+            super::nsm::pcr_extend_rootfs(nsm_fd, &data)
+                .context("unable to extend pcr with rootfs")?;
+        }
```
dorindabassey, line 30: This new Rust code collects the entire entry with read_to_end, but it chunks the data differently from the C version that used `nsm_rootfs_pcr_extend`. For instance a 5KB file where libarchive returns two blocks (3KB + 2KB), C code extends with [0..2K], [2K..3K], [3K..5K] while the Rust code extends with [0..2K], [2K..4K], [4K..5K].
Same number of extends, but the second and third chunks contain different data hence This produces different PCR16 values than the C implementation for entries larger than libarchive's internal block size, because the 2KB chunk boundaries shift when the entire entry is collected with `read_to_end` vs. measured block-by-block from `archive_read_data_block`. Is matching the C measurement sequence a requirement, or will expected PCR values be re-recorded for the Rust init?

init/aws-nitro/src/proxy/mod.rs
```diff
@@ -0,0 +1,528 @@
...
+            // To make this explicit, take shutdown_write by value (OwnedFd)
+            // and use drop() instead — but that prevents passing it to other
+            // proxies from the caller.
+            unistd::close(shutdown_write.as_raw_fd())
+                .context("unable to close shutdown_write pipe")?;
+
+            let addr = VsockAddr::new(VMADDR_CID_HOST, vsock_port);
+            let mut stream = VsockStream::connect(&addr).context("unable to connect to host")?;
+            stream.set_read_timeout(Some(Duration::from_secs(5)))?;
+            stream.set_write_timeout(Some(Duration::from_secs(5)))?;
```
dorindabassey, line 393: The C code set `SO_VM_SOCKETS_CONNECT_TIMEOUT (5s)` on all vsock sockets before calling `connect()`. But here, you only set timeouts after the connection is established so if the host is unresponsive, `connect()` hangs indefinitely.

Makefile
```diff
@@ -5,19 +5,6 @@ LIBRARY_HEADER_INPUT = include/libkrun_input.h
...
-        init/aws-nitro/args_reader.c			\
-        init/aws-nitro/fs.c				\
-        init/aws-nitro/mod.c					\
-		init/aws-nitro/device/include/*			\
-		init/aws-nitro/device/app_stdio_output.c	\
-		init/aws-nitro/device/device.c              \
-		init/aws-nitro/device/net_tap_afvsock.c	\
-		init/aws-nitro/device/signal.c		\
-
 AWS_NITRO_INIT_LD_FLAGS = -larchive -lnsm
```
dorindabassey, line 21: This was used by the old C build rule which is removed in this PR. Should be cleaned up.

dorindabassey (commented):
Thanks for the PR, Nice work. The last commit ("separate nitro init mods into their own files") moves code that was written in main.rs from earlier commits into separate modules. I think it might be cleaner to introduce the module structure from the start so each commit reflects the final layout, that way it's easier to review and bisect. Also, some of the later commits are missing descriptions.

init/aws-nitro/src/proxy/mod.rs
jakecorrenti: yep, good catch.

init/aws-nitro/src/archive/mod.rs
jakecorrenti: I had originally felt that it was ok to diverge from the original C implementation here, especially because we're moving to 2.0 but I'm going to ponder it some more and think about whether that's the right choice or not.

Do you have an opinion?

jakecorrenti:
still need to address @dorindabassey's comments, but I want to wait for a review/test from @tylerfanelli so I can address everything all at once

init/aws-nitro/src/archive/mod.rs
dorindabassey: I'd lean toward matching the C implementation's chunking behavior unless there's a better reason to diverge. Eventually it's up to you.

PR #789, opened 2026-07-16
Support finalized `virgl_renderer_resource_map_fixed` API

jakogut:
Upstream virglrenderer [MR 1374](https://gitlab.freedesktop.org/virgl/virglrenderer/-/merge_requests/1374) introduced a fixed blob mapping API that was enabled as an optional feature in libkrun along with a patched virglrenderer before the API stabilized.

Now that this is present in virglrenderer 1.3.0, let's update libkrun to use the final version of the API.

Additionally, we can detect at runtime when fixed resource mapping is supported by virglrenderer, and use fixed mappings automatically. This obviates the need for a build-time feature flag, while preserving compatibility with older virglrenderer builds.

Tested with virglrenderer 1.3.0 and podman/krun, using a custom build of `muvm-guest`, running vkmark. Additionally tested with virglrenderer 1.3.0 with `virgl_renderer_resource_map_fixed` patched out, to verify the fallback.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,9 +30,6 @@ ifeq ($(TDX),1)
     VARIANT = -tdx
     FEATURE_FLAGS := --features tdx
 endif
-ifeq ($(VIRGL_RESOURCE_MAP2),1)
-	FEATURE_FLAGS += --features virgl_resource_map2
-endif
 ifeq ($(INIT_BLOB),0)
     FEATURE_FLAGS += --no-default-features
 endif
--- a/README.md
+++ b/README.md
@@ -116,8 +116,7 @@ When TSI is enabled, the VMM acts as a proxy for AF_INET, AF_INET6 and AF_UNIX s
 
 #### Optional features
 
-* **GPU=1**: Enables virtio-gpu. Requires virglrenderer-devel.
-* **VIRGL_RESOURCE_MAP2=1**: Uses virgl_resource_map2 function. Requires a virglrenderer-devel patched with [1374](https://gitlab.freedesktop.org/virgl/virglrenderer/-/merge_requests/1374)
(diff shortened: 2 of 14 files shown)


slp:
Thanks for the PR! I was planning on doing this myself, so you actually saved me some work ;-)

One thing, for 2.0 we're planning to drop the vendored rutabaga (which supports map_fixed via map_placed) and use the one from crates.io instead. Could you please retarget this PR to the [stable-1.19.x](https://github.com/libkrun/libkrun/tree/stable-1.19.x) branch?

jakogut:
Of course, I'll retarget this to the 1.x branch. Great to hear 2.x is looking to remove the vendored rutabaga!

dorindabassey:
Just a note: that upstream rutabaga_gfx recently merged PR https://github.com/magma-gpu/rutabaga_gfx/pull/65 which fixes `map_placed` to properly call `virgl_renderer_resource_map_fixed`, the `dlsym` workaround in `mapping.rs` could be replaced when we consume the upstream rutabaga_gfx. So good to have this on the 1.x branch.

dorindabassey (approved):
LGTM

jakogut:
Just checking in, please let me know if there's anything else you'd like me to address before this can be merged. Thanks for reviewing!

slp:
@jakogut Do you see `map_fixed` getting used in your tests? I'm not getting any `OPAQUE_FD` handles in my tests.

jakogut:
@slp I see the issue, and I've pushed a fix.

Before, the branch only called `map_fixed` for `OPAQUE_FD`s. Export  failures fell through, reporting mapping success without actually mapping the BO.

I had been testing RADV native contexts using a branch based on this one that included an export failure fallback. The `map_fixed` calls I observed came from that fallback, not from `OPAQUE_FD` handles.

I've added the fallback here and tested it with the full vkmark suite using RADV native contexts. I also tested the pipewire/v4l2 camera capture path previously mentioned in a comment.

PR #790, opened 2026-07-17
virtio/gpu: reject out-of-range and wrapping blob map offsets

BinSquare:
This is only an issue with GPU passthrough enabled, prevents the panic! paths which crashed the VMM process + unintended host memory access.


jakecorrenti (approved):
LGTM! Thanks

slp:
LGTM too, thanks!

slp (approved):

PR #791, opened 2026-07-19
Upgrade to rand 0.10

benthetechguy:


jakecorrenti (changes requested):
You need to sign your commit with `git commit --amend -s` so the DCO check passes.

Also, can you elaborate on the reasoning why this change is necessary? Objectively I'm not opposed, just curious.

benthetechguy:
Right, I should have explained the rationale. I'm packaging libkrun for Debian, which requires all builds to be offline. This means that all of our Rust packages are locked into specific versions, and we currently only have rand 0.8 and 0.10 in the archive, so I figured it would be easiest to just bump libkrun to use 0.10.

jakecorrenti:
> Right, I should have explained the rationale. I'm packaging libkrun for Debian, which requires all builds to be offline. This means that all of our Rust packages are locked into specific versions, and we currently only have rand 0.8 and 0.10 in the archive, so I figured it would be easiest to just bump libkrun to use 0.10.

Thanks for the clarification. Running the CI now

jakecorrenti (approved):
LGTM!

PR #794, opened 2026-07-20
hvf: reclaim ballooned guest memory on macOS via stage-2 unmap and refault-remap

BinSquare:
Returns inflated virtio-balloon pages to the host on macOS by unmapping them from the HVF stage-2 tables and refaulting them back on next guest access, with MADV_FREE_REUSABLE for eager RSS reclaim.

This enables memory allocations of the guest to be reclaimed for host use on macs.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -709,6 +709,7 @@ version = "0.1.0-2.0.0-dev"
 dependencies = [
  "crossbeam-channel",
  "krun-arch",
+ "libc",
  "libloading",
  "log",
 ]
--- a/src/devices/src/virtio/balloon/device.rs
+++ b/src/devices/src/virtio/balloon/device.rs
@@ -95,14 +95,18 @@ impl Balloon {
                     "balloon: should release guest_addr={:?} host_addr={:p} len={}",
                     desc.addr, host_addr, desc.len
                 );
+                // macOS guest pages are hypervisor-pinned, so reclaim unmaps the
+                // range from stage-2 before the report is acked; it remaps lazily
+                // on the next guest fault (see hvf::balloon_reclaim_range).
+                #[cfg(target_os = "macos")]
+                if hvf::balloon_reclaim_enabled() {
+                    hvf::balloon_reclaim_range(desc.addr.0, host_addr as u64, desc.len as u64);
+                    continue;
+                }
+                // Linux MADV_DONTNEED decommits immediately; on macOS this path
+                // runs only when reclaim is disabled.
                 #[cfg(target_os = "linux")]
                 let advice = libc::MADV_DONTNEED;
-                // MADV_FREE_REUSABLE (not plain MADV_FREE) drops the pages from the
-                // process's phys_footprint immediately and lets the kernel reclaim
-                // them eagerly; MADV_FREE only reclaims under memory pressure, so
-                // idle VMs never shrink. Reported pages are unused by the guest
-                // until acked and come back zero-filled, matching reporting
-                // semantics.
                 #[cfg(target_os = "macos")]
                 let advice = libc::MADV_FREE_REUSABLE;
                 #[cfg(unix)]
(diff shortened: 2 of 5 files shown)


slp:
I'm not entirely convince we really need this. After https://github.com/libkrun/libkrun/pull/703 was merged, while the per-process numbers doesn't go down, the system's memory pressure really goes down, which is the thing that matters in the end.

BinSquare:
> I'm not entirely convince we really need this. After #703 was merged, while the per-process numbers doesn't go down, the system's memory pressure really goes down, which is the thing that matters in the end.

With only #703, the downside is that macOS Jetsam is unfairly prioritizing the libkrun process under memory pressure: https://github.com/jundot/omlx/issues/702#issuecomment-4441473434

Tracking the pages with the reusable bit is also generally adopted for correct resource tracking which to me seems harmless or helpful. Correct me if there's an obvious trade off I'm not considering though
- https://issues.chromium.org/issues/40567923
- https://groups.google.com/a/chromium.org/g/oilpan-reviews/c/Gg7c1vMKOSs

mtjhrc:
So this PR as it stands makes the feature opt-in via an env variable. I'd say that makes it pretty hard to discover for most users this even exists.
Having this is an API function would make it more discoverable and give us a place to document the behavior.

Do you know the downside(s) of using this? (performance?)

slp:
> > I'm not entirely convince we really need this. After #703 was merged, while the per-process numbers doesn't go down, the system's memory pressure really goes down, which is the thing that matters in the end.
> 
> With only #703, the downside is that macOS Jetsam is unfairly prioritizing the libkrun process under memory pressure: [jundot/omlx#702 (comment)](https://github.com/jundot/omlx/issues/702#issuecomment-4441473434)

Are you seeing Jetsam killing libkrun processes even with #703? Do you have a reproducer for that? If so, we should definitely adopt an strategy like this.

> So this PR as it stands makes the feature opt-in via an env variable. I'd say that makes it pretty hard to discover for most users this even exists. Having this is an API function would make it more discoverable and give us a place to document the behavior.

Should be a property of virtio-balloon, in the 2.x API.

> Do you know the downside(s) of using this? (performance?)

We can be sure there's going to be a significant performance impact, given that faults on pages will trigger full VM exits. That said, if Jetsam is killing libkrun, we could have something like this as an opt-in.

mtjhrc:
> Should be a property of virtio-balloon, in the 2.x API.

Yes definitely. It's fine if we merge this one with the env var and add the API later. (less work for me :smile: )
 
Not sure what we want to do for 1.x branch though.

BinSquare:
> > > I'm not entirely convince we really need this. After #703 was merged, while the per-process numbers doesn't go down, the system's memory pressure really goes down, which is the thing that matters in the end.
> > 
> > 
> > With only #703, the downside is that macOS Jetsam is unfairly prioritizing the libkrun process under memory pressure: [jundot/omlx#702 (comment)](https://github.com/jundot/omlx/issues/702#issuecomment-4441473434)
> 
> Are you seeing Jetsam killing libkrun processes even with #703? Do you have a reproducer for that? If so, we should definitely adopt an strategy like this.
> 
> > So this PR as it stands makes the feature opt-in via an env variable. I'd say that makes it pretty hard to discover for most users this even exists. Having this is an API function would make it more discoverable and give us a place to document the behavior.
> 
> Should be a property of virtio-balloon, in the 2.x API.
> 
> > Do you know the downside(s) of using this? (performance?)
> 
> We can be sure there's going to be a significant performance impact, given that faults on pages will trigger full VM exits. That said, if Jetsam is killing libkrun, we could have something like this as an opt-in.

Tried to repro, but was not able to do so in a clean environment: https://github.com/BinSquare/libkrun/tree/macos-balloon-reclaim-repro/repro/balloon-reclaim

1. I could only prove without the patch - the ranking for the process is always higher than the one with the patch in the kill order. https://github.com/BinSquare/libkrun/tree/macos-balloon-reclaim-repro/repro/balloon-reclaim
2. Performance impact of faults is on the order of milliseconds. 

Summary: 

Not able to get jetsam to kick in during the repro's - instead I encountered swapping and needed to manually kill processes. Performance hit did not appear to be significant but correct me on my experimentation setup + interpretation of results.

PR #799, opened 2026-07-28
CI: `cargo clippy --features aws-nitro,net` + fix the error

mtjhrc:
Not sure if it is broken in other ways, but this should make it at least compile + we check for it in the CI.
@jakecorrenti


mz-pdm (approved):

jakecorrenti (approved):
code lgtm, i also made similar changes in https://github.com/libkrun/libkrun/pull/788

PR #803, opened 2026-07-28
linux, gentoo, asahi, guest write to a dma-buf blob mapping kills the VM with EFAULT, prefault the mapping in the VMM

ccharon:
I don´t know if this actually is a proper fix, this pr is at least a reminder that there is a problem (tested with krun 1.18) on asahi with kernel 7.1.5 which lets muvm crash when using this.

libkrun maps a blob resource exported as a dma-buf into the virtio-gpu shm region with mmap(MAP_FIXED). That VMA is VM_PFNMAP,  no struct page, not ordinary managed memory. 

When the guest writes to a page that has no writable PTE yet, KVM resolves the fault through hva_to_pfn_remapped(), finds no writable PTE, and returns KVM_PFN_ERR_RO_FAULT; user_mem_abort() fails with -EFAULT and KVM_RUN tears the VM down (Failure during vcpu run: Bad address). KVM cannot fault such memory in on its own, populating it is the VMM's job, and libkrun doesn't do it. Under muvm on Apple Silicon that means every GPU accelerated workload dies the moment the guest touches the GPU. memfd backed blobs have ordinary VMA flags and are unaffected.

The patch walks the fresh mapping once and rewrites every page with the value it already holds: contents unchanged, but a writable PTE exists before the guest ever reaches the page. It applies only to RUTABAGA_MEM_HANDLE_TYPE_DMABUF exports (memfd blobs keep lazy faulting) and only to PROT_READ|PROT_WRITE mappings, where the read modify write is safe. MAP_POPULATE is not an option: populate_vma_page_range() goes through __get_user_pages(), which rejects VM_PFNMAP in check_vma_flags(), and mm_populate() swallows the error. Cost: every page of a dma-buf blob is faulted in at map time rather than on demand.

### DIFF
--- a/src/devices/src/virtio/gpu/virtio_gpu.rs
+++ b/src/devices/src/virtio/gpu/virtio_gpu.rs
@@ -20,7 +20,7 @@ use krun_display::{
 use libc::c_void;
 #[cfg(target_os = "macos")]
 use rutabaga_gfx::RUTABAGA_MEM_HANDLE_TYPE_APPLE;
-#[cfg(all(feature = "virgl_resource_map2", target_os = "linux"))]
+#[cfg(target_os = "linux")]
 use rutabaga_gfx::RUTABAGA_MEM_HANDLE_TYPE_DMABUF;
 #[cfg(all(not(feature = "virgl_resource_map2"), target_os = "linux"))]
 use rutabaga_gfx::RUTABAGA_MEM_HANDLE_TYPE_OPAQUE_FD;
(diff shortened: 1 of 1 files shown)


slp:
> KVM cannot fault such memory in on its own, populating it is the VMM's job, and libkrun doesn't do it. Under muvm on Apple Silicon that means every GPU accelerated workload dies the moment the guest touches the GPU. memfd backed blobs have ordinary VMA flags and are unaffected.

I'm able to reproduce the issue, but I disagree with this root cause analysis. Userspace (the VMM) shouldn't need to pre-fault the region, that has never been the case before with older kernels, and that breaks the on-demand paging semantics, implying a significant performance impact.

All signs point to a kernel issue. I'm currently investigating when it got introduced and why.

slp:
I think this (or something similar to this) is the right solution: https://lore.kernel.org/kvm/20260729014223.17937-1-slp@redhat.com/T/#u

jannau:
For Fedora-Asahi-Remix Kernel 7.1.6-400.asahi is available with kernel patches fixing the regression

PR #805, opened 2026-07-30
tests: Replace deprecated tempdir with successor tempfile

benthetechguy:
It also looks like the crate is completely unused in tests/test_cases.

### DIFF
--- a/tests/runner/src/main.rs
+++ b/tests/runner/src/main.rs
@@ -7,7 +7,7 @@ use std::io::Write;
 use std::panic::catch_unwind;
 use std::path::{Path, PathBuf};
 use std::process::{Command, Stdio};
-use tempdir::TempDir;
+use tempfile::TempDir;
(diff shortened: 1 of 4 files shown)


tests/runner/src/main.rs
```diff
@@ -349,9 +349,9 @@ fn run_tests(
             fs::create_dir_all(&path).context("Failed to create base directory")?;
             path
         }
-        None => TempDir::new_in("/tmp", "libkrun-tests")
+        None => TempDir::with_prefix_in("/tmp", "libkrun-tests")
```
dorindabassey: In the new API this arg is flipped. Should be: `TempDir::with_prefix_in("libkrun-tests", "/tmp")`

dorindabassey (commented):
Minor comment, Otherwise LGTM

tests/runner/src/main.rs
benthetechguy: Good catch, fixed!

dorindabassey (dismissed):

slp (approved):

PR #806, opened 2026-08-04
virtio/vsock: don't make the host wait for the reaper on a refused connection

### LINKED ISSUE #684: First host-to-guest vsock connection blocks for ~5 seconds before EOF
I hit what appeared to be "slow boot" when testing with libkrun and libkrunfw and a in-guest agent that was communicating over vsock.

I put a reproducer together and made a draft pr for it https://github.com/containers/libkrun/pull/683 . The PR is just as a placeholder to hold the reproducer. Claude did aid in this development, but please don't discount this as AI slop.  If I'm using it wrong, please feel free to let me know.

When a host process `connect()`s to the unix socket exposed by `krun_add_vsock_port2(..., listen=true)`, the very first attempt after VM boot behaves as follows:

- `connect()` returns immediately (libkrun's `UnixAcceptorProxy` accepts on the host)
- `read(1)` blocks for ~5 seconds
- `read()` returns 0 (EOF) — no byte was ever delivered from the guest
- a *second* host connection, opened immediately afterwards, completes a full round trip in well under 1 ms

A caller that probes for guest readiness by retrying on EOF sees its total cold-start time dominated by this 5 s block.

## Reproducer
A self-contained C reproducer lives in #683 at `examples/vsock-latency/`. The guest binary is a tiny statically linked program that binds an AF_VSOCK listener, accepts a connection, and writes one ack byte. The host program configures the VM, spawns a timing thread, and calls
`krun_start_enter`.

```
cd examples/vsock-latency
make
make demo
```

Sample output (libkrunfw 5.4.0 / bundled kernel 6.12.87 / host kernel 6.18):

```
=== PROBE_TIMEOUT_MS=unset  (cold-boot, ~5s)... ===
  attempt[1] +5008.00 ms  read=EOF
  attempt[2]   +0.39 ms  read=ack
phase: first-roundtrip-ok +5330.88 ms  (delta-from-socket=5018.48 ms, attempts=2)

=== PROBE_TIMEOUT_MS=2000  (cold-boot, ~2.5s)... ===
  attempt[1] +2029.04 ms  read=timeout
  attempt[2]   +0.29 ms  read=ack
phase: first-roundtrip-ok +2351.62 ms  (delta-from-socket=2039.41 ms, attempts=2)

=== PROBE_TIMEOUT_MS=100  (cold-boot, ~0.5s)... ===
  attempt[1] +101.94 ms  read=timeout
  attempt[2]   +0.30 ms  read=ack
phase: first-roundtrip-ok + 424.91 ms  (delta-from-socket=112.34 ms, attempts=2)
```

The diagnostic line is `attempt[1] +5008 ms read=EOF`: a host `connect()` + `read(1)` round trip that should take milliseconds takes 5 seconds before libkrun closes the socket with EOF.

## Trace through the code

Following the host-side path on the doomed first attempt:

1. Host `connect()` → libkrun's `UnixAcceptorProxy::process_event`  accepts the unix connection  (`src/devices/src/virtio/vsock/unix.rs:687-714`).
2. libkrun creates a new proxy for the connection and tries to complete the vsock handshake to the guest's AF_VSOCK listener.
3. The vsock leg of that first attempt does not complete and the proxy gets marked `ProxyRemoval::Deferred` (`src/devices/src/virtio/vsock/muxer_thread.rs:99-104`, visible as `deferring proxy removal: <id>` WARN messages).
4. The reaper thread holds the proxy in `released_map` for `TIMEOUT = Duration::new(5, 0)` before removing it (`src/devices/src/virtio/vsock/reaper.rs:10,32-38`).
5. While the proxy is in `released_map`, the host-side unix socket FD stays bound to it. The host's `read()` therefore blocks until the reaper finally removes the proxy and closes the unix FD, yielding EOF.

Subsequent host connections create a fresh proxy and complete normally in sub-millisecond time.

## Environment
- libkrun: 1.18.0 (built from source)
- libkrunfw: 5.4.0
- libkrunfw bundled kernel: 6.12.87
- Host kernel: 6.18.x
- Architecture: x86_64

smoser:
The first host-to-guest vsock connection after VM start blocks for ~5 seconds
before failing, which makes any caller that probes for guest readiness over
vsock see a cold start dominated by that one block.

Fixes #684. A standalone C reproducer is attached to that issue (#683).

## What happens

A host process connects to a unix socket registered with
`krun_add_vsock_port2(..., listen=true)`. libkrun accepts it at the unix layer
immediately -- before the guest leg of the connection exists -- and sends
OP_REQUEST to the guest. If nothing in the guest listens on that port yet, the
guest answers OP_RST about 3 ms later:

.393744 unix]  new_reverse: id=5301052505799 local_port=1062862535 peer_port=1234
.396743 muxer] OP_RST
.396752 unix]  release: id=..., tx_cnt=0
.396756 muxer] deferring proxy removal: ...
   attempt[1] +5003.62 ms  read=EOF

So the refusal is known in 3 ms and delivered in 5 s. `UnixProxy::release()`
returned `ProxyRemoval::Deferred`, which keeps the proxy -- and the accepted
host socket it owns -- alive until the reaper reclaims it after its 5 s TTL.
The host's `read()` blocks until the fd is finally closed.

Deferred removal reserves a connection id the way TIME_WAIT does. That only
makes sense once the connection existed; a proxy that never reached Connected
has nothing in flight and is not yet registered with the muxer's epoll (that
happens later, in `update_peer_credit()`), so it can simply be dropped.

## The two commits

Returning `ProxyRemoval::Immediate` is the obvious fix, and on its own it
**deadlocks the muxer thread**: `process_stream_rst()` holds a `proxy_map` read
guard across `process_proxy_update()`, which takes the write lock on the same
`RwLock`. The trace log stops dead at "immediately removing proxy" and no
further vsock connection on that VM succeeds. The sibling
`process_proxy_release()` already scopes its guard to avoid this, so the first
commit makes `process_stream_rst()` match it. That is latent today -- nothing
reaches `Immediate` from this path until the second commit -- but
`TsiStreamProxy::release()` already returns `Immediate` when Listening, so it
is worth fixing regardless.

The second commit makes `UnixProxy::release()` state-aware: `Immediate` for
`ReverseInit`/`Connecting`, `Deferred` unchanged for connections that reached
Connected.

## Measurements

libkrunfw 5.5.0 (bundled kernel 6.12.87), host kernel 6.18, x86_64. Time to the
first successful round trip after VM start, by the caller's per-probe timeout:

| per-probe timeout | before | after |
| --- | --- | --- |
| none (block to EOF) | 5338 ms | 415 ms |
| 2000 ms | 2335 ms | 398 ms |
| 100 ms | 435 ms | 375 ms |
| 10 ms | 411 ms | 379 ms |

Cold start no longer depends on the caller's probe deadline -- it is now
dominated by guest boot (the unix socket appears at ~334 ms). Warm round trips
are unchanged at ~0.17 ms.

A retrying caller also stops accumulating proxies: 200 back-to-back refused
connections retained 200 file descriptors for the reaper window before this
change, and none after it.

## Testing

New integration test `vsock-host-connect-refused`. Nothing in the suite
exercised `krun_add_vsock_port2(..., listen=true)` before this. Since the
harness has no guest-ready signal, the test derives one from a second port: the
guest connects out on 1234, and the host's `accept()` returning proves the
guest's vsock stack is up and will answer a request for an unbound port.
Probing port 1235, which the guest never binds, is then independent of boot
timing. The host asserts the connection is refused (a real EOF, not a timeout)
within 1 s; measured 0 ms fixed, 5000 ms unfixed, so the threshold has wide
margin either way.

- Full suite passes at the branch tip: 11/11, 18 skipped (freebsd assets,
  `IPERF_DURATION`, `PJDFSTEST_REPO` unset).
- Each commit builds, and vsock + TSI tests pass with the first commit alone.
- Negative control: the new test fails with "took 5000ms to be refused, limit
  is 1000ms" against a library built from the first commit only.
- `cargo fmt --check` clean.

Not verified, flagging honestly:
- Linux/KVM x86_64 only; `unix.rs` and `muxer.rs` are shared with macOS/HVF.

### DIFF
--- a/src/devices/src/virtio/vsock/muxer.rs
+++ b/src/devices/src/virtio/vsock/muxer.rs
@@ -694,18 +694,21 @@ impl VsockMuxer {
     fn process_stream_rst(&self, pkt: &VsockPacket) {
         debug!("OP_RST");
         let id: u64 = ((pkt.src_port() as u64) << 32) | (pkt.dst_port() as u64);
-        if let Some(proxy_lock) = self.proxy_map.read().unwrap().get(&id) {
+        let update = if let Some(proxy_lock) = self.proxy_map.read().unwrap().get(&id) {
             debug!(
                 "allowing OP_RST: id={} src={} dst={}",
                 id,
                 pkt.src_port(),
                 pkt.dst_port()
             );
-            let mut proxy = proxy_lock.lock().unwrap();
-            let update = proxy.release();
-            self.process_proxy_update(id, update);
+            Some(proxy_lock.lock().unwrap().release())
         } else {
             debug!("invalid OP_RST for {}", pkt.src_port());
+            None
+        };
+
+        if let Some(update) = update {
+            self.process_proxy_update(id, update);
         }
     }
 
--- a/src/devices/src/virtio/vsock/unix_proxy/unix.rs
+++ b/src/devices/src/virtio/vsock/unix_proxy/unix.rs
@@ -327,7 +327,15 @@ pub(crate) fn release(proxy: &mut super::UnixProxy) -> ProxyUpdate {
         "release: id={}, tx_cnt={}, last_tx_cnt={}",
         proxy.id, proxy.tx_cnt, proxy.last_tx_cnt_sent
     );
-    let remove_proxy = ProxyRemoval::Deferred;
+
+    // A connection that never reached Connected carries no data and is not
+    // registered for polling yet, so there is nothing to reserve its id for.
+    // Keeping it would leave the accepted host socket open, and its peer
+    // blocked, until the reaper reclaims the proxy.
+    let remove_proxy = match proxy.status {
+        ProxyStatus::ReverseInit | ProxyStatus::Connecting => ProxyRemoval::Immediate,
+        _ => ProxyRemoval::Deferred,
+    };
 
     ProxyUpdate {
         remove_proxy,
--- a/tests/test_cases/src/lib.rs
+++ b/tests/test_cases/src/lib.rs
@@ -12,6 +12,11 @@ mod test_vsock_guest_connect;
 #[cfg(any(feature = "host", target_os = "linux"))]
 use test_vsock_guest_connect::TestVsockGuestConnect;
 
+#[cfg(any(feature = "host", target_os = "linux"))]
+mod test_vsock_host_connect_refused;
+#[cfg(any(feature = "host", target_os = "linux"))]
+use test_vsock_host_connect_refused::TestVsockHostConnectRefused;
+
 mod test_tsi_tcp_guest_connect;
 use test_tsi_tcp_guest_connect::TestTsiTcpGuestConnect;
 
@@ -101,6 +106,11 @@ pub fn test_cases() -> Vec<TestCase> {
         TestCase::new("vm-pause", Box::new(TestVmPause)),
         #[cfg(any(feature = "host", target_os = "linux"))]
         TestCase::new("vsock-guest-connect", Box::new(TestVsockGuestConnect)),
+        #[cfg(any(feature = "host", target_os = "linux"))]
+        TestCase::new(
+            "vsock-host-connect-refused",
+            Box::new(TestVsockHostConnectRefused),
+        ),
         TestCase::new(
             "tsi-tcp-guest-connect",
             Box::new(TestTsiTcpGuestConnect::new()),
--- a/tests/test_cases/src/test_vsock_host_connect_refused.rs
+++ b/tests/test_cases/src/test_vsock_host_connect_refused.rs
@@ -1,0 +1,196 @@
+#![cfg(any(feature = "host", target_os = "linux"))]
+
+use macros::{guest, host};
+
+pub struct TestVsockHostConnectRefused;
+
+/// The guest connects out on this port, which tells the host that the guest's
+/// vsock stack is up and a request to an unbound port will be answered.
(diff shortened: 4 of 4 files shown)


smoser:
@slp, friendly ping. I rebased on main today.  This narrows the ProxyRemoval::Deferred default you introduced in 328d8d7740d8d9fdefae283c4a1ccaf38552d204. A proxy that never reached `Connected` holds no data and isn't registered for polling, so deferring its removal just keeps the accepted host socket (and its blocked peer) alive until the reaper runs.   We've hit this in 2 different ways in our use of libkrun.

There is a reproducer in #683.

Please let me know if there is something else I need to take care of.
Thanks.

lstocchi (commented):
LGTM, tested on Windows by porting your example and i was able to replicate the issue + the fix. Thanks! Going to add it to the Windows part on https://github.com/libkrun/libkrun/pull/779 as well 👍

slp (dismissed):
LGTM, thanks a lot @smoser !

slp:
Hm... looks like we may need `#![cfg(feature = "host")]` on the constants that are only used by the host-side of the test. Is that correct, @mtjhrc ?

smoser:
> Hm... looks like we may need `#![cfg(feature = "host")]` on the constants that are only used by the host-side of the test. Is that correct, @mtjhrc ?

I rebased to address the merge conflicts and addressed the c-i issue above.

slp (approved):
@smoser thanks!

PR #808, opened 2026-08-06
virtio/console: validate guest control messages

dills122:
## Summary

Validate virtio-console driver-to-device control messages before processing them.

The control transmit path previously read a control structure directly from the head descriptor address without validating the complete descriptor layout. It also allowed guest-provided `PORT_READY` and `PORT_OPEN` IDs to reach port and derived-queue lookups without first checking that the port existed.

## Changes

- Reject unknown `PORT_READY` and `PORT_OPEN` port IDs before accessing ports or derived queues.
- Accept an exact 8-byte readable control message in either one contiguous descriptor or a split scatter/gather chain.
- Reject short or trailing payloads, writable or mixed-direction descriptors, unsupported flags, loops, malformed links, oversized lengths, and invalid guest-memory ranges.
- Complete every successfully popped driver-to-device control chain exactly once with a used length of 0, since the device writes no response bytes into that chain.
- Add queue-path unit tests for valid contiguous and split messages, malformed chains, invalid memory, completion behavior, and unknown port IDs.

The change is internal to virtio-console control handling and does not alter the public API or ABI.

## Validation

- `cargo fmt -- --check` in the root, examples, and tests workspaces
- `cargo test --locked -p krun-devices` at each logical commit: 48/48 and 51/51 unit tests passed; doc-tests passed
- `cargo test --locked -p krun-devices virtio::console::device::tests`: 4/4 focused tests passed
- `cargo clippy --locked -p krun-devices -- -D warnings`
- Prior macOS arm64/HVF multiport-console guest regression testing passed twice against identical console, queue, and descriptor implementation bytes

### DIFF
--- a/src/devices/src/virtio/console/device.rs
+++ b/src/devices/src/virtio/console/device.rs
@@ -1,7 +1,7 @@
 use std::cmp;
 use std::io::Write;
 use std::iter::zip;
-use std::mem::{size_of, size_of_val};
+use std::mem::size_of;
 #[cfg(unix)]
 use std::os::unix::io::{AsRawFd, RawFd};
 use std::sync::Arc;
@@ -188,10 +188,7 @@ impl Console {
                     continue;
                 }
             };
-            if let Err(e) = control_tx
-                .queue
-                .add_used(mem, head.index, size_of_val(&cmd) as u32)
-            {
+            if let Err(e) = control_tx.queue.add_used(mem, head.index, 0) {
                 error!("failed to add used elements to the queue: {e:?}");
             }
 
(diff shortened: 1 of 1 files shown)


slp:
@mtjhrc PTAL

src/devices/src/virtio/console/device.rs
```diff
@@ -23,11 +23,78 @@ use crate::virtio::console::port::Port;
...
+            }
+        };
+
+        let has_next = current.flags & VIRTQ_DESC_F_NEXT != 0;
+        descriptor = current.next_descriptor();
+        if has_next && descriptor.is_none() {
+            log::error!("Malformed console control descriptor chain");
+            return None;
+        }
+    }
```
mtjhrc, line 72: So this here walks the descriptor chain manually, but the new code also uses the Reader utility, which is meant to abstract this away. So we should either do the full manual implementation without using Reader, or use Reader — using Reader is preferable, hence I think we should drop this defensive loop.

So I would really like to remove this manual loop. If there is any check that Reader doesn't perform it should be fixed there rather than doing 2 separate walks of the same descriptor chain here.

src/devices/src/virtio/console/device.rs
```diff
@@ -23,11 +23,78 @@ use crate::virtio::console::port::Port;
...
+
+    if readable_len != size_of::<VirtioConsoleControl>() {
+        log::error!(
+            "Invalid console control payload length: expected {}, got {readable_len}",
+            size_of::<VirtioConsoleControl>()
+        );
+        return None;
+    }
```
mtjhrc, line 80: If you drop the manual loop (like my other comment suggests), you can still implement this check by checking if `reader.available_bytes() == 0` after doing `reader.read_obj()`, and reporting an error (or only a warning?) in such case.

mtjhrc (changes requested):

PR #811, opened 2026-08-10
vhost-user: add media device support

dorindabassey:
### Summary

- Upgrade vhost crate from 0.15 to 0.17 for SHMEM and BACKEND_REQ protocol support
- Add shared memory region allocation for vhost-user devices and implement the BACKEND_REQ channel for handling shmem_map/shmem_unmap requests from backends
- Expose vhost-user media device type (VIRTIO_ID_MEDIA = 48) with 4 GiB SHMEM region

### How to test

For the backend run:
RUST_LOG=debug ./target/debug/vhost-device-media \
--socket-path /tmp/vhost-user-media.sock \
--v4l2-device /dev/video0 --backend v4l2-proxy

In the guest:
LD_LIBRARY_PATH=./target/release ./examples/chroot_vm \
--vhost-user-media=/tmp/vhost-user-media.sock / /bin/bash

To Capture:
v4l2-ctl -d /dev/video0 --set-fmt-video=pixelformat=MJPG \
--stream-mmap --stream-count=60 --stream-to=/tmp/video.mjpg

Test Playback (on host):
ffplay -f mjpeg /tmp/video.mjpg

### DIFF
--- a/src/libkrun/src/lib.rs
+++ b/src/libkrun/src/lib.rs
@@ -86,7 +86,6 @@ const MAX_ARGS: usize = 4096;
 /// Maximum number of virtqueues allowed by virtio spec (16-bit queue index: 0-65535)
 #[cfg(feature = "vhost-user")]
 const VIRTIO_MAX_QUEUES: usize = 65536;
-
 // krunfw library name for each context
 #[cfg(all(target_os = "linux", not(feature = "tee")))]
 const KRUNFW_NAME: &str = "libkrunfw.so.5";
@@ -1738,6 +1737,7 @@ pub unsafe extern "C" fn krun_add_vhost_user_device(
     name: *const c_char,
     num_queues: u16,
     queue_sizes: *const u16,
+    shm_size: u64,
 ) -> i32 {
     use vmm::resources::VhostUserDeviceConfig;
 
@@ -1786,12 +1786,19 @@ pub unsafe extern "C" fn krun_add_vhost_user_device(
     match CTX_MAP.lock().unwrap().entry(ctx_id) {
         Entry::Occupied(mut ctx_cfg) => {
             let cfg = ctx_cfg.get_mut();
+            let shm_size = if shm_size > 0 {
+                Some(shm_size as usize)
+            } else {
+                None
+            };
+
             cfg.vmr.vhost_user_devices.push(VhostUserDeviceConfig {
                 device_type,
                 socket_path: socket_path_str.to_string(),
                 name: name_opt,
                 num_queues,
                 queue_sizes: queue_sizes_vec,
+                shm_size,
             });
             KRUN_SUCCESS
         }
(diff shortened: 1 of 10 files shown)


dorindabassey:
Marking as draft because it needs the media guest driver counterpart in libkrunfw(WIP).

dorindabassey:
the driver part - https://github.com/libkrun/libkrunfw/pull/139

src/libkrun/src/lib.rs
```diff
@@ -1731,12 +1733,19 @@ pub unsafe extern "C" fn krun_add_vhost_user_device(
     match CTX_MAP.lock().unwrap().entry(ctx_id) {
         Entry::Occupied(mut ctx_cfg) => {
             let cfg = ctx_cfg.get_mut();
+            let shm_size = match device_type {
+                VIRTIO_ID_MEDIA => Some(1 << 32),
+                VIRTIO_ID_GPU => Some(1 << 33),
```
valpackett: hm, where do these hardcoded sizes come from?

dorindabassey: The virtio spec doesn't define SHM region sizes - it only defines the region IDs, so the sizes are the VMM's decision. The driver just sees whatever region the device advertises. 
For the GPU `1 << 33 (8 GiB)` - this matches the existing built-in virtio-gpu default in https://github.com/libkrun/libkrun/blob/main/src/vmm/src/builder.rs#L1671 which is large enough to hold host-visible GPU resources (textures, buffers) for typical workloads. For the Media `1 << 32 (4 GiB)` seems sufficient for V4L2 MMAP buffers.
I guess we can make these configurable by adding an optional `shm_size` parameter to the existing `krun_add_vhost_user_device()` like the `krun_set_gpu_options2` does for the built-in virtio-gpu.

dorindabassey: Hi @valpackett , I considered adding the configurable shm_size parameter but these sizes are device-specific implementation details that users shouldn't need to know about, so I replaced the hardcoded values with named constants, felt like the right approach.

mtjhrc: > these sizes are device-specific implementation details that users shouldn't need to know about

I disagree, I think it is better to make the vhost user backend be as generic as possible, allowing the user to use any vhost user device without us having to explicitly add support for it. 

> I guess we can make these configurable by adding an optional shm_size parameter to the existing krun_add_vhost_user_device() like the krun_set_gpu_options2 does for the built-in virtio-gpu.

Yeah I think this is the way. I think we shouldn't the implicit value like this here at all. 

What we could do however is `#define` the recommend shm sizes in the header though, like we do for queue sizes.

dorindabassey: > What we could do however is #define the recommend shm sizes in the header though, like we do for queue sizes.

Yeah I think this make sense, fixed thanks!

slp (approved):
LGTM, thanks!

PR #815, opened 2026-08-17
display/edid: limit standard timings to configured resolution

dorindabassey:
## Summary
Filter EDID standard timing entries to only include resolutions that fit within the display dimensions, preventing the guest from picking an oversized mode that causes stride mismatches.
Found this issue while testing vhost-user-gpu display output: glmark2 was picking 1920x1200 from the EDID standard timings instead of the configured 1024x768, causing garbled output from the stride mismatch.


mtjhrc (approved):
LGTM, thanks!

PR #816, opened 2026-08-18
Add vhost user gpu scanout

dorindabassey:
### Summary

- Wire the display backend into vhost-user GPU devices to handle software scanout messages (SCANOUT type 7, UPDATE type 8)
- Enable display output when using vhost-device-gpu with the gfxstream backend, which sends raw pixel data over the GPU socket
- Add --vhost-user-gpu option to gui_vm example
- Depends on #815 (EDID fix for standard timings)

The gfxstream backend sends pixel data directly in the UPDATE message payload (software scanout). The
virglrenderer backend instead sends DMABUF file descriptors via SCM_RIGHTS (DMABUF scanout, message
types 9/10/12) - that path is not yet supported and depends on #465  so it'll be a follow-up PR.

The guest falls back to llvmpipe for rendering. Display output works because the gfxstream backend reads 2D resource pixels and sends them over the socket via software scanout.

## Testing

Start the vhost-device-gpu backend with gfxstream:

vhost-device-gpu --socket-path /tmp/gpu.socket --gpu-mode gfxstream

Start gui_vm with vhost-user GPU:

./examples/target/release/gui_vm \
--root-dir /path/to/rootfs \
--display 1024x768 \
--vhost-user-gpu /tmp/gpu.socket \
/bin/bash

In the guest, set the DRM mode before running graphical apps (the CRTC is not configured by default):

modetest -M virtio_gpu -s <connector_id>:<mode>
Find the connector ID and available modes with modetest -M virtio_gpu.
E.g: modetest -M virtio_gpu -s 37:1024x768
Then run benchmarks:

* Spinning cube (~310 FPS with llvmpipe)
kmscube

* glmark2 (score ~128)
glmark2-drm --visual-config id=37

### DIFF
--- a/src/devices/src/virtio/vhost_user/device.rs
+++ b/src/devices/src/virtio/vhost_user/device.rs
@@ -11,14 +11,17 @@ use std::os::fd::{AsRawFd, FromRawFd, RawFd};
 use std::os::unix::net::UnixStream;
 use std::sync::{Arc, Mutex};
 
+use krun_display::{
+    DisplayBackend, DisplayBackendBasicFramebuffer, DisplayBackendInstance, ResourceFormat,
+};
 use log::{debug, error, warn};
 use nix::sys::socket::{ControlMessage, MsgFlags, sendmsg};
 use polly::event_manager::{EventManager, Subscriber};
 use utils::epoll::{EpollEvent, EventSet};
 use utils::eventfd::{EFD_NONBLOCK, EventFd};
 use vhost::vhost_user::gpu_message::{
-    GpuBackendReq, VhostUserGpuHeaderFlag, VirtioGpuDisplayOne, VirtioGpuRect,
-    VirtioGpuRespDisplayInfo, VirtioGpuRespGetEdid,
+    GpuBackendReq, VhostUserGpuHeaderFlag, VhostUserGpuScanout, VhostUserGpuUpdate,
+    VirtioGpuDisplayOne, VirtioGpuRect, VirtioGpuRespDisplayInfo, VirtioGpuRespGetEdid,
 };
 use vhost::vhost_user::message::{
     FrontendReq, VhostUserConfigFlags, VhostUserMMap, VhostUserMMapFlags,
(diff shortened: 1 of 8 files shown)


slp:
LGTM, but I'd like another review from @mtjhrc

mtjhrc:
FYI I rebased this by clicking the rebase button on Github, but that didn't work out well (the problem was the vmm crate was moved but it recreated it). I'm fixing it manually.

src/devices/src/virtio/vhost_user/device.rs
```diff
@@ -858,6 +901,164 @@ impl VhostUserDevice {
...
+
+        if let Some(display) = self.get_display_instance()
+            && let Err(e) = display.configure_scanout(
+                scanout_id,
+                display_width,
+                display_height,
+                scanout.width,
+                scanout.height,
+                ResourceFormat::XRGB,
+            )
```
mtjhrc:
```suggestion
                ResourceFormat::BGRX,
            )
```

On my machine the colors are shifted. Changing this fixes it.

Before (with `ResourceFormat::XRGB`):
Image


After (with `ResourceFormat::BGRX`):
Image

----

### Why this happens
> Generated-By:  Gemini 3.8 Flash

The [vhost-user-gpu spec](https://www.qemu.org/docs/master/interop/vhost-user-gpu.html#vhostusergpuupdate) defines the update payload format as `PIXMAN_x8r8g8b8`.

The confusion comes from two different ways libraries name pixel formats:
- **Pixman** names formats by the bits in a 32-bit integer: `0xXXRRGGBB` (hence `x8r8g8b8`).
- **libkrun** (like virglrenderer and GTK/GDK) names formats by how those bytes actually sit in RAM. Because little-endian CPUs store the lowest byte first, `0xXXRRGGBB` is stored in memory as `[Blue, Green, Red, X]`.

So Pixman's `x8r8g8b8` payload is physically stored as `[B, G, R, X]`, which in libkrun corresponds to `ResourceFormat::BGRX`.

dorindabassey: I didn't notice that, Thanks!

mtjhrc:
FYI you force pushed the bad automatic rebase - you need to manually remove the `‎src/vmm/Cargo.toml`

dorindabassey:
> FYI you force pushed the bad automatic rebase - you need to manually remove the `‎src/vmm/Cargo.toml`

Fixed, thanks!

mtjhrc (approved):
It would be nice to unhardcode this a bit to be able to use multiple displays but this LGTM as initial support. Thanks!

PR #821, opened 2026-08-25
arch/x86_64: keep initrd below the MMIO gap

ben-gordon-dev:
## Problem

The x86 Linux direct-boot path writes the initrd address to the legacy
32-bit `ramdisk_image` field without populating `ext_ramdisk_image`.

When guest RAM crosses the MMIO gap, calculating the initrd address
from `ram_last_addr` can place it above 4 GiB. Converting that address
to `u32` discards its upper bits, causing the kernel to read the initrd
from the wrong address and fail to find `/init`.

## Fix

Keep the initrd below the x86 MMIO gap:

```rust
initrd_addr: ram_last_addr.min(MMIO_MEM_START) - initrd_size,
```

This leaves placement unchanged when RAM ends below the gap. When RAM
crosses the gap, it places the initrd at the end of the existing
low-memory RAM region.

The regression covers 3330 MiB, immediately above the MMIO boundary,
and 4096 MiB. It verifies the exact initrd address and that the complete
image is contained within one guest RAM region.

## Testing

- `cargo fmt -- --check`
- `cargo test`
- Verified that the regression fails on the parent commit and passes
  with the fix
- Validated placement at 3330, 4096, 5120, and 8192 MiB on an AWS Linux
  host with nested KVM
- Ubuntu 24.04.4 completed initramfs → rootfs → systemd multi-user boot
  under nested KVM at 5120 MiB and shut down cleanly

### DIFF
--- a/src/arch/src/x86_64/mod.rs
+++ b/src/arch/src/x86_64/mod.rs
@@ -196,7 +196,7 @@ pub fn arch_memory_regions(
         ram_last_addr,
         shm_start_addr,
         page_size,
-        initrd_addr: ram_last_addr - initrd_size,
+        initrd_addr: ram_last_addr.min(MMIO_MEM_START) - initrd_size,
         firmware_addr,
     };
     (info, regions)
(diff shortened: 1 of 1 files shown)


src/arch/src/x86_64/mod.rs
```diff
@@ -196,7 +196,7 @@ pub fn arch_memory_regions(
         ram_last_addr,
         shm_start_addr,
         page_size,
-        initrd_addr: ram_last_addr - initrd_size,
+        initrd_addr: ram_last_addr.min(MMIO_MEM_START) - initrd_size,
```
bilelmoussaoui, line 199: I can confirm this issue and used this as a workaround, see https://github.com/bilelmoussaoui/libkrun/commit/a6aeb65d68bbdb6333c94e72fbf80aff9558cb5d

jakecorrenti (approved):
code lgtm

slp:
@ben-gordon-dev LGTM too, thanks!

PR #826, opened 2026-08-31
vmm/tee: size TDX qboot measured RAM from guest mapping

jakecorrenti:
measure_qboot_regions hardcoded a 2 GiB PAGE.ADD region, which broke guests with mem_size_mib != 2048. Derive the region from actual GuestMemoryMmap RAM below the MMIO gap instead.


slp (approved):
LGTM, thanks @jakecorrenti

Issue #565, opened 2026-02-27
virtio-gpu device hardcodes the number of capsets instead of querying Rutabaga


valpackett:
(From https://github.com/containers/libkrun/pull/560#pullrequestreview-3864399227)

As we're going for the ability to eventually configure which capsets Rutabaga would be configured with, I was shocked to discover that…

https://github.com/containers/libkrun/blob/757b080b4c5f5934f8e5320a38b401aaec116764/src/devices/src/virtio/gpu/device.rs#L162-L166

`num_capsets` in the MMIO-readable config is hardcoded to 5 and has been ever since rutabaga was imported into libkrun!

And this is what Linux uses to determine how many capsets to query. If we have fewer capsets, it all blows up.

```
[    0.119895] [drm] cap set 0: id 6, max-version 0, max-size 592
[    0.119946] [drm] cap set 1: id 5, max-version 0, max-size 16
[    0.119992] [drm:virtio_gpu_dequeue_ctrl_func] *ERROR* response 0x1200 (command 0x108)
[    5.129781] [drm:virtio_gpu_get_capsets] *ERROR* timed out waiting for cap set 2
[    5.130347] [drm] Initialized virtio_gpu 0.1.0 for a005000.virtio_mmio on minor 0
```

```
[2026-02-27T03:41:29Z WARN  rutabaga_gfx::rutabaga_core] CapSets [RutabagaCapsetInfo { capset_id: 6, component: VirglRenderer, name: "drm" }, RutabagaCapsetInfo { capset_id: 5, component: CrossDomain, name: "cross-domain" }]
[2026-02-27T03:41:29Z WARN  rutabaga_gfx::rutabaga_core] get_capset_info 0
[2026-02-27T03:41:29Z WARN  rutabaga_gfx::virgl_renderer] virgl_renderer_get_cap_set..
[2026-02-27T03:41:29Z INFO  rutabaga_gfx::virgl_renderer] drm: drm_renderer_capset:135: c=(nil)
[2026-02-27T03:41:29Z WARN  rutabaga_gfx::virgl_renderer] virgl_renderer_get_cap_set: 0 592
[2026-02-27T03:41:29Z WARN  rutabaga_gfx::rutabaga_core] get_capset_info 0: 0 592
[2026-02-27T03:41:29Z WARN  rutabaga_gfx::rutabaga_core] get_capset_info 1
[2026-02-27T03:41:29Z WARN  rutabaga_gfx::cross_domain] CroDo get_capset_info
[...]

PR #829, opened 2026-09-01
gpu: set num_capsets to the count of capsets passed to rutabaga, fix #565

valpackett:
The capset_mask defines which virtio-gpu capability sets are enabled, but the number of them must be available very early, before the worker is even initialized. So extract the current way of calculating the mask and call it from the virtio device code.

---

Now that #560 is in but didn't include this…

### DIFF
--- a/src/devices/src/virtio/gpu/device.rs
+++ b/src/devices/src/virtio/gpu/device.rs
@@ -11,6 +11,7 @@ use super::super::{
 use super::defs;
 use super::defs::uapi;
 use super::defs::uapi::virtio_gpu_config;
+use super::virtio_gpu::virgl_flags_to_capsets;
 use super::worker::Worker;
 use crate::display::DisplayInfo;
 use crate::virtio::InterruptTransport;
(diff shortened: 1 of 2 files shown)


src/devices/src/virtio/gpu/device.rs
```diff
@@ -163,7 +164,7 @@ impl VirtioDevice for Gpu {
             events_read: 0,
             events_clear: 0,
             num_scanouts: self.displays.len() as u32,
-            num_capsets: 5,
+            num_capsets: 1 + virgl_flags_to_capsets(self.virgl_flags).count_ones(),
```
mtjhrc: Hmm why is this `1+` ?

valpackett: Corresponds to the unconditional `| 1 << rutabaga_gfx::RUTABAGA_CAPSET_CROSS_DOMAIN`… Well yeah actually better not to hoist it outside of the thing

mtjhrc (approved):
Nice! Thanks!

PR #831, opened 2026-09-01
Merge `vmm` crate into `libkrun` crate

mtjhrc:
As part of the #718 PR I had originally moved some parts of the VMM create into libkrun, this has however been very annoying when rebasing. Then I realized that we could just move the whole crate since libkrun is the only user and I think this should make the code easier to wrok with here. 

After moving it I have discovered a lot of dead code, which this PR attempts to remove.


jakecorrenti (approved):
LGTM. TDX seems to compile and run fine.

PR #836, opened 2026-09-04
virtio/fs/win: Windows backend optimizations, WSL symlink support, and flag fixes

lstocchi:
Recently i had to test libkrun on a specific workflow to create/edit a ext4 fs inside the VM from a rootfs shared through virtio-fs in Windows. It came out that something did not work as expected/was too slow and i tried to enhance it. This PR contains the changes I made and still did not push + a pair macOS's PermissionSemantics impl


slp (approved):
LGTM, thanks @lstocchi

PR #837, opened 2026-09-04
libkrun: add WHP MMIO device manager

lstocchi:
Add a device_manager/whp/ module following the HVF pattern: simpler than the KVM version because there are no ioeventfd/irqfd mechanisms. MMIO device registration, address allocation, and interrupt routing are handled entirely in userspace.


slp (approved):
Nice addition, let's go!

PR #843, opened 2026-09-07
libkrun: implement an API to set render server fd

slp:
Implement an API that allows users passing to us a render server fd to
be used instead of having libvirglrenderer starting it on its own.

This provides users total control over the context where the render
server is started.

### DIFF
--- a/src/devices/src/virtio/gpu/virtio_gpu.rs
+++ b/src/devices/src/virtio/gpu/virtio_gpu.rs
@@ -3,6 +3,7 @@ use std::env;
 use std::io::IoSliceMut;
 #[cfg(target_os = "linux")]
 use std::os::fd::AsRawFd;
+use std::os::fd::RawFd;
 use std::path::PathBuf;
 use std::sync::{Arc, Mutex};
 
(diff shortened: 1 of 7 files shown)


src/devices/src/virtio/gpu/virtio_gpu.rs
```diff
@@ -278,7 +281,13 @@ impl VirtioGpu {
 
         let fence =
             Self::create_fence_handler(mem, queue_ctl.clone(), fence_state.clone(), interrupt);
-        builder.clone().build(fence.clone(), None).ok()
+        builder
+            .clone()
+            .build(
+                fence.clone(),
+                render_server_fd.map(|fd| unsafe { RutabagaDescriptor::from_raw_descriptor(fd) }),
```
dorindabassey, line 288: missing safety comment.

src/devices/src/virtio/gpu/virtio_gpu.rs
```diff
@@ -221,6 +223,7 @@ impl VirtioGpu {
         interrupt: InterruptTransport,
         fence_state: Arc<Mutex<FenceState>>,
         virgl_flags: u32,
+        render_server_fd: Option<RawFd>,
```
dorindabassey, line 226: if `render_server_fd` is now the caller responsibility, should we use `OwnedFd` to prevent the user from accidentally closing it mid-operation?

dorindabassey (commented):
LGTM, just a few comments.

src/devices/src/virtio/gpu/virtio_gpu.rs
mtjhrc: I'm not sure if we want to dup every file descriptor... But yeah RawFd is not great and we should prefer `BorrowedFd`/`OwnedFd`. But for 1.x branch I think this is good enough. 

Ideally in 2.x we should use `BorrowedFd` in all public APIs and propagate the lifetime, but I am not sure how feasible it is. (we may need some kind of helper to help us manage lifetime across threads).

PR #844, opened 2026-09-08
virtio/gpu: Allow enabling gfxstream feature

bilelmoussaoui:
Needed for using that with Android as a rendering backend instead of virgl.

Tested with Android 17.

### DIFF
--- a/src/devices/Cargo.toml
+++ b/src/devices/Cargo.toml
@@ -15,6 +15,7 @@ tdx = ["blk", "tee"]
 net = []
 blk = []
 gpu = ["rutabaga_gfx", "thiserror", "zerocopy", "krun_display"]
+gfxstream = ["rutabaga_gfx/gfxstream"]
 input = ["zerocopy", "krun_input"]
 virgl_resource_map2 = []
 aws-nitro = []
@@ -52,7 +53,6 @@ hvf = { package = "krun-hvf", version = "=0.1.0-2.0.0-dev", path = "../hvf" }
 lru = ">=0.9"
 
 [target.'cfg(target_os = "linux")'.dependencies]
-rutabaga_gfx = { version = "0.1.85", features = ["virgl_renderer"], optional = true }
 caps = "0.5.5"
 kvm-bindings = { version = "0.14.1", features = ["fam-wrappers"] }
 kvm-ioctls = "0.25"
--- a/src/devices/src/virtio/gpu/virtio_gpu.rs
+++ b/src/devices/src/virtio/gpu/virtio_gpu.rs
@@ -1156,5 +1156,11 @@ pub fn virgl_flags_to_capsets(flags: u32) -> u64 {
         capset_mask |= 1 << rutabaga_gfx::RUTABAGA_CAPSET_DRM;
     }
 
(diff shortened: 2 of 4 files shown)


src/devices/Cargo.toml
```diff
@@ -52,7 +53,6 @@ hvf = { package = "krun-hvf", version = "=0.1.0-2.0.0-dev", path = "../hvf" }
 lru = ">=0.9"
 
 [target.'cfg(target_os = "linux")'.dependencies]
-rutabaga_gfx = { version = "0.1.85", features = ["virgl_renderer"], optional = true }
```
bilelmoussaoui, line 55: This dep is a duplicate because it is defined both for linux and globally, same thing.

src/devices/src/virtio/gpu/virtio_gpu.rs
```diff
@@ -1156,5 +1156,11 @@ pub fn virgl_flags_to_capsets(flags: u32) -> u64 {
         capset_mask |= 1 << rutabaga_gfx::RUTABAGA_CAPSET_DRM;
     }
 
+    if flags & VIRGLRENDERER_NO_VIRGL != 0 {
+        capset_mask |= 1 << rutabaga_gfx::RUTABAGA_CAPSET_GFXSTREAM_VULKAN;
+        capset_mask |= 1 << rutabaga_gfx::RUTABAGA_CAPSET_GFXSTREAM_GLES;
+        capset_mask |= 1 << rutabaga_gfx::RUTABAGA_CAPSET_GFXSTREAM_COMPOSER;
+    }
```
mtjhrc, line 1163: This isn't correct `VIRGLRENDERER_NO_VIRGL` is intended to disable `virgl` feature but still use virglrenderer - this is inteded so you can only enable  venus or drm native context. 
Note that you can't really use gfxstream and virglrenderer at the same time in Rutabaga. 

We need to introduce another API or flag here.

mtjhrc (changes requested):

src/devices/src/virtio/gpu/virtio_gpu.rs
bilelmoussaoui: I see. I wasn't aware of that, so i just "hacked" something around that. Do you have any ideas on how such API could look like? I can spend sometime working on it

mtjhrc: Well, we could accept 2 separate arguments 1 being GpuMode (or backend?) which would currently be  either Virgl or Gfxstream (but there might be others in the future) and another being a capset mask. 

IIRC capset mask being zero would map to unaccelerated display output only for both backends basically.

Or we could have a separate function for adding a virgl vs gfxstream gpu - I think this would be more ergonomic in the upcoming new Rust API.

Btw you might find this code in vhost-device-gpu useful:
https://github.com/rust-vmm/vhost-device/blob/20fa14c4c56e40a12104794a934dd70dc7642ff2/vhost-device-gpu/src/lib.rs#L53-L301

bilelmoussaoui: Thanks, i will dig into this :)

PR #848, opened 2026-09-09
Disable pit reinjection

jakogut:
PIT reinjection inhibits KVM APIC acceleration, including inter-vCPU interrupts. KVM recommends disabling it except for older guests that rely on PIT-based timekeeping [0].

libkrunfw ships Linux v6.12. External kernels need virtio-MMIO v2 support (Linux v4.0 [1]) to use libkrun's MMIO devices, and the bundled init requires cgroup2 (Linux v4.5 [2]). These requirements mean typical guests are already new enough not to need PIT reinjection.

Disable reinjection with KVM_REINJECT_CONTROL after PIT creation, retaining the device for calibration and fallback use.

[0] https://www.kernel.org/doc/html/v6.12/virt/kvm/api.html#kvm-reinject-control
[1] https://github.com/torvalds/linux/blob/v4.0/drivers/virtio/virtio_mmio.c
[2] https://github.com/torvalds/linux/blob/v4.5/kernel/cgroup.c

Assisted-by: Codex: gpt-6

EDIT: Disabling PIT reinjection increases vkmark throughput by 12-14%, and reduces CPU time per frame by appx. 10%, as tested on a Ryzen 9950X and Radeon RX 9070 XT with RADV native contexts.


src/devices/src/legacy/kvmioapic.rs
```diff
@@ -2,15 +2,22 @@
...
 use crate::Error as DeviceError;
 use crate::bus::BusDevice;
 use crate::legacy::irqchip::IrqChipT;
 
-use kvm_bindings::{KVM_PIT_SPEAKER_DUMMY, kvm_pit_config};
+use kvm_bindings::{KVM_PIT_SPEAKER_DUMMY, kvm_pit_config, kvm_reinject_control};
 use kvm_ioctls::{Error, VmFd};
 use utils::eventfd::EventFd;
 
+nix::ioctl_write_ptr_bad!(
```
slp, line 17: Strictly speaking, we should avoid implementing the KVM ioctls manually, relying on what `kvm-ioctls` provides. In this particular case, it doesn't yet support `KVM_REINJECT_CONTROL`, so I'm fine doing this here, but please add a comment here that this needs to be replaced once a version of `kvm-ioctls` including https://github.com/rust-vmm/kvm/pull/386 is released.

slp (changes requested):

src/devices/src/legacy/kvmioapic.rs
jakogut: Thanks for the review, and kvm-ioctls PR. I'll add the comment.

slp (approved):
LGTM, thanks @jakogut !

PR #851, opened 2026-09-11
Move bindings to a separate directory in the repo

mtjhrc:
The  can live at a similar level than `include` which makes conceptual sense since they basically accomplish a similar goal and are easier to find.


slp (approved):
LGTM, thanks @mtjhrc

PR #852, opened 2026-09-13
virtio/vsock: release proxies after clean shutdown

RobertDeRose:
A stream proxy currently handles VSOCK_OP_SHUTDOWN by shutting down the backing host socket, but it leaves the proxy in proxy_map. If the guest closes the stream cleanly and never sends a later VSOCK_OP_RST, that entry and its host socket remain alive indefinitely.

This is especially visible with guest-to-host Unix port mappings. After the guest closes a mapped stream, reconnecting from the same guest port can find the stale proxy instead of creating a fresh connection.

Track the peer shutdown bits on stream proxies, matching the cumulative handling used by the Linux virtio-vsock transport. Once both receive and send shutdown have been observed, send the reset that completes the virtio-vsock close handshake and return the proxy's normal release update so the existing reaper can remove it safely.

Keep one-sided shutdowns as half-closes.

Add an end-to-end regression test that reconnects from the same guest vsock port after the reaper interval. Without the fix the stale proxy is still present and the second mapped connection cannot complete.

Needs label: `cherry-pick-1.19.x`


slp (approved):
Nice PR, thanks @RobertDeRose

PR #853, opened 2026-09-14
vmm/tdx: support high memory and multiple vCPUs with TD-Shim

billionairiam:
## Summary

- expose CPUID.01H:ECX.x2APIC to TDX guests without clearing unrelated feature bits
- include every guest RAM region, including RAM above the 32-bit MMIO gap, in TDX private-memory initialization and TD-Shim HOBs
- exclude separately mapped firmware ranges from RAM regions and reuse the helper for qboot
- clarify that the 1-vCPU/3072-MiB limit applies to the legacy qboot path

TD-Shim multi-vCPU boot also requires confidential-containers/td-shim commit [`e3a692b1`](https://github.com/confidential-containers/td-shim/commit/e3a692b1e58c59b40647d919f6de8ae69b2c8846), which removes RBP from the TDVMCALL exposed-register mask for `TDX_CONFIG_FLAGS_NO_RBP_MOD` guests. The hardware tests below used unmodified TD-Shim commit `0362d0bc6d784170af28551e0b9bc6fd5ad414d2`, which contains that fix.

## Validation

- `cargo fmt -- --check`
- `cargo test --locked -p krun-cpuid --features tdx`: 41 unit tests and 5 doctests passed
- `cargo clippy --locked --features tdx -- -D warnings`
- `make TDX=1 FFI=1`
- Intel TDX hardware:

| Firmware | vCPUs | RAM | Guest result |
|---|---:|---:|---|
| qboot | 1 | 2048 MiB | CPU 0 online, 1908028 kB MemTotal |
| qboot | 1 | 3072 MiB | CPU 0 online, 2804540 kB MemTotal |
| TD-Shim | 1 | 4096 MiB | CPU 0 online, 3834540 kB MemTotal |
| TD-Shim | 2 | 4096 MiB | CPUs 0-1 online, 3834356 kB MemTotal |
| TD-Shim | 4 | 8192 MiB | CPUs 0-3 online, 7691396 kB MemTotal |
| TD-Shim | 8 | 16384 MiB | CPUs 0-7 online, 15405472 kB MemTotal |

Every CPU in the TD-Shim multi-vCPU runs reported both `x2apic` and `tdx_guest`. The guest emitted its success marker in every run, with no `KVM_EXIT_SHUTDOWN`, kernel panic, BUG, or OOM marker.

The targeted `libkrun` TDX unit-test binary currently does not compile because existing TDX-gated tests still call older `setup_vm`, `Vm::new`, and `Vcpu::new_x86_64` signatures and contain an unrelated packed-field assertion. These failures are outside this change; the new RAM helper is covered by a unit test that will run once that existing test target is repaired.


jakecorrenti (commented):
Thanks, @billionairiam! This was on my TODO list :) 

Code LGTM, and I tested it on my TDX machine and it seems to work fine. Could you rebase before I approve this?

I'll also open a separate PR fixing the broken tests.

mtjhrc:
@jakecorrenti 
> I'll also open a separate PR fixing the broken tests.

If you do, it would be nice to also look into at least compiling the tests in CI (even if they can't run). I think we could do `cargo clippy --tests --features ...` (or at least `cargo check --tests --features ...`).

jakecorrenti (approved):
LGTM!

PR #855, opened 2026-09-14
devices: Fix build errors and add platform support for CMOS RTC and Error Translation

lstocchi:
When moving from my working branch to upstream, AI and me forgot something when preparing old PRs. This one addresses several Windows build errors across the devices crate. It enables real-time clock (RTC) support for CMOS on Windows, adds WSA-to-Linux error translation, re-exports shared platform abstractions in vsock proxies, and fixes cross-platform struct compatibility (statvfs64, ino64_t).


src/devices/src/virtio/linux_errno.rs
```diff
@@ -106,6 +145,7 @@ pub fn linux_errno_raw(errno: i32) -> i32 {
         libc::ENOMEM => LINUX_ENOMEM,
         libc::EACCES => LINUX_EACCES,
         libc::EFAULT => LINUX_EFAULT,
+        #[cfg(not(target_os = "windows"))]
```
mz-pdm, line 157: Why are some of the mappings excluded?  I don't get which ones should be excluded and if another mechanism could be possibly used.

mz-pdm (commented):
Not a Windows expert here but the PR looks plausible, just one question inside.

src/devices/src/virtio/linux_errno.rs
lstocchi: The libc crate on Windows does not define some POSIX constants so i just excluded them

mz-pdm: OK, I don't have a better idea how to deal with it.

mz-pdm (dismissed):

src/devices/src/virtio/linux_errno.rs
```diff
@@ -87,6 +87,45 @@ const LINUX_ENOTRECOVERABLE: i32 = 131;
...
 pub const LINUX_ERANGE: i32 = 34;
 
+/// Translate WSA error codes to Linux errno values.
+///
+/// On Windows, WinSock errors live in the 10000+ range (WSA* constants).
+/// We map them to their Linux equivalents so the guest sees the right errno.
+#[cfg(target_os = "windows")]
+pub fn wsa_errno_to_linux(wsa_err: i32) -> i32 {
+    match wsa_err {
+        10035 => LINUX_EAGAIN,          // WSAEWOULDBLOCK
```
slp: Looks like these error types are defined in `windows_sys`:

https://docs.rs/windows-sys/latest/windows_sys/Win32/Networking/WinSock/constant.WSAEWOULDBLOCK.html

Could we use those symbols instead of hard coding the values?

lstocchi: Good point, done 👍

slp (approved):
Thanks @lstocchi !

PR #856, opened 2026-09-14
virtio/net: add Windows AF_UNIX stream support

lstocchi:
This introduces Windows support to the virtio-net stream backend by leveraging Winsock AF_UNIX sockets.

The Unix stream backend and network worker have been split into platform-specific modules, preserving the existing Unix implementation and behavior. For Windows targets, this change adds Winsock initialization, AF_UNIX connection handling, non-blocking I/O, and custom socket buffer configurations alongside dedicated RX and TX workers.


slp (approved):
LGTM, thanks @lstocchi !

PR #857, opened 2026-09-14
arch/x86_64: fix FADT 8042 flag and boot protocol version

jakecorrenti:
Advertise the emulated i8042 so Linux does not treat it as firmware-absent, and stop claiming burnt protocol 2.14; the RSDP already lives at boot_params.acpi_rsdp_addr.

This fixes the comments that were left on #807 post-merge.


mtjhrc (approved):
LGTM!

PR #858, opened 2026-09-15
windows: enable libkrun builds with WHP

lstocchi:
Add the remaining Windows support needed to build libkrun using the Windows Hypervisor Platform (WHP).

These commits adds WIndows support to the libkrun crate with additional fixes to make it work with WHP. The last commit must be dropped before merging as it is only used to inject the linux-loader version we need for test this.


slp (approved):
LGTM, thanks @lstocchi

PR #861, opened 2026-09-16
Makefile: update FreeBSD test image to 14.5

slp:
Version 14.3 is no longer available, switch to 14.5.


nohajc:
Btw, not relevant for the sysroot part but FreeBSD VM tests rely on the `kernel.bin` image for arm64 which is no longer distributed since FreeBSD 15.0.

This is the same format Linux uses on arm, so it was always bootable without any changes in libkrun (except some kernel command-line processing quirks).

They only distribute the ELF version now but you can still generate the `.bin` with a custom kernel build config.

ELF kernel on arm64 isn't a supported combination in libkrun, right?

slp:
> ELF kernel on arm64 isn't a supported combination in libkrun, right?

If it doesn't work, it should be a very easy fix.

nohajc:
I just remember it was gated to x86_64. Haven't really investigated why...

mtjhrc (approved):
LGTM, thanks!

PR #863, opened 2026-09-16
tests/vsock_guest_reconnect: increase REAPER time

slp:
Our poor aarch64 runner is having a hard time giving enough CPU time to the VMs, to the point vsock reaper may take longer than 7 seconds to run.

Be conservative by increasing the reaper timeout to 20 seconds.


mtjhrc (approved):
LGTM. I don't love having a fixed 20-second wait in the test (I do tend to run the tests locally pretty often), but this is good enough for now to fix the CI.

I haven't looked into this enough, but it looked like trying multiple opportunistic reconnects would mess up the test.

slp:
> LGTM. I don't love having a fixed 20-second wait in the test (I do tend to run the tests locally pretty often), but this is good enough for now to fix the CI.

Me neither. Let's put it on hold to see how frequently fails on aarch64. Perhaps the runner was having a bad day. We could even put it behind an environment variable.

dorindabassey (approved):

PR #865, opened 2026-09-17
virtio/blk: add opt-in parallel reads

### LINKED ISSUE #824: virtio-blk services one request at a time per disk, so guest I/O cannot scale beyond a single host thread
In `src/devices/src/virtio/block/worker.rs`, one worker thread per disk pops each request and executes it synchronously (`process_queue` → `process_request` → blocking `preadv`), completing it before the next request is popped. The device therefore never sees more than one outstanding request per disk, regardless of the queue depth the guest presents. This is unchanged in 1.19.4 and current main.

The effect is easy to observe: inside the guest, several parallel readers on a raw disk (attached with `direct_io` so the host page cache is out of the picture) achieve no more aggregate throughput or IOPS than a single reader — in fact slightly less — while the host reading the same file gains from a second thread before saturating the device. The gap grows with per-request latency, so it is most visible on external/USB disks.

### Minimal repro (macOS, Apple Silicon)

Three files; needs only `brew install libkrun libkrunfw` and the Alpine minirootfs.

`repro.c`
```c
#include <libkrun.h>
#include <stdio.h>
int main(int argc, char **argv) {
    if (argc != 3) { fprintf(stderr, "usage: %s <rootfs> <raw_disk>\n", argv[0]); return 1; }
    const char *args[] = { "/bench.sh", NULL };
    const char *env[]  = { "PATH=/bin:/usr/bin", NULL };
    int ctx = krun_create_ctx();
    krun_set_vm_config(ctx, 4, 1024);
    krun_set_root(ctx, argv[1]);
    krun_add_disk3(ctx, "d", argv[2], KRUN_DISK_FORMAT_RAW, true, /*direct_io*/ true, KRUN_SYNC_NONE);
    krun_set_exec(ctx, "/bin/sh", args, env);
    return krun_start_enter(ctx);
}
```

`bench.sh` (copied into the rootfs; runs in the guest)
```sh
#!/bin/sh
t() { cut -d' ' -f1 /proc/uptime; }
r() { dd if=/dev/vda of=/dev/null bs=1M count=1024 skip=$1 2>/dev/null; }
i() { dd if=/dev/vda of=/dev/null bs=4k count=16384 skip=$1 iflag=direct 2>/dev/null; }
a=$(t); r 0; b=$(t);                              echo "seq 1 reader :  $(echo $a $b | awk '{printf "%5d", 1024/($2-$1)}') MiB/s"
a=$(t); r 1024 & r 2048 & r 3072 & r 4096 & wait; b=$(t); echo "seq 4 readers:  $(echo $a $b | awk '{printf "%5d", 4096/($2-$1)}') MiB/s aggregate"
a=$(t); i 0; b=$(t);                              echo "4K  1 reader :  $(echo $a $b | awk '{printf "%5d", 16384/($2-$1)/1000}')k IOPS"
a=$(t); i 262144 & i 524288 & i 786432 & i 1048576 & wait; b=$(t); echo "4K  4 readers:  $(echo $a $b | awk '{printf "%5d", 65536/($2-$1)/1000}')k IOPS aggregate"
```

`run.sh`
```sh
#!/bin/sh
set -e; cd "$(dirname "$0")"
brew list libkrun libkrunfw >/dev/null 2>&1 || { brew tap slp/krun; brew install libkrun libkrunfw; }
[ -d rootfs ] || { curl -fsSL https://dl-cdn.alpinelinux.org/alpine/v3.22/releases/aarch64/alpine-minirootfs-3.22.1-aarch64.tar.gz | { mkdir rootfs; tar -xz -C rootfs; }; }
cp bench.sh rootfs/bench.sh; chmod +x rootfs/bench.sh
DISK=${1:-disk.img}; [ -e "$DISK" ] || dd if=/dev/zero of="$DISK" bs=1m count=6144 status=none
cc -O2 -w repro.c -o repro $(pkg-config --cflags --libs libkrun)
codesign --entitlements hvf-entitlements.plist --force -s - repro   # com.apple.security.hypervisor
DYLD_LIBRARY_PATH=/opt/homebrew/lib ./repro rootfs "$DISK"
```

Expected with a concurrent backend: the "4 readers" lines well above the "1 reader" lines. Observed: equal or lower. Reading the same file from the host with 2–4 threads (`pread` + `F_NOCACHE`, disjoint ranges) shows the device itself has headroom.

### Context

This looks consistent with the fio results in containers/podman#27216; #428/#452 changed sync and cache semantics but not the serialized request loop, and it may contribute to the residual gap in #414.

### Possible fix

Service reads concurrently with out-of-order completion (virtio-blk permits it). imago's `readv`/`writev`/`flush` are `&self`; only `discard_*` need `&mut self`, so the current `Mutex` around `SyncFormatAccess` could become an `RwLock`, with the worker dispatching queued reads to a small pool and completing them as they land. A quick experiment of this shape gave a substantial read throughput improvement on a USB SSD with byte-identical results.

jakecorrenti:
virtio-blk handled one request at a time on the device thread, so guest
  I/O could not scale past a single host thread no matter what queue depth
  the guest presented.
  This change:
  - Switches the disk image from `Mutex` to `RwLock` so reads can run
    concurrently while writes, flushes, and discards still take exclusive
    access.
  - Dispatches `VIRTIO_BLK_T_IN` to a host thread pool (2–8 workers from
    `available_parallelism`) and completes them out of order. The device
    does not advertise `VIRTIO_F_IN_ORDER`.
  - Keeps writes, flushes, and discards on the device thread.
  - Makes the pool **opt-in**. Default remains serialized. Enable with:
```c
    // C API
    krun_block_device_set_parallel_reads(blk, true);
```
```rust
   // Rust API
   BlockDevice::set_parallel_reads(true)
```

  Fixes #824

  ## Benchmark
  Linux x86_64 / KVM, 4 vCPUs, 1 GiB RAM, 6 GiB raw image, `O_DIRECT` on
  host and guest. Same #824 `dd` workload, only
  `krun_block_device_set_parallel_reads()` changed.
  | Workload | serialized (default) | parallel reads |
  |---|---|---|
  | Sequential 1 reader (1 GiB, 1 MiB blocks) | 2008 MiB/s | 1796 MiB/s |
  | Sequential 4 readers | 959 MiB/s aggregate | **1489 MiB/s** aggregate |
  | 4K 1 reader | 13.0k IOPS | 11.1k IOPS |
  | 4K 4 readers | 29.3k IOPS (2.3×) | **45.5k IOPS (4.1×)** |


jakecorrenti:
@RobertDeRose just following up to see if you've had a chance to test this out

